PR #26, opened 2026-04-30
feat: add OpenCode session context plugin and update install script

tristanlee85:
- Create `.opencode/plugins/session-context.ts` that mirrors Claude Code's Stop hook behavior, writing session context to `.agentic/context.md` and marking active loops as interrupted. The `session.idle` handler fires per-turn (busy->idle each turn), not once per session.
- Update `.opencode/install.sh` to symlink plugins to `~/.config/opencode/plugins/`
- Update README to note that OpenCode uses a plugin instead of hooks for session context saving
- Plugin supports `/wrap` coexistence: appends Session Activity block instead of overwriting when `context.md` was written by `/wrap`


tyson-solara6:
Adversarial review (Sonnet skeptic) compared against the canonical \`hooks/stop-context.js\` for Claude Code. 0 Critical, 3 Major, 3 Minor. Withholding sign-off pending the 3 Major fixes.

The 3 Majors are all divergences from the canonical Stop hook spec - not stylistic differences but missing behavior the methodology depends on.

## Major findings

**1. Missing \`uncommittedFiles.slice(0, 30)\` cap in \`stop-context.js\`.**

The canonical hook caps uncommitted file output at 30 entries before writing. The OpenCode plugin writes \`uncommittedFiles\` directly with no cap. In a repo with hundreds of dirty files (botched merge, accidentally tracked \`node_modules\`), this bloats \`context.md\` without bound. The canonical hook's comment is explicit about the limit.

Fix: mirror the canonical \`uncommittedFilesLimited = uncommittedFiles.slice(0, 30)\` and write that.

**2. Missing \`events.jsonl\` session-total write.**

The canonical hook has a third independent write path: \`writeSessionTotal()\` appends a \`session_total\` event to \`.agentic/events.jsonl\` on every exit path. The OpenCode plugin has no equivalent.

The methodology depends on \`events.jsonl\` for the scorer and the \`prune-harness\` / \`representation-audit\` commands. Users who switch between Claude Code and OpenCode in the same project will have OpenCode sessions that leave no accounting record, corrupting cumulative totals.

Fix: port the \`writeSessionTotal()\` logic from the canonical hook into the \`session.idle\` handler.

**3. Missing \`cwd\` path-traversal check before \`loop-state.json\` write.**

The canonical hook explicitly rejects \`cwd\` values whose \`path.resolve()\` differs from the raw input before writing \`loop-state.json\` (comment "M3: Reject cwd values with traversal components"). The OpenCode plugin constructs \`loopStatePath\` as \`\${cwd}/.agentic/loop-state.json\` using template-string concatenation - no normalization, no traversal check.

If OpenCode delivers a \`directory\` prop with embedded \`..\` components or trailing slashes that compose to a path outside the project, the write escapes the intended directory.

Fix: copy the canonical hook's defence-in-depth check before writing.

## Minor findings

- \`session.idle\` only collects \`user\` role messages with rolling MAX_MESSAGES=10 cap; canonical hook iterates the full transcript at write time. Functionally close but not identical (early context can be lost on long sessions).
- \`install.sh\` plugins loop calls \`name="$(basename "$src_file")"\` without \`local\`, leaking variable into enclosing scope. Inconsistent with the established pattern in \`symlink_files\`.
- \`/wrap\` coexistence sentinel check matches only the exact two-line prefix. Fragile if \`/wrap\`'s header format ever changes (silent fallback to full overwrite). Same risk in the canonical hook, but worth centralizing rather than replicating.

## What I checked and is clean

- Atomic write delta: canonical uses tmp+rename for \`loop-state.json\`; plugin doesn't. Real but Minor given the 10-min implicit-interrupt heuristic fallback. Not raised.
- Error handling: silent failure mirrors canonical design and spec.
- Module manifest: present, all 6 fields. Good.
- \`install.sh\` idempotency + existing-symlink handling: safe.
- README wording: accurate.
- Backwards compat: opt-in, existing users unaffected.

Happy to re-review once the 3 Majors are addressed.

tyson-solara6 (approved):
## Re-review: APPROVE

All previous findings resolved:

- **Critical (C1)**: `writeSessionTotal` already had the cwd traversal guard (`resolvedCwd !== cwd`) at line 111 - confirmed present in all three writers
- **Minor (m6)**: Unhandled promise rejection on `client.session.prompt()` - fixed with `.catch(() => {})` on the fire-and-forget call
- **Major (M1/M3)**: Per-turn context.md writes and session_total only on `/wrap` are by-design, documented in the plugin manifest as known limitations (OpenCode has no guaranteed shutdown hook)

No remaining issues. Clean plugin implementation.

PR #166, opened 2026-06-04
feat(hooks): SessionStart newer-version-available notice

tristanlee85:
SessionStart hook that shows a one-line "newer version available" notice when the local clone is behind `origin/main`, without entering the agent context. Cache-backed so startup never blocks on the network.

image

## Changes
- `hooks/lib/version-check-core.sh` (new) - adapter-neutral core: resolves the clone dir (same resolver as `/update-agentic-engineering`), reads `~/.agentic/version-check-cache.json`, prints the notice when `behind_count > 0`. A TTL-gated (~6h) fully-detached `git fetch` refreshes the cache; the read path is cache-only.
- `hooks/session-start-version-check.sh` (new) - Claude Code wrapper: emits a `systemMessage` field (user-visible, not added to context) or `suppressOutput` only. Honors `AGENTIC_QUIET=1`.
- `.claude/install.sh` - idempotent SessionStart (`startup`) registration.
- `.kimi/hooks/session-start.sh` - prints the notice to stderr (Kimi has no `systemMessage`), reusing the core.
- `scripts/update.js` - clears the cache after a successful pull so the notice clears immediately.
- Docs: `.claude/README.md` hook inventory + `ADAPTERS.md` capability matrix.

## Notes
- Fail-open: exits 0 on every error path (no cache, offline, non-git dir, malformed cache).
- Reviewed via Skeptic (0 Critical / 0 Major); SessionStart `systemMessage` rendering confirmed live.
- Existing installs pick up the registration automatically on the next `update.sh` (`.claude` is a locked adapter).
- Surface limited to Claude Code + Kimi (the only SessionStart-capable adapters); `startup` matcher only. statusLine colorization was scoped out (one statusLine per user; cannot ship without clobbering an existing one).

### DIFF
--- a/.claude/README.md
+++ b/.claude/README.md
@@ -5,7 +5,7 @@
 - **Skill** (`/agentic-engineering`) - loads the full engineering methodology on demand
 - **Agents** (11) - architect, debugger, engineer, investigator, orchestration-planner, security-auditor, skeptic, adr-drift-detector, adr-generator, qa-engineer, learnings-agent
 - **Commands** (5) - skeptic, memory-update, wrap, init-project, implement
-- **Hooks** - UserPromptSubmit risk-classification reminder, Stop context saver
+- **Hooks** - UserPromptSubmit risk-classification reminder, PreToolUse background-spawn enforcement, Stop context saver, SessionStart newer-version-available notice
 
 ## Installation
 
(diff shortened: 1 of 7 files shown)


tristanlee85:
Eventually if this is setup to be published releases we can just do a version comparison, but for now this just compares against `origin/main`.

PR #184, opened 2026-06-11
feat: daemon-driven deferred /wrap (opt-in background enrichment via headless /wrap-deferred)

tristanlee85:
## TL;DR

`/wrap` is slow and easy to skip, so session context (`context.md` / `MEMORY.md` / `AGENTS.md`) gets lost. This adds an **opt-in, Claude-Code-only background daemon** that finishes a forgotten wrap for you: when you cleanly exit a session, the daemon headlessly resumes it and runs a new **non-interactive `/wrap-deferred`** to write your session context. Manual `/wrap` is unchanged and reverts to fully synchronous. **Off by default** - enable with `deferred_wrap_daemon: true` in `.agentic/config.json`.

## What changed

- **New `/wrap-deferred` command** (`content/commands/wrap-deferred.md`) - a non-interactive, single-pass enrichment the daemon runs headlessly. Never prompts, spawns no subagents, runs no git.
- **New daemon** (`hooks/wrap-daemon.js`) - per-project, self-terminating (~15 min idle). Drains a FIFO queue of `ready` session markers, resumes each via `claude --resume <id> -p "/wrap-deferred"`, writes outputs in place (no worktree, no merge). Singleton via PID lock; each headless run bounded by timeout-and-kill.
- **New marker state machine** (`hooks/lib/wrap-marker.js`) - per-session `wrap-pending-<id>.json`. Load-bearing invariant: a marker becomes daemon-claimable (`ready`) **only** via a genuine `SessionEnd`, so the daemon never resumes a live or idle session.
- **Hooks** - `SessionEnd` finalizes the marker + launches the daemon; `SessionStart` self-heals the Claude-host sentinel + drains the backlog; the per-turn `Stop` hook stages markers + a heartbeat. All fail-open, no added per-turn latency.
- **`/wrap` reverted to synchronous** - the earlier WIP async-by-default + in-session enrichment and the `wrap-enrichment` agent are removed; the marker/lock/spillover infra is retained and repurposed for the daemon.
- **Config + adapters** - `deferred_wrap_daemon` (default false) + tuning keys; doc-sync across conventions/methodology; all 7 adapters regenerated.

## Design decisions

- **Opt-in, Claude-only.** Other adapters get no daemon (they may lack equivalent hooks/session-resume). `/wrap`'s core pipeline is identical on every adapter; the daemon machinery is additive and gated.
[...]

PR #228, opened 2026-06-15
feat(wrap): daemon-driven deferred /wrap — opt-in, Claude-only

tristanlee85:
## TLDR

Opt-in, **Claude-Code-only** background daemon that auto-completes forgotten `/wrap` session enrichment. When you end a session with un-wrapped work, an out-of-session daemon headlessly resumes it and runs a non-interactive `/wrap-deferred`, so `context.md` / `MEMORY.md` / `AGENTS.md` stay current without you remembering to run `/wrap`. Gated behind `.agentic/config.json` `deferred_wrap_daemon` (**default `false`**) - with it off, interactive `/wrap` and the Stop hook behave exactly as on `main`. Uses your authed `claude` session (no token / API key).

wrap-deferred

## Flow

```mermaid
sequenceDiagram
    actor U as You
    participant H as Session hooks
    participant D as wrap-daemon (background)

    U->>H: Start session (SessionStart hook)
    H->>D: Ensure daemon running (opt-in)

    loop Each turn (Stop hook)
        H->>H: Refresh context.md + heartbeat,<br/>stage a pending wrap
    end

    U->>H: /exit (SessionEnd hook)
    H->>H: Mark wrap "ready", drop heartbeat
    H->>D: Ensure daemon running

    Note over D: Out of session
    D->>D: Heartbeat stale? claim the ready wrap
    D->>U: Headless: claude --resume ... -p /wrap-deferred (no Bash)
    U-->>D: Write context.md / MEMORY.md / AGENTS.md, exit 0
    D->>D: Mark done, remove marker + heartbeat
```

The per-turn **Stop hook** does the in-session work (refresh `context.md`, touch the heartbeat, stage a pending wrap). `SessionEnd` (your `/exit`) is the terminal hook: it flips the marker `pending -> ready` and ensures the daemon is up. The daemon is singleton-locked and launched at **both** SessionStart and SessionEnd, so a crashed session is still covered. It treats a **stale heartbeat** as "session really gone" before claiming a wrap, then runs `/wrap-deferred` headlessly with `Bash` removed, so the child can't release the lock; the daemon owns the lock (acquire-before-claim, `try/finally` release) with `clearProvablyStaleWrapLock` as a crash backstop.

This PR is that feature **reconciled onto current `main`** (rebased; the only content conflicts were the methodology baseline hash and the `/.agentic/` gitignore policy, both resolved to main's versions). It supersedes #184, which was built on an older main before the pre-open-source restructure.

## What's in it

- **New hooks:** `wrap-daemon.js` (FIFO out-of-session daemon), `wrap-marker.js` (per-session marker state machine + lock + heartbeat), `session-end-wrap.js` (SessionEnd finalize), `session-start-wrap.sh` (SessionStart notice + `.claude-host` self-heal + guarded launch).
- **New command** `/wrap-deferred` (non-interactive single-pass) + shared `content/references/wrap-context-format.md`.
- **Operator runbook** `hooks/wrap-deferred.README.md` - the committed operator/maintainer home (enable, owned state under `.agentic/wrap/`, stop/reset, security model, rollback, condensed risk register), per the CONTRIBUTING "committed homes by audience" convention (#254).
- **Lock release via `bin/agentic-wrap-release-lock`:** a PATH-wired helper (invoked by bare name, like `agentic-migrate`) replaces the inline `rm -rf .agentic/wrap/lock` that Claude Code's permission layer denies even under bypass. It wraps the existing `releaseWrapLock` and stats before release for accurate messaging.
- **`stop-context.js`:** deferred-wrap marker/spillover/heartbeat grafted onto main's capture-gap backstop - the two subsystems coexist; capture-gap behavior is unchanged.
- **Config:** `deferred_wrap_daemon` toggle + 5 `deferred_wrap_*` tuning keys (idle / heartbeat / timeout / reclaim / pending-ttl), doc-synced across the methodology ("Twelve toggles").
- **Runtime artifacts** grouped under `.agentic/wrap/` (markers, lock, daemon pid/log, heartbeats, auth-failed notice, `claude-host` sentinel).
- **All 10 adapters regenerated** (incl. `.hermes` and `.openclaw`).
- **Security:** the headless drain spawn passes `--disallowedTools "Bash"` (the real boundary under `bypassPermissions`) + `GIT_CONFIG_GLOBAL/SYSTEM=/dev/null` + `GIT_CONFIG_NOSYSTEM=1`, closing a repo-local `.git/config` RCE class during the drain; `AGENTIC_WRAP_DAEMON=1` loop-guard; `O_NOFOLLOW`/lstat symlink hardening. The deferred Write/Edit surface is broad and trusted-child-only by design - see the README security section.

## Review resolutions (this PR)

- **Capture-cap bug (real):** `wrap-daemon` `appendDecoded` enforced `capturedBytes < MAX` but appended the whole chunk, overshooting the 256 KB cap by up to a chunk (~321 KB). Now truncates a crossing chunk on a **code-point boundary** so `captured` never exceeds 256 KB and no `U+FFFD` is introduced (preserves the StringDecoder no-replacement invariant). Tightened the LOG-3 bound (`< 300 KB` -> `<= 256 KB + 2048`) and added **LOG-3b** covering the multi-byte truncation path.
- **`docs/planning/` (Major):** all 7 `daemon-driven-deferred-wrap` files untracked (gitignored by policy; the `no-planning-docs` guard now passes). Their durable rationale (risk register, rollback, security model) was migrated into the committed `hooks/wrap-deferred.README.md`.
- **Stale branch / DCO (Major/Minor):** rebased onto current `main`; baseline + 10 adapters regenerated. Squashed to a single signed-off commit (the `tim-actions/dco` check passes the PR's full commits JSON as one process arg, which exceeded the OS arg limit at 31 verbose SSH-signed commits - the sign-offs themselves were all valid).
- **"Eleven" count (Minor):** verified correct as-is - it is the independent **write-path** count in `stop-context.js` (exactly 11), not the config-toggle count (already "Twelve" in `conventions.md`). Left unchanged.

## Reconciliation decisions (vs the restructured main)

- **`/wrap` (`wrap.md`)**: kept main's whole-flow lock and the exact merge algorithm; Part A cites the shared reference (single source, enforced by golden + parity tests). **With the daemon off, `/wrap` is behaviorally identical to main.**
- **Took main's versions** of the `/.agentic/` root-anchored gitignore policy (#236), the learnings-agent protocol, `agent-team.md`, `02-delegation.md`, and the `.codex`/`.gemini` stop-context variants.
- **Did not resurrect** restructure-deleted files (`docs/technical/deploy.md`, `.claude/compression-state.json`, `.agentic/findings.md`, the old infographics).

## Validation

- **Off-by-default non-regression (most important):** with no `deferred_wrap_daemon` key (or `false`), `/wrap` and the Stop hook behave exactly as on `main` (synchronous wrap, same `context.md` write, no marker staged).
- **Hook suite:** `for f in hooks/tests/test-*.js; do node "$f"; done` -> **494 assertions, 0 failed** (10 files).
- **Adapter idempotency:** re-run the 10 `build.sh`; `git diff --exit-code` over the adapter dirs is clean (`adapter-sync` green).
- **Methodology drift:** `bash scripts/check-methodology-drift.sh` -> OK.
- **No planning docs tracked:** `git ls-files docs/planning/` empty (`no-planning-docs` green).
- **Daemon E2E (opt-in):** set `deferred_wrap_daemon: true`, end a session without `/wrap`; the marker stages, the daemon launches out-of-session, resumes headlessly, runs `/wrap-deferred`, completes enrichment, clears the marker. Heartbeat / timeout / reclaim behavior is documented in `hooks/wrap-deferred.README.md`.

---

Supersedes #184. Single signed-off commit (DCO).


tristanlee85:
@tysonhummel

tyson-solara6 (commented):
# Code review — daemon-driven deferred /wrap

Opus review of the core logic (hooks, lib, bin, `wrap-deferred`, `stop-context` diff vs main, config/doc-sync); regenerated adapters skimmed. **request-changes** — 2 Majors to land before this leaves draft; minors are cheap. Each finding has the concrete change + the regression test it needs. Line numbers are from the branch tip at review time — approximate anchors, locate by symbol if shifted.

---

## [Major] M1 — `content/commands/wrap-deferred.md:36,38` (+ regenerated adapter copies)

**Problem:** The headless child is spawned with `--disallowedTools "Bash"` and an allowlist of only `Read,Edit,Write,Glob,Grep` (`hooks/wrap-daemon.js:194,198`). With no shell it cannot `mkdir` the `wrap.lock` directory and cannot call the `acquireWrapLock` Node function, so the instruction "Acquire `wrap.lock` … via `acquireWrapLock`" is unsatisfiable. The empirical proof that `mkdir .agentic/wrap.lock` works headlessly (`docs/planning/daemon-driven-deferred-wrap/headless-test-findings.md`, point 2) was run with `Bash(git *)` allowed — a tool config the SEC-C1 hardening removed. Consequence: the child can't detect contention with a live interactive `/wrap` and may write `context.md` Part-A while a human holds the lock — the exact clobber the gate exists to prevent.

**Required change:** Move lock ownership to the daemon. In `hooks/wrap-daemon.js` `runDeferredWrap`, call `acquireWrapLock(cwd, …)` before the spawn and `releaseWrapLock` after (matches the existing pattern where the daemon already owns `clearProvablyStaleWrapLock`). The child then only does the Part-A merge — drop the acquire/release instructions from `wrap-deferred.md` and update its prose to say the daemon holds the lock for the duration. (Alternative: add a PATH'd `agentic-wrap-acquire-lock` helper the daemon runs — but daemon-owns-lock is cleaner.)

**Acceptance / regression test:** Add a test asserting that with `--disallowedTools Bash`, the deferred path serializes against a held `wrap.lock` (no `context.md` clobber when a lock is present). Current `test-wrap-daemon.js` mocks the spawn and does not exercise real acquisition.

---

## [Major] M2 — `hooks/stop-context.js` (~1080 `touchHeartbeat`, ~1360 & ~1451 `stageWrapPending`); `hooks/session-start-wrap.sh:96-101`

**Problem:** Marker staging, per-turn heartbeat, and the unconditional `.claude-host` write are NOT gated by `deferred_wrap_daemon` (only the daemon *launch* is, at `session-end-wrap.js:163` / `session-start-wrap.sh:120`). But `cleanStalePending` and the reclaimer run only inside the daemon (`wrap-daemon.js:856,848`). So with the flag off, `.agentic/wrap-pending-<id>.json` and `.agentic/.heartbeats/<id>` are written every session and nothing ever drains them → unbounded accumulation on every harness, opt-in or not. This also contradicts the planning doc's "with the toggle off … no marker, no daemon" claim (`docs/planning/daemon-driven-deferred-wrap.md:12-13`).

**Required change:** Pick one and make code + docs agree:
- (a) Gate `stageWrapPending` and `touchHeartbeat` on `deferred_wrap_daemon` too, so truly nothing is written when off; **or**
- (b) Keep them flag-independent but run `cleanStalePending` from the Stop hook (or an unconditional lightweight reaper) so markers self-expire even when the daemon never launches.

Then fix the planning doc + `wrap.md` prose: if (b), state that markers/heartbeats are staged regardless of the toggle and are inert without the daemon.

**Acceptance / regression test:** "flag off ⇒ no pending marker accumulates across N sessions" (option a), or "stale pending markers are reaped with the daemon disabled" (option b). `test-stop-context-deferred-wrap.js` tests the loop-guard but not flag-off accumulation.

---

## [Minor] m1 — `hooks/stop-context.js` (5× `// keep in sync with .opencode/plugins/session-context.ts`)

**Problem:** Five "keep in sync" comments assert a parity obligation with the opencode plugin, but that plugin (unmodified in this PR) contains none of the lock-gate/heartbeat/staging logic. The feature is intentionally Claude-only, so the obligation is silently unmet and will mislead the next maintainer.

**Required change:** Drop the comments, or reword to "Claude-only; the opencode adapter intentionally does not implement this."

---

## [Minor] m2 — `bin/agentic-wrap-release-lock:49`

**Problem:** Deletes `<cwd>/.agentic/wrap.lock` for any `cwd` passed as `argv[2]`. No injection or traversal (argv array, `releaseWrapLock`→`safeCwd` rejects non-absolute/dirty paths), so low severity — but it's a PATH-wired bare-name binary that removes a directory based on an arg.

**Required change:** Add a one-line guard asserting the resolved target ends in `.agentic/wrap.lock` (true by construction today), and/or a comment that the caller is trusted. Acceptable to keep as-is; hardening only.

---

## [Minor] m3 — `content/templates/.agentic/config.json` vs `content/rules/conventions.md`

**Problem:** Pre-existing (not introduced here): the template lists `theme_aware`/`storybook_enabled` while conventions documents `motion_aware`/`storybook_version`. The PR's 6 new `deferred_wrap_*` keys are consistent across both; this is an adjacent inconsistency.

**Required change:** Since this PR already touches both files for the toggle count, reconcile the toggle lists in the same change. Optional / not blocking.

---

## [Nit] Security framing — `content/commands/wrap-deferred.md`, `docs/planning/daemon-driven-deferred-wrap.md`

**Problem:** Prose says the child "can't run code / can't release the lock." The `--disallowedTools Bash` boundary stops execution *during the drain* but does not stop the child from *planting* code via Write/Edit (a hook, `.git/hooks/`, `core.hooksPath`) that runs later under normal permissions. And per M1, "can't release the lock" also implies "can't acquire it."

**Required change:** Narrow the wording to "can't run code *during the drain*"; note the deferred write surface (`.agentic/`, `AGENTS.md`, `memory.md`) is broad and unreviewed by design (no Skeptic). The RCE-on-read-only-git-verb class IS genuinely closed — keep that claim.

---

**Not blocking — confirmed good:** `--disallowedTools Bash` is the correct boundary under `bypassPermissions` (allowlist only suppresses prompts); `GIT_CONFIG_*=/dev/null`+`NOSYSTEM=1` is correct defense-in-depth; session-resume is UUID-filtered with validated clean-absolute cwd; symlink hardening (O_NOFOLLOW, lstat guards) is thorough; all 5 new source files carry module manifests; toggle-count doc-sync (Eleven→Twelve) and `wrap-deferred` adapter propagation are consistent; methodology drift check passes.

tristanlee85:
All 6 findings addressed. Pushed `00d385d..469992f`; all 8 wrap test files green (443 assertions, 0 failures).

Note: the deferred-wrap artifacts moved under `.agentic/wrap/` (reorg `bdd0a97`/`00d385d`) after the review, so anchors below were located by symbol, not the review's line numbers.

**M1 (lock) — `fc77fa0`, `52f0cac`.** Lock ownership moved to the daemon. `drainOnce` acquires `acquireWrapLock` after a ready marker is confirmed but before claiming it, and releases via `try/finally` on every post-acquire path. On a held lock it returns `'idle'` (marker stays `ready`, daemon self-exits, no watchdog spin) — the child no longer touches the lock. `wrap-deferred.md` reconciled on both acquire and release sides. New `LOCK-M1` regression test: held lock defers, release drains to a `done` tombstone, no lock dir persists. Also documented the `timeoutMs < reclaimMs` invariant so the daemon's own live lock can't be reclaimed mid-run.

**M2 (ungated accumulation) — `6642c5f`.** `touchHeartbeat`, both `stageWrapPending` calls, and `finalizeReady`/`removeHeartbeat` are now gated on `deferred_wrap_daemon` (option a). Pre-existing tests that asserted staging with no flag were updated to opt in; added flag-off zero-accumulation and flag-on regression cases. No legacy reaper — the feature is unshipped, so there is nothing on real installs to drain.

**m1 — `6642c5f`.** The 5 `keep in sync with .opencode/...` comments now state OpenCode intentionally omits the lock-gate/heartbeat/staging logic (Claude-only).

**m2 — `fc77fa0`.** `bin/agentic-wrap-release-lock` now refuses (fail-open) any resolved path not ending in `.agentic/wrap/lock`.

**m3 — `2762c4f`.** Added `motion_aware`, `storybook_version`, `commit_telemetry` to the config template. Left the separate `capability_preflight_mode` advisory-vs-blocking drift untouched (out of scope) — flagging it for a follow-up.

**Nit — `fc77fa0`.** Narrowed the security prose: Bash removal closes the RCE-via-read-only-git-verb class, but the deferred Write/Edit surface (`.agentic/`, `.git/hooks/`, `core.hooksPath`) is broad and trusted-child-only, unreviewed for adversarial input.

tyson-solara6:
Agent review of this PR (posting findings; not modifying the branch). The design is sound and opt-in-gated correctly - these are the blockers to land it:

**Code bug (real, not just CI):** `hooks/wrap-daemon.js` `appendDecoded()` does not enforce the 256 KB capture cap. It checks `capturedBytes < MAX` but then appends the *entire* incoming chunk, so a chunk arriving near the limit overshoots (the LOG-3 test captures ~321 KB against the intended ≤256 KB). Suggested fix - truncate the chunk at the remaining budget:

```js
const appendDecoded = (s) => {
  if (!s) return;
  if (capturedBytes < wrapMarker.MAX_CHILD_CAPTURE_BYTES) {
    const remaining = wrapMarker.MAX_CHILD_CAPTURE_BYTES - capturedBytes;
    const sBytes = Buffer.byteLength(s);
    if (sBytes <= remaining) {
      captured += s;
      capturedBytes += sBytes;
    } else {
      captured += Buffer.from(s, 'utf8').slice(0, remaining).toString('utf8');
      capturedBytes = wrapMarker.MAX_CHILD_CAPTURE_BYTES;
      capTruncated = true;
    }
  }
  // else: drain-and-discard (listener stays attached; pipe never fills)
};
```
(Minor caveat: slicing a UTF-8 buffer mid-codepoint can emit one replacement char / overshoot by ≤2 bytes - inconsequential for a log, but worth a comment.)

**[Major] `docs/planning/daemon-driven-deferred-wrap/` (7 files) is committed.** `docs/planning/` is gitignored by design (planning artifacts stay local). A CI guard (`no-planning-docs`) now enforces this on main, so this will fail CI. Remove with `git rm -r --cached docs/planning/daemon-driven-deferred-wrap`.

**[Major] Branch is stale vs `main`** - `scripts/.methodology-baseline.sha256` mismatches, so `check-drift` fails. Rebase onto current `main`, then regenerate the baseline and rebuild the 10 adapters.

**[Minor] DCO** - the merge commits lack `Signed-off-by`. Rebase with `-s` (or `git rebase --signoff`).

**[Minor]** Stale "Eleven toggles" cross-reference (the count is now twelve after the new toggle).

tyson-solara6:
One more thing - and to be clear, this is a strong feature, not just a defect list. The daemon design (singleton PID, `in_progress` reclaim, lock ownership, the `pending → ready` invariant, the Bash-removed security loop-guard) is solid work.

A documentation suggestion that ties into the cleanup: when you remove `docs/planning/daemon-driven-deferred-wrap/` (it's gitignored by policy and the new `no-planning-docs` guard enforces it), that folder is currently the only home for the **risk register, rollback procedure, and headless-test findings**. Don't let that walk out the door with the folder.

Consider giving the feature its own **`hooks/wrap-deferred.README.md`** as the committed operator/maintainer home. It serves the audience the other docs don't:
- module manifests → the code-reader (already strong)
- `content/commands/wrap-deferred.md` → the end user invoking it
- a README → the **operator**: how to enable it (the config toggle + the 5 tuning params), what state it owns (`.agentic/wrap/` - PID file, `daemon.log`, markers), how to stop/reset it, the security model in plain terms, and the rollback.

Keep it the operator/integration view and link down to the manifests for code-level invariants rather than restating them (avoids drift). Lifting the condensed risk-register + rollback content into this README before deleting `docs/planning/` keeps the "why" in the repo. A one-line pointer from `wrap-deferred.md` makes it discoverable.

tristanlee85:
Addressed all findings from both review notes; rebased onto current `main` and squashed to one signed-off commit.

**Capture-cap bug:** `appendDecoded` now truncates a cap-crossing chunk on a **code-point boundary** - `captured` never exceeds 256 KB and no `U+FFFD` is introduced (kept the StringDecoder no-replacement invariant rather than the suggested byte-slice). LOG-3 bound tightened (`< 300 KB` -> `<= 256 KB + 2048`); added **LOG-3b** for the multi-byte truncation path. Hook suite: 494 assertions, 0 failed.

**[Major] planning docs:** untracked all 7 `docs/planning/daemon-driven-deferred-wrap` files (the dir's 6 + the sibling `.md` - your "7 files" count was exact); `no-planning-docs` green.

**[Major] stale branch / DCO:** rebased onto current `main`; baseline + 10 adapters regenerated (`check-drift` + `adapter-sync` green). Note on DCO: it was **not** a missing sign-off - `tim-actions/dco` passes the PR's entire commits JSON as one process argument, which exceeded the OS `ARG_MAX` (`Argument list too long`) at 31 verbose SSH-signed commits. Squashed to a single signed-off commit; the sign-offs were all valid.

**[Minor] "Eleven":** verified correct as-is - it's the independent **write-path** count in `stop-context.js` (exactly 11, evidenced per path), not the config-toggle count (already "Twelve" in `conventions.md`). Left unchanged.

**Doc follow-up (your second note):** added `hooks/wrap-deferred.README.md` as the committed operator runbook per the #254 convention - migrates the risk register, rollback, security model, and owned-state inventory out of the local planning folder, with a pointer from `wrap-deferred.md`. Every path/key/default was fact-checked against the code (the old planning docs had stale top-level paths).

Every code/doc change went through engineer + independent Skeptic review.

tyson-solara6:
Re-review after the update - all prior findings resolved. Nice work.

- **Daemon output-cap bug (was the blocker):** fixed. `appendDecoded()` truncates on a code-point boundary at the 256 KB cap and forces drain-and-discard thereafter; LOG-3 bound tightened and a LOG-3b test added verifying no U+FFFD at the truncation point.
- **`docs/planning/` files:** removed (`git ls-files docs/planning/` empty, `check-no-planning-docs` green), and the durable content was migrated into a committed **`hooks/wrap-deferred.README.md`** - exactly the operator home that was missing. 
- **Stale branch / `check-drift`:** rebased onto main, baseline regenerated, adapter-sync green.
- **DCO:** signed.
- **"Eleven toggles":** confirmed a non-issue - the manifest's "Eleven write paths" is a different count from the (correct) "Twelve" config toggles.

Opt-in gating, daemon lifecycle, and the security model (Bash-removed child, `GIT_CONFIG_*=/dev/null`, O_NOFOLLOW on the log) all still hold. All 12 CI checks pass.

Only remaining note (Minor, non-blocking): `deferredDaemonEnabled` is duplicated identically in `hooks/stop-context.js` and `hooks/session-end-wrap.js` with a "keep in sync" comment. The no-import rationale is reasonable; worth extracting to `hooks/lib/wrap-marker.js` if a third call site ever appears.

Sign-off from this review. (Merge is yours to call - I'm only reviewing.)

PR #249, opened 2026-06-18
feat(pioromp): role-model routing with cross-family antagonist reviewers

bloodf:
## What this does

Adds Pi / oh-my-pi role-model routing: `bin/agentic-models` probes `NINEROUTER_URL/v1/models`, ranks models per agentic-engineering role (architect → Opus-class, engineer → Sonnet-class, qa-engineer → Haiku-class, etc.), and writes `~/.agentic/role-models.yml`. A cross-family reviewer pool (Kimi, GPT-5.x, GLM) provides antagonist diversity.

`bin/agentic-configure` is the interactive setup (plain-stdin prompts now; whiptail / dialog TUI is a follow-up PR).

Activation preflight Step 6.5 fires `agentic-configure --non-interactive` on first activation when the harness is Pi/oh-my-pi and `~/.agentic/role-models.yml` is missing. Sentinel pattern mirrors `.agentic/.activated`. Optional `hooks/role-models-bootstrap.py` for harnesses with a UserPromptSubmit hook surface.

Schema accepts scalar `engineer: sonnet` or mapping `{model, effort, reasoning}`. Existing scalar files keep working. Effort and reasoning are pass-through; the conductor forwards only keys that are set.

## How to use it

**Operator guide:** [`docs/role-model-routing.md`](docs/role-model-routing.md) is the entry point. Read this first.

**Three-step setup:**

```bash
# 1. Probe the harness
bin/agentic-models --json                    # one GET to NINEROUTER_URL/v1/models

# 2. Seed ~/.agentic/role-models.yml
bin/agentic-configure                        # interactive wizard reads probe payload

# 3. Verify the resolution
bin/agentic-status                           # shows resolved model per role
bin/agentic-models --suggest engineer        # one-role preview
```

**Minimum useful config:**

```yaml
roles:
  architect: opus
  engineer: sonnet
  qa-engineer: haiku

reviewers:
  strategy: distinct-from-author
  pool: [gpt-5, glm-4.6, kimi-k2.7]
  fallback: gpt-5
```

With this, an Opus engineer produces a diff reviewed by GPT-5; if GPT-5 is unavailable, the strategy falls back to the next pool entry whose model differs from Opus. Reviewers never run on the same model as the author.

**Common patterns:** cost-optimised default (Opus architect, Sonnet engineer, Haiku QA); antagonist-only (just `reviewers:`, no per-role pins); per-task reviewer split via `strategy: by-task` (security → GPT-5, architecture → Opus, correctness → GLM); effort-only pinning (vary `effort` per role, leave model unset). Full examples in [`docs/role-model-routing.md`](docs/role-model-routing.md).

**Schema reference:** [`content/references/role-models.md`](content/references/role-models.md) — full resolution algorithm, interaction with Tier and presets, every supported role key.

**Discovery reference:** [`content/references/model-discovery.md`](content/references/model-discovery.md) — probe protocol, hint heuristics, effort/reasoning field semantics.

**Example library:** [`content/references/role-models-example.yml`](content/references/role-models-example.yml) — copy-paste starter.

## Review fixes (commit 0fc1d4e)

Addressed multi-model review findings:

- **C1** — probe URL now appends `/v1/models` with single trailing-slash strip (was malformed when `NINEROUTER_URL` already ended in `/v1`).
- **M6** — non-interactive configure without a probe URL is a no-op exit 0 (no file written); probe failure exits 2 so the bootstrap sentinel retries on next activation.
- **M2** — harness predicate consolidated into `PI_OMP_HARNESS_ENV_VARS = ("PI_HARNESS", "OMP_HARNESS", "OH_MY_PI_HARNESS", "AGENTIC_HARNESS")` with `any(...)` truthy check, shared verbatim between `hooks/role-models-bootstrap.py` and `bin/agentic-status`.
- **M11** — `--probe-key` default now reads `NINEROUTER_KEY` primary, `NINEROUTER_API_KEY` fallback.
- **M3** — `author_model` field added to task-state schema (`content/sections/08-task-state-file.md`) and recorded at spawn time in `implement-ticket.md` Phase 5.
- Tests rewritten: `test_agentic_configure.py` covers no-probe no-write, probe-failure exit 2, URL append, bootstrap sentinel gating. New `test_agentic_status.py` (5 tests) covers the harness predicate.
- Doc fix: `content/references/model-discovery.md` worked example corrected to `{NINEROUTER_URL}/v1/models`.

Skipped findings (already satisfied in-tree): m1, X2, m2-Kimi.

## Verification

```bash
python3 -m pytest bin/tests --ignore=bin/tests/test_agentic_migrate.py
# 29 passed (migrate test has a pre-existing unrelated failure)

# Targeted checks from the review-fix plan
python3 -m pytest bin/tests/test_agentic_configure.py -k probe_url    # C1
python3 -m pytest bin/tests/test_agentic_configure.py -k no_probe     # M6 no-write
python3 -m pytest bin/tests/test_agentic_configure.py -k with_probe   # M6 positive
python3 -m pytest bin/tests/test_agentic_status.py                    # M2 predicate
bash scripts/check-methodology-drift.sh                               # zero drift
bash scripts/build-slides.sh                                          # all 15 decks
for d in .claude .codex .cursor .gemini .kimi .opencode .omp .pi .hermes .openclaw; do bash $d/build.sh; done
```

## Files (commits in this PR)

- **Feature commits (3):** role-model-routing feature, bootstrap, review fixes.
- **Adapter sync (2):** `c4d521a` and `5468b8c` rebuild all 10 adapters (`.claude`, `.codex`, `.cursor`, `.gemini`, `.kimi`, `.opencode`, `.omp`, `.pi`, `.hermes`, `.openclaw`) so the harness-specific manifests, methods, and skill wiring match `content/`.
- **Slides sync:** `contributing-slides.html` regenerated to reflect the 19-reference-docs count.
- **Docs (1):** `docs/role-model-routing.md` — operator-facing entry point.

## Future direction: where this layer goes next

These are follow-ups, deliberately NOT in this PR. Each is a self-contained surface; ship them when the surface above stabilises.

**1. Reuse the routing layer in QA-tier spawns (qa-engineer, qa-tester).** Today only `skeptic` and `security-auditor` consult the `reviewers:` pool. Extending to qa-engineer would let the runtime gate run on a deliberately different model from the engineer who wrote the diff it is verifying — same antagonist diversity the reviewer pool already provides for code review, applied to runtime checks. Low risk, high signal: the QA gate already exists, the model slot is already typed, and the resolution algorithm is generic. Concrete change: drop the `skeptic / security-auditor` restriction in `content/references/role-models.md` Step 4 and let qa-engineer join the reviewer pool. Wires up naturally as soon as role-model routing is the default for reviewers.

**2. Per-task reviewer split as the default `by-task` mapping.** The `by-task` strategy exists in the schema today but no shipped content pins a default mapping. Once enough operators have run with `distinct-from-author` and formed an opinion on whether GLM does better architecture review than GPT-5, ship a `by_task:` default in `content/references/role-models-example.yml` so the setup wizard can suggest it. This is content, not code; the schema already supports it. Wait until at least 5 operator reports confirm the split beats the pool strategy on quality-vs-cost before promoting the mapping to the example library.

**3. Reuse the routing layer for QA-method selection.** `bin/agentic-models` and `bin/agentic-configure` already know how to probe a harness and rank candidates against role heuristics. The same machinery could rank QA method candidates (perceptual diff, motion emulation, storybook iframe) against harness capabilities (does the model expose screenshots, CDP emulation, storybook story IDs?). Different problem — QA methods are not models — but the probe-and-rank shape transfers cleanly. Defer until the QA gate itself stabilises (DS-20 / PR #239 on code scanning is the more pressing CI gate right now).

**4. Per-session effort curves.** Today `effort` is per-role and static. A future iteration could let the conductor ramp `effort` over the session — `low` for the first few easy tickets, `medium` for routine multi-unit work, `high` only on the largest Brief-tier tasks. The wiring is straightforward (`effort` is already a pass-through field) but the policy question (when to ramp, who decides) is open. Revisit once the Effort / reasoning fields have shipped and operators have a feel for what `effort: high` actually buys them on their model.

**5. Cross-harness role-model library.** Today `role-models.yml` is Pi/omp-only. Claude Code, Codex, and Cursor all accept a model parameter on the Agent tool call; the same YAML could map roles there too with a different resolution rule (Claude Code accepts `model: opus | sonnet | haiku` natively; Cursor uses `model: gpt-5 | sonnet-4 | ...`). The schema is harness-agnostic today; the resolution is the per-harness adaptation. Worth doing once one operator asks for it across two harnesses — do not pre-empt.

**6. Anthropic / OpenAI model-spec probe.** `bin/agentic-models` probes only the OpenAI-compatible `/v1/models` shape. Anthropic does not expose an equivalent endpoint; their model catalog is provider-side. A future iteration could probe Anthropic directly (if/when they ship an OpenAI-compat layer or their own probe protocol) so the setup wizard surfaces Claude candidates without requiring the operator to type them. Defers to Anthropic roadmap.

**7. Bootstrap to Codex / Cursor on first activation.** Today Step 6.5 only fires on Pi/omp. The non-interactive path (`agentic-configure --non-interactive`) is harness-agnostic; extending the activation preflight to Codex and Cursor is a one-line `PI_OMP_HARNESS_ENV_VARS`-style predicate change. Defer — the cost savings on Claude Code (where the user can already type a model in the host UI) are marginal, and getting the Pi/omp story right first is the priority.

## Follow-ups (deliberately not in this PR)

- TUI replacement for `bin/agentic-configure` (whiptail / dialog). Current stdin prompts are functional; polished TUI is a separate PR.
- Schema hardening: add `comments:` and `# yaml` to `role-models.yml` parser so the setup wizard round-trips comments back to the file.

## Out of scope

- Hardcoded model catalogs. The heuristics are substring matches against family names, not exact model IDs; no code change is required when a provider releases a new model. This is a load-bearing design choice and any "let us add a list of known good models" PR will be rejected on principle.
- Per-engine model routing within a single engineer spawn. The conductor picks one model per spawn and that is the whole task.


tyson-solara6:
Heads up: this was reverted on `main` via #253. A post-merge agent review found a Critical runtime bug plus two Major issues. The design, schema, methodology integration, and docs are sound - please resubmit with these fixed:

**[Critical] `bin/agentic-models` has no `main()` function.** The argparse block is indented inside `_print_human(...)`, so the file's trailing `sys.exit(main(sys.argv[1:]))` raises `NameError` on every invocation. CI missed it because `py_compile` sees valid syntax and the unit tests only `importlib` the internal helpers (`_score`/`_rank`/`_suggestions`) - the CLI itself is never run. Fix: extract the argparse block into `def main(argv) -> int:` and add a subprocess-based test that actually invokes `bin/agentic-models --suggest engineer`.

**[Major] `agentic-configure` doesn't fall back to `NINEROUTER_API_KEY`.** `--probe-key` defaults to `NINEROUTER_KEY` only, and configure passes `--probe-key ""` explicitly to `agentic-models`, overriding agentic-models's own fallback. A user with only `NINEROUTER_API_KEY` set gets a probe failure. Fix: `default=os.environ.get("NINEROUTER_KEY") or os.environ.get("NINEROUTER_API_KEY", "")` at the configure call site.

**[Major] `hooks/role-models-bootstrap.py` writes the sentinel on the no-probe-URL `exit 0`.** A user who activates before setting `NINEROUTER_URL` gets the sentinel written and the bootstrap never retries even after they configure the router. Mirror the preflight Step 6.5b check: skip the sentinel write when `NINEROUTER_URL` is unset.

**[Minor] x2:** `_probe_or_empty` is passed the raw `args.probe_url` instead of the normalized local `probe_url`; and the hook's sentinel body is two lines vs the spec's three (cosmetic - only file existence is checked).

Net: revert was to keep `main` clean, not a rejection of the feature. Spin it back up with the `main()` fix + a CLI smoke test and it should be good to go.

PR #290, opened 2026-06-22
fix(install): guard identity.sh pipelines so set -e abort no longer silently kills the installer

tristanlee85:
## Installer silently aborts at "Developer identity..." when no identity is set

`scripts/lib/identity.sh` is sourced by all 10 adapter installers, every one of which runs under `set -euo pipefail`. Two unguarded command substitutions in `_ae_setup_identity` could trip `set -e` and kill the installer mid-function with no error message.

### Symptom

Running `.claude/install.sh` with no developer identity configured printed `Developer identity...` and then returned to the shell - the `Run 'agentic-identity show'` line and every subsequent step (MCP setup, permissions, bin/ symlinking, summary) silently never ran.

### Root cause

- `identity.sh:75` - `grep '^developer_id:'` matches nothing when `agentic-identity show` reports "No identity set" (exit 0, no `developer_id:` line), so grep exits 1. But `awk` (the pipeline's last command) exits 0 - so the abort is only possible under `pipefail`, which surfaces grep's exit 1 as the pipeline's status (without it, awk's 0 would mask grep's failure and the assignment would succeed). Under `pipefail` that non-zero status propagates to the bare assignment and `set -e` aborts. **Fires whenever no identity exists yet** (the common fresh-install case), and only in the installers' `set -euo pipefail` context.
- `identity.sh:131` - the manual-handle prompt's `xargs` pipeline exits 1 on an unmatched quote in user input, aborting the same way and defeating the handle-validation that follows. **Fires on the `gh`-absent fallback path** when a user types a handle containing a stray quote.

### Fix

Append the same `|| var=""` guard already used on the two adjacent substitutions (lines 73, 106). The existing `[[ -n ]]` / `[[ -z ]]` checks handle the empty case gracefully. Restores the module manifest's "Never aborts the caller" contract. Also reconciles the manifest's downstream-consumer list to the actual 10 adapters (adds `.openclaw`).

The manual-prompt path additionally distinguishes a deliberate skip (empty input) from an unparseable handle (the `xargs` guard fired): the latter now prints "typed handle could not be parsed, skipping identity setup" instead of the generic skip message, so a user whose input was silently dropped is told why.

Single shared source - all 10 adapters `source` it directly, so the change covers every adapter.

### Verification

Isolated bash harness under `set -euo pipefail`, function called as a bare statement (matching the real call site at `.claude/install.sh:625`), with a stubbed `agentic-identity`:

- no-identity input → before: aborts (exit 1); after: completes, all follow-up steps run (exit 0)
- malformed-quote handle → before: aborts (exit 1); after: graceful "could not be parsed" skip (exit 0)
- valid / empty / uppercase handles → behave correctly (exit 0), normalization intact

No committed shell regression test added - no shell-test harness sources this lib (the only identity test exercises the Python binary), and scaffolding a new framework for a one-line guard is out of scope.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01UYDKXD78xTbNwkYScE5T2x

### DIFF
--- a/scripts/lib/identity.sh
+++ b/scripts/lib/identity.sh
@@ -72,7 +72,7 @@ _ae_setup_identity() {
   local show_out
   show_out="$(agentic-identity show --scope effective 2>/dev/null)" || show_out=""
   local existing_handle
-  existing_handle="$(echo "$show_out" | grep '^developer_id:' | awk '{print $2}')"
+  existing_handle="$(echo "$show_out" | grep '^developer_id:' | awk '{print $2}')" || existing_handle=""
   if [[ -n "$existing_handle" ]]; then
     if echo "$show_out" | grep -q 'provisional:'; then
       echo "  = identity already set to '$existing_handle' (provisional - run 'agentic-identity confirm' to lock it in)"
--- a/scripts/lib/identity.sh
+++ b/scripts/lib/identity.sh
@@ -128,10 +128,16 @@ _ae_setup_identity() {
   # Branch 7: gh absent or unauthenticated - prompt manually
   echo "  Developer identity links telemetry to your handle across sessions."
   local typed_handle=""
+  local raw_handle=""
   read -r -p "  GitHub handle [skip]: " typed_handle </dev/tty || typed_handle=""
+  raw_handle="$typed_handle"
   typed_handle="$(echo "$typed_handle" | xargs | tr '[:upper:]' '[:lower:]')" || typed_handle=""
   if [[ -z "$typed_handle" ]]; then
-    echo "  - identity setup skipped (run 'agentic-identity init <handle>' later)"
+    if [[ -n "${raw_handle//[[:space:]]/}" ]]; then
+      echo "  - typed handle could not be parsed, skipping identity setup (run 'agentic-identity init <handle>' later)"
+    else
+      echo "  - identity setup skipped (run 'agentic-identity init <handle>' later)"
+    fi
     return
   fi
   if ! echo "$typed_handle" | grep -qE '^[a-z0-9._-]{1,64}$'; then
--- a/scripts/lib/identity.sh
+++ b/scripts/lib/identity.sh
@@ -19,7 +19,8 @@
 # Downstream consumers:
 #   .claude/install.sh, .codex/install.sh, .cursor/install.sh,
 #   .gemini/install.sh, .hermes/install.sh, .kimi/install.sh,
-#   .omp/install.sh, .opencode/install.sh, .pi/install.sh
+#   .omp/install.sh, .openclaw/install.sh, .opencode/install.sh,
+#   .pi/install.sh
 #
 # Failure modes:
 #   Never aborts the caller - every agentic-identity call captures rc;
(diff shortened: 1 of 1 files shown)


tyson-solara6 (approved):
**Automated review - identity.sh pipeline guards**

Reviewed the two `|| var=""` guards and manifest update. No blocking issues.

**Fix correctness**: Both guards are correct for the `set -euo pipefail` context the adapter installers run under. Under `pipefail`, a `grep | awk` pipeline where `grep` exits 1 (no match) propagates that exit code even though `awk` exits 0 - the assignment fails and `set -e` aborts. The `|| var=""` idiom matches the pattern already established on lines 73 and 106 of the same file.

**Regression risk**: Low. Line 73 already guards the `agentic-identity show` call, so `show_out` is always `""` when the binary is absent - `existing_handle`'s new guard only fires on the grep-no-match path, which is exactly the "no identity set" case.

**Manifest update**: Accurate - adds `.openclaw/install.sh` to the downstream consumers list.

**Minor findings (non-blocking):**

1. **PR description imprecision (line 75)**: The description frames root cause 1 as "grep exits 1 → set -e aborts" without mentioning that `pipefail` is required for this to happen (`awk` would otherwise mask grep's exit code). The fix is correct for the callers' `set -euo pipefail` context; the description is just incomplete. No code change needed.

2. **Silent handle drop (line 131)**: When `xargs` fails on unmatched input (e.g. `o'brian`), the `|| typed_handle=""` guard discards the partial output and silently skips identity setup with no explanation. This is strictly better behavior than aborting the installer, but a follow-up improvement could add a warning message when the guard fires, e.g. `"  - typed handle could not be parsed, skipping identity setup"`. Not a blocker.

CI: 13/13 checks pass. Approving.

tristanlee85:
Both Minor findings addressed.

- **Silent handle drop** (Branch 7 manual prompt): fixed in f2a7a5cb. The path now captures the raw input before the `xargs | tr` transform and, when the guard fires on unparseable input, prints `- typed handle could not be parsed, skipping identity setup (run 'agentic-identity init <handle>' later)` instead of the generic skip. The load-bearing `|| typed_handle=""` guard is unchanged. Verified under `set -euo pipefail` (incl. bash 3.2.57): empty input -> original skip message; `o'brian` -> new parse-failure message; valid/mixed-case handle -> normalizes and proceeds; `bad handle!` still falls through to the validity check. All exit 0, no abort.
- **Description precision** (root cause 1): updated. It now states the abort is only possible under `pipefail` - `awk` is the pipeline's last command and exits 0, so without `pipefail` it would mask grep's exit 1 and no abort would occur; the bug is specific to the installers' `set -euo pipefail` context.

tyson-solara6 (approved):
**Approving** - new commits since last review covered.

The two pipeline guards are correct and consistent with the existing pattern (lines 73 and 106 use the same idiom). Specific notes:

-  (line 75): Under `set -euo pipefail`, grep exits 1 when no identity exists (fresh install). The awk exit code masks grep's in the pipeline without `pipefail`, but with it grep's failure propagates - this guard restores the assignment-to-empty semantics correctly.
- `raw_handle` / `typed_handle=... || typed_handle=""` (lines 128-137): The distinction between deliberate skip and xargs parse failure is a genuine UX improvement. The `${raw_handle//[[:space:]]/}` check correctly identifies non-empty-but-unparseable input (pure-whitespace input correctly falls through to the "skipped" branch). No edge cases found.
- Manifest downstream consumer update (.openclaw added): correct.

CI: all checks pass. Mergeable with no conflicts.

Issue #258, opened 2026-06-19
Deep audit (commit f8ad222): index of findings + rejected items


bloodf:
## Context

Deep audit of the code in `bin/` (12 Python CLIs, 2.4K LOC), `hooks/` (Node.js + Python enforcement scripts, 1.6K LOC), and `scripts/` (shell build pipeline, 195 LOC). Performed against commit `f8ad222`.

The methodology prose under `content/` (~6,500 LOC of Markdown) was cross-checked for drift against code, but was NOT line-audited for internal inconsistencies - that is a separate concern.

## Verification gates (apply to every plan below)

| Purpose | Command | Expected |
|---|---|---|
| Python tests | `python3 bin/tests/test_<name>.py` | exit 0 |
| JS tests | `node hooks/tests/test-<name>.js` | exit 0 |
| Shell tests | `bash bin/tests/test_<name>.sh` | exit 0 |
| Methodology drift | `bash scripts/check-methodology-drift.sh` | exit 0 |
| Adapter sync | `bash .claude/build.sh && git diff --exit-code -- .claude` | no diff |
| Slides sync | `bash scripts/build-slides.sh && git diff --exit-code -- docs/slides` | no diff |
| Secret scan | `gitleaks git . --redact=100` | no findings |

No `pytest.ini`, `pyproject.toml`, `mypy.ini`, `ruff.toml`, `tsconfig.json`, or `.eslintrc` at root or depth <=2 today. CodeQL is enabled; it is not a style/type checker.

## Trust model (do NOT report as findings)

Per `SAFETY.md`: the Bash deny-list in `.claude/install.sh` is explicitly a "finite rail, not a boundary". Hooks fail open by design. No host sandboxing. Subprocess bypass of the deny-list is a documented limitation, not a bug.

## Dependency ordering between issues

- **#2 (lint config) should land before #5 (stop-context read caching).** Refactoring without a type checker is how refactors regress.
- **#1 (test coverage) should land before #4, #5, #7, #9** - all four modify currently-untested code.
- **#3 (README drift) is independent** - doc-only PR.
- **#10 (telemetry health) pairs naturally with #5** - both touch stop-context.js read paths.
- **#8 (migrate containment), #11 (marp bump), #12 (AGENTS.md) are all independent** - parallel-safe.

## Issues in this batch

| # | Title | Effort | Risk |
|---|---|---|---|
| 1 | Add tests for `agentic-memory` and 4 other untested CLIs | M | LOW |
[...]

PR #331, opened 2026-06-26
perf(identity): single-read dedup in flushPendingBuffer, O(M*N) -> O(M+N) (#268) [bin-stack 1/3]

### LINKED ISSUE #268: Fix flushPendingBuffer O(M*N) dedup scan in agentic-identity
## Problem

`bin/agentic-identity` `flushPendingBuffer` does a linear substring scan of the entire global session log **inside** the per-pending-file loop. With M pending files and N lines in the global log, the cost is O(M*N) under an exclusive `fcntl.flock` that serializes all callers.

At 20 pending files x 50K-line log = ~1M substring operations. The 30s lock timeout absorbs it today, but delay accumulates under load and the lock blocks every concurrent `agentic-identity confirm`.

## Evidence

`bin/agentic-identity:233-247` (the dedup loop):

```python
# Dedup: scan global log for this session_uuid.
global_log = GLOBAL_SESSION_LOG_DIR / f"{confirmed_dev_id}.jsonl"
already_flushed = False
if session_uuid and global_log.is_file():
    try:
        if session_uuid in global_log.read_text(encoding="utf-8"):
            already_flushed = True
    except OSError:
        pass
```

This `read_text` + `in` substring check happens once per pending file. The file grows monotonically.

## Goal

Read the global log **once**, before the loop. Build a set of seen `session_uuid` values. Check membership per pending file in O(1). Single parse pass instead of M passes.

Preserve the existing semantics exactly:
- If a session_uuid appears in the global log, the pending file is unlinked and skipped.
- If the global log is unreadable, fall back to flushing (current behavior treats OSError as "not already flushed").

## Repo conventions to match

- Stdlib only - no new deps.
- Atomic writes, flock discipline already established in the file.
- Module-level docstring documents failure modes - update it.
- There is an existing test file: `bin/tests/test_agentic_identity.py`. Extend it - do NOT create a new test file.

## Steps

1. **Read `bin/agentic-identity` lines 165-324** end-to-end. Understand the full `flushPendingBuffer` function.
2. **Read `bin/tests/test_agentic_identity.py`** - find the existing `flushPendingBuffer` tests (search the file for `flushPendingBuffer` or `flush_pending`). Understand the fixtures.
3. **Refactor the dedup**:

   Before the `for pending_path in pending_files:` loop (around line 215), add a single-read set build:

   ```python
   # Build a set of session_uuids already in the global log (single read).
   # Falls back to empty set on read failure - current behavior treats OSError
   # as "not already flushed" so we proceed with flushing.
   seen_uuids: set[str] = set()
   if global_log.is_file():
       try:
           for line in global_log.read_text(encoding="utf-8").splitlines():
               # Cheap substring pre-filter, then exact field extraction.
               # JSON lines are small; this parse is O(N) once, not O(M*N).
               line = line.strip()
               if not line or "session_uuid" not in line:
                   continue
               try:
                   obj = json.loads(line)
                   uuid_val = obj.get("session_uuid")
                   if isinstance(uuid_val, str) and uuid_val:
                       seen_uuids.add(uuid_val)
               except json.JSONDecodeError:
                   continue
       except OSError:
           # Could not read global log - fall back to "flush everything" behavior.
           seen_uuids = set()
   ```

   Then inside the loop, replace the per-file `read_text` + `in` check:

   ```python
   already_flushed = bool(session_uuid) and session_uuid in seen_uuids
   ```

4. **Update the module docstring** at the top of `bin/agentic-identity`. The "Performance" line currently says "Standard. File I/O only". Update to mention: "flushPendingBuffer: O(N) single-pass dedup against the global log regardless of pending-buffer size."

5. **Add tests** to `bin/tests/test_agentic_identity.py`:

   - Test that flushing N pending files where some are already in the global log: verify already-flushed files are unlinked, new files are appended, global log gets exactly the new lines.
   - Test the OSError fallback: patch `global_log.is_file` to return True but `read_text` to raise OSError - verify all pending files flush (current behavior preserved).
   - Performance smoke test (optional but recommended): create 50 pending files and a 10K-line global log, verify flush completes in under 2 seconds.

## Files in scope

Modify:
- `bin/agentic-identity` (refactor `flushPendingBuffer` dedup)
- `bin/tests/test_agentic_identity.py` (add tests)

- Do NOT change the file-write paths, lock semantics, or attribution logic. Only the dedup lookup changes.
- Do NOT introduce a SQLite or JSONL index file. The set is rebuilt on each call - that is fine because confirm is rare.
- Do NOT touch `hooks/stop-context.js` - the JS side has its own dedup in `writePendingBuffer` cap-100 logic, unrelated.

```
# Existing tests pass
python3 bin/tests/test_agentic_identity.py

# New tests pass
python3 bin/tests/test_agentic_identity.py  # (same command - new tests are in the same file)

# Functional smoke test
cd /tmp && mkdir -p test-flush/.agentic/session-log/.pending
echo '{"session_uuid":"abc","ts":"2024-01-01T00:00:00Z","repo_root":"/tmp/test-flush","project_slug":"test-flush"}' \
  > test-flush/.agentic/session-log/.pending/abc.json
python3 /path/to/repo/bin/agentic-identity init testuser --force
# Verify: ~/.agentic/session-log/testuser.jsonl contains the abc line
grep -q '"session_uuid":"abc"' ~/.agentic/session-log/testuser.jsonl

# Re-running flush with the same pending file should skip (dedup)
echo '{"session_uuid":"abc","ts":"2024-01-01T00:00:00Z","repo_root":"/tmp/test-flush","project_slug":"test-flush"}' \
  > test-flush/.agentic/session-log/.pending/abc.json
python3 /path/to/repo/bin/agentic-identity confirm
# Verify: global log has exactly ONE abc line (dedup worked)
test "$(grep -c '"session_uuid":"abc"' ~/.agentic/session-log/testuser.jsonl)" = "1"
```

## Escape hatches

- If the existing tests use `unittest.mock.patch` on `GLOBAL_SESSION_LOG_DIR` in a way that breaks the single-read pattern, adapt the tests - the production code change should stand.
- If the global log is so large that even a single read is too slow (e.g. multi-GB), STOP and report back - the fix would need incremental bookkeeping, not an in-memory set. This is unlikely; session logs are per-developer and ~MB-scale.

## Maintenance note

Reviewers: the dedup logic is idempotence-critical - a bug here means duplicate session-total lines in the global log, which corrupts `agentic-cost` aggregates. Verify the dedup test covers the "already flushed" path specifically. Future changes that add fields to the attributed line should not affect this code - it only cares about `session_uuid`.

bloodf:
> **Audit batch #258 — bin-stack 1/3.** Merge order: **#331 → #332 → #333** (all touch `bin/agentic-identity` / `bin/agentic-migrate`). This is the base of the stack — merge first. #332 and #333 rebase onto `main` if this squash-merges (SHA changes).

Fixes the O(M*N) dedup scan in `flushPendingBuffer`. Closes #268.

- Global session log is read **once** before the per-pending-file loop into a `seen_uuids` set; per-file dedup is now O(1) membership instead of a `read_text()` + substring scan per file. Total O(M+N).
- `fcntl.flock` scope unchanged — the read sits inside the existing lock (no TOCTOU).
- Dedup is now JSON-field-extraction (`session_uuid` per line) rather than raw substring match: marginally stricter (no false-positive skips on substring coincidences), never looser. Malformed lines contribute nothing; worst case one self-healing re-flush (the re-written line is well-formed and dedups on the next run).
- Fallback preserved: unreadable/missing global log → empty set → everything flushes.

Tests: 4 new dedup regression tests (skip-already-flushed, flush-new, missing-log fallback, multi-pending correctness); full file green (`python3 bin/tests/test_agentic_identity.py`).

### DIFF
--- a/bin/agentic-identity
+++ b/bin/agentic-identity
@@ -41,12 +41,17 @@ Failure modes: exits 1 on validation error, 2 on file-exists-without-force.
                30s timeout via LOCK_NB+sleep loop (SIGALRM avoided for portability);
                on timeout prints stderr warning and returns without data loss.
                flushPendingBuffer dedup: global log is read ONCE before the
-               per-pending-file loop into a set of seen session_uuid values
-               (O(M+N) total instead of O(M*N)). On OSError the set is empty
-               and all pending files are flushed (same fallback as before).
-               Membership check uses substring scan of the raw log text split
-               by lines, preserving exact dedup precision of the prior
-               `session_uuid in full_text` check (no precision change).
+               per-pending-file loop; each line is JSON-parsed and the
+               session_uuid field is collected into a set (O(M+N) total
+               instead of O(M*N)). Per-pending membership check is O(1).
+               Malformed (non-JSON) lines contribute nothing to the set; at
+               most one self-healing re-flush is possible, and because the
+               re-written line is well-formed JSON it dedups on the next run.
+               This is marginally STRICTER than the prior `session_uuid in
+               full_text` substring check (no false-positive skips on
+               substring coincidences in other fields), never looser.
+               On OSError the set is empty and all pending files are flushed
+               (same fallback as before).
                flushPendingBuffer with repo_root_filter: records whose repo_root
                does not match the filter are skipped (left in the buffer, not
                unlinked); records with no/empty repo_root are also skipped under
@@ -239,9 +244,8 @@ def flushPendingBuffer(confirmed_dev_id: str, repo_root_filter: str | None = Non
                         if _uuid:
                             seen_uuids.add(_uuid)
                     except (json.JSONDecodeError, AttributeError):
-                        # Malformed line: fall back to substring scan so we
-                        # never become LESS strict than the old `in text` check.
-                        # Extract any 36-char UUID-shaped token as a best effort.
+                        # Malformed line: contributes no uuid. Worst case one
+                        # self-healing re-flush (next run's well-formed line dedups it).
                         pass
             except OSError:
                 pass  # Empty set -> flush everything (fallback preserved).
--- a/bin/agentic-identity
+++ b/bin/agentic-identity
@@ -227,9 +227,9 @@ def flushPendingBuffer(confirmed_dev_id: str, repo_root_filter: str | None = Non
         # Dedup: read global log ONCE before the per-pending-file loop.
         # Build a set of session_uuid strings seen in the log so each
         # per-file check is O(1) instead of re-reading the whole file.
-        # Precision: split on newlines then substring-scan each line,
-        # which is at least as precise as the prior full-text `in` check
-        # (a uuid that spans a line boundary would be missed by both).
+        # Precision: each line is JSON-parsed and its session_uuid field collected,
+        # stricter than the prior full-text `in` check (no substring-coincidence
+        # false-positive skips). Malformed lines contribute nothing.
         # On OSError the set is empty -> nothing is treated as already-flushed
         # -> all pending files flush (same fallback semantics as before).
         global_log = GLOBAL_SESSION_LOG_DIR / f"{confirmed_dev_id}.jsonl"
--- a/bin/agentic-identity
+++ b/bin/agentic-identity
@@ -40,6 +40,13 @@ Failure modes: exits 1 on validation error, 2 on file-exists-without-force.
                flushPendingBuffer: serialized via fcntl.flock on .flush.lock;
                30s timeout via LOCK_NB+sleep loop (SIGALRM avoided for portability);
                on timeout prints stderr warning and returns without data loss.
+               flushPendingBuffer dedup: global log is read ONCE before the
+               per-pending-file loop into a set of seen session_uuid values
+               (O(M+N) total instead of O(M*N)). On OSError the set is empty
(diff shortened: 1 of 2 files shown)


tyson-solara6 (approved):
**Approving** - correct performance fix with regression tests, all CI green.

The O(M*N) -> O(M+N) refactor is clean:

- **Lock safety**: the global log read sits inside the existing `fcntl.flock` scope, so concurrent writers are excluded during the snapshot. No TOCTOU introduced.
- **Empty uuid guard**: `bool(session_uuid and session_uuid in seen_uuids)` correctly short-circuits on empty string (returns False, pending file flushes). No behavior change vs the old code.
- **Stricter dedup claim holds**: JSON field extraction cannot false-positive-skip on a uuid that appears as a value in another field (e.g., in a `note` string). The old substring check could. Never looser.
- **Fallback preserved**: OSError on log read -> empty set -> all pending files flush (same semantics as before).
- **Tests**: the 4 new regression cases (already-flushed skip, new-uuid flush, missing-log fallback, multi-pending mixed) cover the critical paths. The test helpers correctly patch the path constants used by the production code.

Note: this is bin-stack 1/3 (merge before #332 and #333).

PR #256, opened 2026-06-19
feat(agents): cross-harness agent teams (any harness) + role-model routing

bloodf:
## Summary

A **harness-agnostic** multi-model agent process. Two layered, opt-in capabilities, plus the cleanup that makes them fully harness-native:

1. **Cross-harness agent teams (any harness)** — the conductor on **any** harness (Claude, Codex, Gemini, Kimi, Cursor, Pi, omp, …) dispatches *other CLIs* (codex, gemini, cursor-agent, kimi, pi, omp, claude) as leaf workers, config + discovery driven, self-contained. This is the headline capability: it works the same regardless of which harness you run. *(Folded in from #294.)*
2. **Role -> model routing** — pin a concrete model per methodology role and per adversarial reviewer, with cross-family reviewer diversity. This narrower layer is **Pi/omp-only** by design (on Claude/Codex/Gemini the harness Tier mechanism already covers it). *(Re-lands #249; all five revert reasons fixed with regression tests.)*

Both are **off by default**. Neither requires any account, URL, or key beyond the login/subscription you already use in your harness — the AI harness owns auth and model access.

```
┌──────────────────────────────────────────────────────────────────┐
│  LAYER 2  cross-harness teams   role → (harness, model)            │
│           any conductor harness  ·  bin/agentic-team · team.yml    │
├──────────────────────────────────────────────────────────────────┤
│  LAYER 1  role-model routing    role → model   (Pi/omp only)       │
│           bin/agentic-configure · bin/agentic-models · role-models.yml │
└──────────────────────────────────────────────────────────────────┘
```

---

## How you select models (the important part)

**The AI harness owns auth and model access.** You authenticate with whatever you already use — Claude subscription, Codex/OpenAI plan, Gemini plan, Kimi, Pi login. The methodology never holds a router URL, an API key, or a model catalog. There is **no network probe** and **no required env var.**

Model selection has **three paths** — use whichever fits your setup (full reference: `content/references/model-discovery.md`, operator guide: `docs/role-model-routing.md`):

1. **Ask-user (the wizard).** Run `bin/agentic-configure` (Layer 1) or `bin/agentic-team configure` (Layer 2). It prompts you role by role, ranks the model names *you provide* against built-in role hints (opus-class -> architect/security, sonnet-class -> engineer, cross-family -> reviewer pool), and writes a starter config you can edit.
[...]

PR #332, opened 2026-06-26
fix(migrate): add path-containment guard on manifest writes (#269) [bin-stack 2/3]

### LINKED ISSUE #269: Add path-containment guard in agentic-migrate manifest writes
## Problem

`bin/agentic-migrate` accepts a user-supplied manifest path (via `--manifest PATH` or `AGENTIC_MANIFEST_PATH` env var) and writes files to locations derived from `manifest["files"][i]["path"]` joined to `project_root`. There is no path-containment check - a manifest entry with `"path": "../../../../etc/cron.d/backdoor"` would write outside `project_root`.

Realistic exposure is LOW - the user supplying the manifest is the same user running the CLI, so it is not a privilege escalation. But it is the one path in the codebase where untrusted-shaped data crosses into filesystem writes without a containment guard. Adding the guard is a 5-line fix and makes the contract explicit.

## Evidence

`bin/agentic-migrate:465-475` (the file-copy loop):

```python
# --- file rules ---
for entry in manifest.get("files", []):
    try:
        target = project_root / entry["path"]
        if not target.exists():
            seed_path = manifest_dir / entry["seed"]
            if seed_path.is_file():
                target.parent.mkdir(parents=True, exist_ok=True)
                shutil.copy2(str(seed_path), str(target))
                wrote_anything = True
            else:
                errored = True
    except Exception:
        errored = True
```

`target = project_root / entry["path"]` - no containment check.

The manifest source resolution at `bin/agentic-migrate:39-43`:

```python
candidates = [
    os.environ.get("AGENTIC_MANIFEST_PATH"),
    os.path.expanduser("~/.claude/skills/agentic-engineering/project-scaffolding.yml"),
    str(SCRIPT_DIR.parent / "content" / "project-scaffolding.yml"),
]
```

The first candidate is an environment variable - user-controlled.

## Goal

Add a `Path.resolve().is_relative_to(project_root.resolve())` containment check before any file write. If the check fails, set `errored = True` and skip the write. Optionally: print a stderr warning naming the offending path.

## Repo conventions to match

- `Path.is_relative_to()` is Python 3.9+. The file already uses `from __future__ import annotations` and 3.9+ syntax - verify in `bin/agentic-migrate` header.
- Silent-fail discipline on migration writes: never raise to the caller, set `errored = True` and return exit code 3 at the end (existing pattern).
- Stdlib only.

## Steps

1. **Read `bin/agentic-migrate` lines 455-500** - the full `cmd_apply` function body around the file-copy loop.

2. **Add a containment check** before the `target.parent.mkdir` call:

   ```python
   # --- file rules ---
   for entry in manifest.get("files", []):
       try:
           target = project_root / entry["path"]
           # Path containment: target must resolve inside project_root.
           # Prevents manifest entries like "../../../etc/passwd" from escaping.
           try:
               target_resolved = target.resolve()
               project_root_resolved = project_root.resolve()
               if not target_resolved.is_relative_to(project_root_resolved):
                   print(
                       f"agentic-migrate: refusing to write outside project root: "
                       f"{entry['path']} -> {target_resolved}",
                       file=sys.stderr,
                   )
                   errored = True
                   continue
           except (OSError, ValueError) as exc:
               print(
                   f"agentic-migrate: path resolution failed for {entry['path']}: {exc}",
                   file=sys.stderr,
               )
               errored = True
               continue

           if not target.exists():
               seed_path = manifest_dir / entry["seed"]
               # ... rest unchanged ...
   ```

3. **Apply the same check to the seed_path** if the manifest could point at arbitrary seed files. Verify whether `manifest_dir` is derived from the manifest path (which is user-supplied) - if so, the seed copy source should also be contained OR explicitly trusted. Read `bin/agentic-migrate:431` (`manifest_dir = manifest_path.parent`) to confirm. If `manifest_dir` is the user-supplied manifest's parent directory, seed paths are already user-trusted (same trust as running the CLI). Leave seed_path alone - only contain the write target.

4. **Add a test** to `bin/tests/test_agentic_migrate.py`:

   - Create a manifest YAML with a file entry whose `path` is `../../../../tmp/should_not_exist`.
   - Run `agentic-migrate apply --manifest <path> --project-root <tmpdir>`.
   - Verify: exit code is 3 (error), the target file does NOT exist outside tmpdir, stderr contains "refusing to write outside project root".

## Files in scope

Modify:
- `bin/agentic-migrate` (add containment check)
- `bin/tests/test_agentic_migrate.py` (add test)

- Do NOT add containment to the pending-buffer path in `bin/agentic-identity` - that is a different trust boundary (local-only files, see index issue #258 rejected items).
- Do NOT refactor the manifest loading logic.
- Do NOT change the exit codes.

```
# Existing tests pass
python3 bin/tests/test_agentic_migrate.py

# New containment test
python3 bin/tests/test_agentic_migrate.py  # (new test is in the same file)

# Manual: a malicious manifest is rejected
mkdir -p /tmp/migrate-test/{project,manifest-dir}
cat > /tmp/migrate-test/manifest-dir/evil.yml <<YML
scaffolding_version: "99.0"
files:
  - path: "../../../../etc/should_not_exist"
    seed: "seed.txt"
YML
echo "content" > /tmp/migrate-test/manifest-dir/seed.txt
python3 bin/agentic-migrate apply \
  --manifest /tmp/migrate-test/manifest-dir/evil.yml \
  --project-root /tmp/migrate-test/project
# Verify: exit code is 3, /etc/should_not_exist does not exist
test ! -f /etc/should_not_exist && echo "PASS: no escape"
```

## Escape hatches

- If `Path.is_relative_to` is not available (Python < 3.9), use `os.path.commonpath([target_resolved, project_root_resolved]) == str(project_root_resolved)`. But the repo already targets 3.9+ - verify and prefer the built-in.
- If the manifest format has changed since this issue was filed, grep for `manifest.get("files"` to find the current write loop and apply the check there.

## Maintenance note

Reviewers: this is defense-in-depth, not a critical fix. The trust boundary is "user supplies their own manifest" which is equivalent to "user runs their own code". But the guard makes the contract explicit and protects against a future where manifests come from a less-trusted source (e.g. a shared template registry). Do not weaken the check in future refactors.

bloodf:
> **Audit batch #258 — bin-stack 2/3.** Merge order: #331 → **#332** → #333. Merge **after #331**. Already rebased onto current `main`; re-rebase if #331 squash-merges.

Adds a path-containment guard to `agentic-migrate` manifest writes. Closes #269.

- Before any manifest-derived write in `cmd_apply`, checks `target.resolve().is_relative_to(project_root.resolve())`. Out-of-root entries set `errored=True`, print a stderr warning, and `continue` (skip) — the `continue` precedes `mkdir`/`copy2`, so no directory or file is created outside root.
- Catches relative (`../`), absolute (`/etc/...` — pathlib absolute-reset), and symlink-escape paths. The gitignore loop writes a fixed path (pattern is content, not a path) and is not an attack surface; no other `entry["path"]`-derived write exists.
- Exit code 3 (existing partial-apply path); other valid entries still process. Module docstring updated.

Tests: `TestPathTraversalGuard` — relative + absolute traversal cases, each asserting exit 3, the out-of-root file is NOT created, and stderr names the path. Mutation-checked (guard removed → test fails). 15/15 pass, rebased onto current main (keeps #256's test additions).

### DIFF
--- a/bin/agentic-migrate
+++ b/bin/agentic-migrate
@@ -13,7 +13,8 @@
 #   0 = success / no-op
 #   1 = drift detected (check subcommand only)
 #   2 = manifest not found / parse error
-#   3 = partial apply (some rules errored)
+#   3 = partial apply (some rules errored; also raised when a manifest entry's
+#       path resolves outside project_root - path-traversal guard)
 
 from __future__ import annotations
 
@@ -462,9 +463,17 @@ def cmd_apply(args: argparse.Namespace) -> int:
                 errored = True
 
         # --- file rules ---
+        resolved_root = project_root.resolve()
         for entry in manifest.get("files", []):
             try:
                 target = project_root / entry["path"]
+                if not target.resolve().is_relative_to(resolved_root):
+                    print(
+                        f"agentic-migrate: skipping out-of-root path: {entry['path']!r}",
+                        file=sys.stderr,
+                    )
+                    errored = True
+                    continue
                 if not target.exists():
                     seed_path = manifest_dir / entry["seed"]
                     if seed_path.is_file():
--- a/bin/tests/test_agentic_migrate.py
+++ b/bin/tests/test_agentic_migrate.py
@@ -430,5 +430,79 @@ markers: []
         # 4. No crash (result.returncode already checked above)
 
 
+class TestPathTraversalGuard(unittest.TestCase):
+    """Manifest entries with traversal paths must not write outside project_root; exit 3."""
+
+    def _make_fixture(self):
+        """Return (tmp, project, manifest_dir) with a minimal project scaffold."""
+        tmp = tempfile.mkdtemp()
+        project = Path(tmp) / "project"
+        project.mkdir()
+        agentic = project / ".agentic"
+        agentic.mkdir()
+        (agentic / "config.json").write_text(json.dumps({"debugger_on_failure": False}) + "\n")
+        (project / ".gitignore").write_text("")
+        manifest_dir = Path(tmp)
+        (manifest_dir / "innocent.json").write_text("{}\n")
+        return tmp, project, manifest_dir
+
+    def test_relative_traversal_blocked(self):
+        """../escape.txt must not be written outside project_root."""
+        tmp, project, manifest_dir = self._make_fixture()
+
+        manifest_text = """
+scaffolding_version: 1
+gitignore: []
+files:
+  - path: "../escape.txt"
+    seed: "innocent.json"
+    purpose: "relative traversal attempt"
+markers: []
+"""
+        manifest_path = manifest_dir / "traversal-manifest.yml"
+        manifest_path.write_text(manifest_text)
+
+        result = run(
+            ["apply", "--manifest", str(manifest_path), "--project-root", str(project)],
+        )
+
+        self.assertEqual(result.returncode, 3, msg=f"Expected exit 3, got {result.returncode}\n{result.stderr}")
+        escaped = project.parent / "escape.txt"
+        self.assertFalse(escaped.exists(), "Traversal target must not be written outside project_root")
+        self.assertIn("escape.txt", result.stderr, "stderr must mention the offending path")
+
+    def test_absolute_path_blocked(self):
+        """An absolute path outside project_root must not be written."""
+        tmp, project, manifest_dir = self._make_fixture()
+
+        # Use a predictable temp path that is clearly outside project
+        import tempfile as _tf
+        target_dir = Path(_tf.mkdtemp())
+        absolute_target = str(target_dir / "agentic-escape-test.txt")
+
+        manifest_text = f"""
+scaffolding_version: 1
+gitignore: []
+files:
+  - path: "{absolute_target}"
+    seed: "innocent.json"
+    purpose: "absolute path attack"
+markers: []
+"""
+        manifest_path = manifest_dir / "absolute-manifest.yml"
+        manifest_path.write_text(manifest_text)
+
+        result = run(
+            ["apply", "--manifest", str(manifest_path), "--project-root", str(project)],
+        )
+
+        self.assertEqual(result.returncode, 3, msg=f"Expected exit 3, got {result.returncode}\n{result.stderr}")
+        self.assertFalse(
+            Path(absolute_target).exists(),
+            "Absolute out-of-root target must not be written",
+        )
+        self.assertIn("agentic-escape-test.txt", result.stderr, "stderr must mention the offending path")
+
+
 if __name__ == "__main__":
     unittest.main(verbosity=2)


tyson-solara6 (approved):
**Approving** - path-containment guard is correct; CI all green.

The write-path guard is implemented correctly: `target.resolve().is_relative_to(project_root.resolve())` handles `../`, absolute paths, and symlink-within-root escapes (since `resolve()` follows symlinks to the real path). The guard fires before any `mkdir`/`copy2` call, so no filesystem mutation happens for out-of-root entries. Exit code 3, stderr output, and `continue` are all consistent with the existing partial-apply contract.

Tests cover the two main attack vectors (`../escape.txt` and absolute out-of-root paths). Both assert exit 3, no file created, and path mentioned in stderr. Mutation-checked (removing the guard causes test failure).

---

**Non-blocking findings (no change needed to land this PR):**

**[Minor]** `Path.is_relative_to()` requires Python 3.9+. No minimum Python version is declared in the repo (no `pyproject.toml` with `python_requires`, no CI matrix entry for 3.8). Python 3.8 EOL'd in 2024 so this is likely a non-issue in practice, but worth one line in `pyproject.toml` (`python_requires = ">=3.9"`) to make the constraint explicit.

**[Minor - follow-up issue, not this PR's scope]** The `entry["seed"]` read source is not containment-checked. A manifest with `seed: "../../../etc/shadow"` makes `seed_path = manifest_dir / "../../../etc/shadow"`, and `shutil.copy2(seed_path, target)` reads from that path. The write side is now guarded; the read side is still unrestricted. Worth a follow-up PR applying the same `is_relative_to(manifest_dir.resolve())` guard to `seed_path`.

**[Minor - theoretical for a local dev tool]** TOCTOU: between `target.resolve()` at the check and `shutil.copy2()` at the write, an adversarial process with concurrent FS access could swap a path to a symlink. Practical risk is near-zero for a local dev tool run on the developer's own machine, but worth noting if `agentic-migrate` is ever run in a shared or container environment.

**[Nitpick]** Test suite covers `../` and absolute paths but no symlink-within-root case. The guard does handle it correctly via `resolve()`; a third test case would complete the picture.

tyson-solara6 (approved):

APPROVE - security fix is correct and well-tested. Findings below are informational.

**[OK] Guard logic is correct.** `resolved_root` is computed once outside the loop (preventing TOCTOU), and the `continue` on a blocked entry precedes both `mkdir` and `shutil.copy2` - nothing is written for any out-of-root path. Relative traversal (`../`), absolute injection, and symlink escape via `resolve()` are all handled.

**[OK] gitignore write is not a risk.** The gitignore section writes only to the hardcoded `project_root / ".gitignore"` - no attacker-controlled path. No other manifest-derived write sites exist outside the guarded `files` loop.

**[Minor] Python 3.9+ implicit requirement.** `Path.is_relative_to()` was added in 3.9. On Python 3.8 (EOL but still common in CI images), the call would raise `AttributeError`, which is caught by the outer `except Exception` and sets `errored=True` - silently failing to block the traversal. If the project's CI already enforces Python 3.9+, this is a non-issue; worth a comment in the PR body or a version guard.

**[Minor] Missing symlink-as-entry test.** No test for a manifest entry whose resolved path exits via an internal symlink (e.g. `path: "link-out"` where `link-out -> ../../../etc/passwd`). The `resolve()` call handles this correctly in production, but the test suite doesn't prove it. Suggest a follow-up issue.

**[Minor] Temp dir leak in test.** `target_dir = Path(_tf.mkdtemp())` in `test_absolute_path_blocked` is never cleaned up. Leaves temp dirs on the test runner. Cosmetic - consider `addCleanup(shutil.rmtree, target_dir)`.

CI: all 13 checks pass. Mergeable. Ready for merge after #331 lands.

PR #333, opened 2026-06-26
refactor(bin): extract shared atomic_write + acquire_exclusive_lock to _lib.py (#265) [bin-stack 3/3]

### LINKED ISSUE #265: Extract bin/_lib.py shared helpers (atomic_write, acquire_exclusive_lock)
## Problem

Two Python CLIs (`bin/agentic-identity`, `bin/agentic-migrate`) each implement the `fcntl.flock` acquire/release pattern inline. The two implementations have already drifted:

- `agentic-identity:185-208` uses `fd.close()` on failure.
- `agentic-migrate:168-172` uses `os.close(fd)` on failure.

Different error-handling semantics for the same operation. As more CLIs need cross-process locks (e.g. a future `agentic-cost` write lock), the drift will compound.

## Evidence

```
$ grep -n "fcntl.flock" bin/agentic-identity bin/agentic-migrate
bin/agentic-identity:40:               flushPendingBuffer: serialized via fcntl.flock on .flush.lock;
bin/agentic-identity:169:    Race-safe via fcntl.flock exclusive lock on .flush.lock for the whole loop.
bin/agentic-identity:196:            fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
bin/agentic-identity:323:        fcntl.flock(fd, fcntl.LOCK_UN)
bin/agentic-migrate:169:            fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
bin/agentic-migrate:218:        fcntl.flock(fd, fcntl.LOCK_UN)
```

Also relevant: atomic-write pattern duplication (issue #6 covers the JS-side; Python side has the correct pattern but it is inline in `agentic-identity:145-163` and would benefit from extraction).

## Goal

Create `bin/_lib.py` with two shared helpers:

1. `acquire_exclusive_lock(lock_path, timeout=30.0)` - context manager that acquires `fcntl.fLOCK_EX` with LOCK_NB + sleep loop, yields the fd, releases on exit.
2. `atomic_write(path, content, mode=0o600)` - writes content to `<path>.tmp` then renames; cleans up `.tmp` on failure.

Refactor `agentic-identity` and `agentic-migrate` to use them. Behavior is identical; only the implementation is shared.

## Repo conventions to match

- Stdlib only - no new deps in `_lib.py`.
- Module docstring on `_lib.py` (follow the manifest convention used in every other `bin/` file).
- Helpers are private (`_` prefix would conflict with the filename; use plain names and document that `_lib.py` is internal).
- `_lib.py` is imported by sibling CLIs via `from _lib import acquire_exclusive_lock, atomic_write`. This works because `bin/` is on `sys.path` when the CLIs run from the same directory. Verify with `python3 -c "import sys; sys.path.insert(0, 'bin'); from _lib import atomic_write"`.

## Steps

1. **Create `bin/_lib.py`** with module docstring and both helpers:

   ```python
   """
   Purpose: Shared helpers for bin/ Python CLIs. Internal module - not part of
            the public API. Imported by sibling CLIs via `from _lib import ...`.

   Public API:
     acquire_exclusive_lock(lock_path: Path, timeout: float = 30.0) -> contextmanager
     atomic_write(path: Path, content: str, mode: int = 0o600) -> None

   Upstream deps: Python 3 stdlib (fcntl, os, time, pathlib, contextlib).

   Downstream consumers: bin/agentic-identity, bin/agentic-migrate
     (extend this list as more CLIs adopt the helpers).

   Failure modes:
     acquire_exclusive_lock: on timeout, raises TimeoutError. Callers should
       catch and handle (print stderr, return). On success, yields a locked fd;
       lock is released and fd closed on context exit.
     atomic_write: on any write/rename failure, unlinks the .tmp file and re-raises.
       Callers should catch and handle (print stderr, return non-zero).

   Performance: O(1). File I/O only.
   """

   from __future__ import annotations

   import fcntl
   import os
   import time
   from contextlib import contextmanager
   from pathlib import Path
   from typing import Iterator


   @contextmanager
   def acquire_exclusive_lock(lock_path: Path, timeout: float = 30.0) -> Iterator[int]:
       """Acquire an exclusive flock on lock_path with a bounded busy-wait.

       Uses LOCK_NB + sleep loop (no SIGALRM, for portability). Yields the fd
       on success; releases and closes on context exit. Raises TimeoutError
       if the lock cannot be acquired within `timeout` seconds.
       """
       lock_path.parent.mkdir(parents=True, exist_ok=True)
       lock_path.touch(exist_ok=True)
       fd = open(lock_path, 'r')
       deadline = time.monotonic() + timeout
       acquired = False
       try:
           while time.monotonic() < deadline:
               try:
                   fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
                   acquired = True
                   break
               except BlockingIOError:
                   time.sleep(0.1)
           if not acquired:
               raise TimeoutError(
                   f"could not acquire lock at {lock_path} within {timeout}s"
               )
           yield fd.fileno()
       finally:
           if acquired:
               try:
                   fcntl.flock(fd, fcntl.LOCK_UN)
               except OSError:
                   pass
           fd.close()


   def atomic_write(path: Path, content: str, mode: int = 0o600) -> None:
       """Atomically write content to path via tmp + rename.

       Writes to <path>.tmp, chmods, then renames into place. On any failure,
       unlinks the .tmp file (best-effort) and re-raises so the caller can
       handle (print stderr, return non-zero).
       """
       path.parent.mkdir(parents=True, exist_ok=True)
       tmp = path.with_suffix(path.suffix + '.tmp')
       try:
           tmp.write_text(content, encoding='utf-8')
           os.chmod(tmp, mode)
           tmp.rename(path)
       except Exception:
           try:
               tmp.unlink(missing_ok=True)
           except OSError:
               pass
           raise
   ```

2. **Refactor `bin/agentic-identity:145-163`** (`_write_identity`) to use `atomic_write`:

   ```python
   from _lib import atomic_write

   def _write_identity(lines: list[str], target_path: Path | None = None) -> int:
       dest = target_path if target_path is not None else IDENTITY_PATH
       content = "\n".join(lines) + "\n"
       try:
           atomic_write(dest, content, mode=0o600)
       except Exception as exc:
           print(f"agentic-identity: write failed: {exc}", file=sys.stderr)
           return 1
       return 0
   ```

3. **Refactor `bin/agentic-identity:185-208`** (the flock acquire block in `flushPendingBuffer`) to use `acquire_exclusive_lock`:

   ```python
   from _lib import acquire_exclusive_lock

   # ... inside flushPendingBuffer ...
   try:
       with acquire_exclusive_lock(FLUSH_LOCK_PATH, timeout=30.0):
           # ... existing loop body, minus the lock boilerplate ...
   except TimeoutError:
       print(
           "agentic-identity confirm: lock timeout after 30s; retry.",
           file=sys.stderr,
       )
       return 0
   ```

   Preserve the exact stderr message and return value.

4. **Refactor `bin/agentic-migrate:168-172` and `:217-221`** similarly.

5. **Add `bin/tests/test_lib.py`** with focused tests for each helper:
   - `atomic_write`: verify content, mode, that .tmp is cleaned up on simulated failure, that rename is atomic.
   - `acquire_exclusive_lock`: verify acquire/release, verify TimeoutError on contention (spawn a child holding the lock).

## Files in scope

Create:
- `bin/_lib.py`
- `bin/tests/test_lib.py`

Modify:
- `bin/agentic-identity` (refactor to use helpers)
- `bin/agentic-migrate` (refactor to use helpers)
- `bin/tests/test_agentic_identity.py` (adjust if any tests assert on the internal lock/write implementation)
- `bin/tests/test_agentic_migrate.py` (same)

- Do NOT refactor `hooks/stop-context.js` - JS side is a separate concern (issue #6 covers the .tmp cleanup there).
- Do NOT add helpers for yaml parsing, glob, or anything else - only the two helpers listed. `_lib.py` should stay under ~100 lines.
- Do NOT change the public API or stderr output of any CLI.

```
# New lib tests pass
python3 bin/tests/test_lib.py

# Existing tests pass unchanged (behavior preserved)
python3 bin/tests/test_agentic_identity.py
python3 bin/tests/test_agentic_migrate.py

# No remaining inline flock boilerplate
# (each CLI should import from _lib, not call fcntl.flock directly)
grep -n "fcntl.flock" bin/agentic-identity bin/agentic-migrate
# Expected: only docstring references remain, no live fcntl.flock calls.

# _lib.py is under 100 lines
test "$(wc -l < bin/_lib.py)" -le 100
```

## Escape hatches

- If importing `from _lib import ...` fails because `bin/` is not on `sys.path` when a CLI is invoked from a different directory, use `sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))` at the top of each CLI before the import. Verify by running `python3 /full/path/to/bin/agentic-identity show` from a non-`bin/` directory.
- If existing tests assert on the internal lock implementation (e.g. mock `fcntl.flock` directly), update them to mock the new helper instead.

## Maintenance note

Reviewers: `_lib.py` is intentionally small. Resist adding "just one more helper" - every addition makes it harder to keep stdlib-only. If a future helper needs a third-party dep, it does not belong in `_lib.py`. The two helpers here are the ones with demonstrated drift.

bloodf:
> **Audit batch #258 — bin-stack 3/3.** Merge order: #331 → #332 → **#333** (last). Stacked on #331+#332 — its diff shows their files until they merge. Merge **after both #331 and #332**; rebase onto `main` if either squash-merges.

Extracts two duplicated primitives from `agentic-identity` and `agentic-migrate` into a new `bin/_lib.py`. Closes #265.

- `bin/_lib.py` (new, stdlib-only): `atomic_write(path, content, mode=0o600)` (write-`.tmp` → optional chmod → rename, unlink `.tmp` on failure; `mode=None` skips chmod) and `acquire_exclusive_lock(lock_path, timeout=30.0)` (LOCK_EX|LOCK_NB sleep-retry context manager, raises `RuntimeError` on timeout, fd closed on every path).
- `agentic-identity`: `_write_identity` → `atomic_write`; `flushPendingBuffer` flock loop → `acquire_exclusive_lock`. The global-log read + entire pending-file loop stay **inside** the lock (preserves #268's dedup correctness). Timeout RuntimeError is caught → same "retry" message, no flush, no crash.
- `agentic-migrate`: `_stamp_project_version` → `atomic_write(mode=None)` (config.json was never chmod'd — preserved).
- Reconciled the `fd.close()` vs `os.close(fd)` drift onto the file-object pattern.
- `agentic-migrate`'s `_acquire_lock` is **not** extracted — it's a different mechanism (one-shot, PID-body stale detection), documented in the header note.

Tests: `test__lib.py` (10, incl lock contention + release-on-exception); both consumer suites still green (#268 dedup + #269 guard tests pass). 35 tests total.

**Stacked on #268 + #269** — merge after both land on `main` (rebase if their SHAs change on squash-merge).

### DIFF
--- a/bin/_lib.py
+++ b/bin/_lib.py
@@ -1,0 +1,131 @@
+#!/usr/bin/env python3
+"""
+Internal shared helpers for agentic-engineering bin/ CLIs.
+NOT a public CLI - do not invoke directly.
+
+Purpose: Provide two cross-process primitives reused by multiple CLIs:
+  1. acquire_exclusive_lock - fcntl.LOCK_EX context manager with sleep-retry
+     until timeout; used for multi-process coordination (e.g. flush lock).
+  2. atomic_write - write content to <path>.tmp then rename; cleans up .tmp on
+     failure; optional chmod mode.
+
+Public API:
+  acquire_exclusive_lock(lock_path, timeout=30.0)
+    Context manager. Opens lock_path as a Python file object (buffered), acquires
+    fcntl.LOCK_EX | LOCK_NB via a 0.1s sleep-retry loop until timeout, yields the
+    file object, releases (LOCK_UN) and closes on exit. Raises RuntimeError on
+    timeout so callers can distinguish "another holder" from a filesystem error.
+    Caller is responsible for ensuring lock_path and its parent exist before entry.
+
+  atomic_write(path, content, mode=0o600)
+    Writes str content to path.with_suffix('<ext>.tmp') then renames into place.
+    When mode is not None, applies os.chmod to the tmp file before rename.
+    On any exception, unlinks the tmp file (missing_ok) and re-raises.
+    path must be a pathlib.Path.
+
+Upstream deps: Python 3 stdlib only (contextlib, fcntl, os, time, pathlib).
+
+Downstream consumers: bin/agentic-identity (both helpers),
+                      bin/agentic-migrate (atomic_write).
+
+Failure modes:
+  acquire_exclusive_lock: raises RuntimeError("lock timeout") after timeout seconds
+    with no lock held; the underlying fd is always closed before raising.
+    OS errors opening the lock file propagate to the caller unchanged (the file
+    must exist before calling; existence is the caller's responsibility).
+  atomic_write: on any write/chmod/rename failure, removes the .tmp file
+    (missing_ok semantics) and re-raises the original exception. The destination
+    file is never partially written. The .tmp suffix is appended to the full
+    filename (e.g. identity.yml -> identity.yml.tmp) to stay in the same
+    directory and on the same filesystem as the destination.
+
+Performance: Standard. acquire_exclusive_lock sleeps 0.1s per retry (~300 retries
+  over 30s); atomic_write is a single write + fsync-less rename (same filesystem).
+"""
+
+from __future__ import annotations
+
+import fcntl
+import os
+import time
+from contextlib import contextmanager
+from pathlib import Path
+from typing import Generator
+
+
+@contextmanager
+def acquire_exclusive_lock(
+    lock_path: Path,
+    timeout: float = 30.0,
+) -> Generator[object, None, None]:
+    """Context manager: acquire fcntl.LOCK_EX on lock_path.
+
+    Opens lock_path as a Python file object ('r' mode - the file must already
+    exist), retries with 0.1s sleep until timeout, yields the file object,
+    then releases LOCK_UN and closes on exit.
+
+    Raises RuntimeError on timeout (lock not acquired; fd is closed before
+    raising). OS errors on open propagate unchanged.
+
+    Usage:
+        with acquire_exclusive_lock(lock_path) as fd:
+            # critical section
+            ...
+    """
+    fd = open(lock_path, "r")  # noqa: SIM115 - intentional: file stays open for flock
+    try:
+        deadline = time.monotonic() + timeout
+        acquired = False
+        while time.monotonic() < deadline:
+            try:
+                fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+                acquired = True
+                break
+            except BlockingIOError:
+                time.sleep(0.1)
+
+        if not acquired:
+            fd.close()
+            raise RuntimeError(f"acquire_exclusive_lock: timeout after {timeout}s on {lock_path}")
+
+        try:
+            yield fd
+        finally:
+            fcntl.flock(fd, fcntl.LOCK_UN)
+            fd.close()
+
+    except RuntimeError:
+        raise
+    except BaseException:
+        # Covers exceptions from open() after fd is assigned but before acquired.
+        # If fd was opened but flock not yet attempted (shouldn't happen in normal
+        # flow but guard anyway), close it.
+        try:
+            fd.close()
+        except Exception:
+            pass
+        raise
+
+
+def atomic_write(path: Path, content: str, mode: int | None = 0o600) -> None:
+    """Write content to path atomically via a .tmp sibling.
+
+    Steps:
+      1. Write content to <path>.tmp (text, utf-8).
+      2. If mode is not None, chmod <path>.tmp to mode.
+      3. Rename <path>.tmp -> path.
+
+    On any failure, unlinks <path>.tmp (missing_ok) and re-raises. The
+    destination file is never partially overwritten.
+
+    path.parent must already exist (no mkdir here - callers handle that).
+    """
+    tmp = path.parent / (path.name + ".tmp")
+    try:
+        tmp.write_text(content, encoding="utf-8")
+        if mode is not None:
+            os.chmod(tmp, mode)
+        tmp.rename(path)
+    except Exception:
+        tmp.unlink(missing_ok=True)
+        raise
--- a/bin/agentic-identity
+++ b/bin/agentic-identity
@@ -81,13 +81,18 @@ import importlib.util as _ilu
 import importlib.machinery as _ilm
 
 _LIB_PATH = Path(__file__).parent / "_lib.py"
-_loader = _ilm.SourceFileLoader("_lib", str(_LIB_PATH))
-_spec = _ilu.spec_from_loader("_lib", _loader)
-_lib_mod = _ilu.module_from_spec(_spec)  # type: ignore[arg-type]
-_loader.exec_module(_lib_mod)
+if "_lib" in _sys.modules:
+    _lib_mod = _sys.modules["_lib"]
+else:
+    _loader = _ilm.SourceFileLoader("_lib", str(_LIB_PATH))
+    _spec = _ilu.spec_from_loader("_lib", _loader)
+    _lib_mod = _ilu.module_from_spec(_spec)  # type: ignore[arg-type]
+    _loader.exec_module(_lib_mod)
+    _sys.modules["_lib"] = _lib_mod
+    del _loader, _spec
 acquire_exclusive_lock = _lib_mod.acquire_exclusive_lock
 atomic_write = _lib_mod.atomic_write
-del _sys, _ilu, _ilm, _loader, _spec, _lib_mod
+del _sys, _ilu, _ilm, _lib_mod
 
 IDENTITY_PATH = Path(os.path.expanduser("~/.agentic/identity.yml"))
 PENDING_DIR = Path(os.path.expanduser("~/.agentic/session-log/.pending"))
@@ -206,15 +211,9 @@ def flushPendingBuffer(confirmed_dev_id: str, repo_root_filter: str | None = Non
     FLUSH_LOCK_PATH.parent.mkdir(parents=True, exist_ok=True)
     FLUSH_LOCK_PATH.touch(exist_ok=True)
 
-    try:
-        ctx = acquire_exclusive_lock(FLUSH_LOCK_PATH, timeout=30.0)
-    except OSError as exc:
-        print(f"agentic-identity confirm: cannot open lock file: {exc}", file=sys.stderr)
-        return 0
-
     flushed = 0
     try:
-        with ctx:
+        with acquire_exclusive_lock(FLUSH_LOCK_PATH, timeout=30.0):
             pending_pattern = str(PENDING_DIR / "*.json")
             pending_files = sorted(glob.glob(pending_pattern))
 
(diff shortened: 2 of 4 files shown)


tyson-solara6 (changes requested):
**Requesting changes** - one blocking correctness regression; two minors.

---

**[Major - Blocker] `except OSError` in `flushPendingBuffer` is dead code**

`acquire_exclusive_lock` is a `@contextmanager`-decorated generator function. Calling `ctx = acquire_exclusive_lock(...)` only creates the generator object - the function body does not execute. Execution begins when `with ctx:` is entered (at that point `__enter__` calls `next()` on the generator, which runs until `yield`).

This means `open(lock_path, "r")` inside `acquire_exclusive_lock` executes during `with ctx:`, not during the assignment. The `except OSError` wrapping the assignment catches nothing:

```python
try:
    ctx = acquire_exclusive_lock(FLUSH_LOCK_PATH, timeout=30.0)  # no open() here yet
except OSError as exc:            # <- dead: open() hasn't run
    print(...)
    return 0

try:
    with ctx:                     # <- open() runs HERE
        ...                       # OSError from open() propagates through this
except RuntimeError:              # <- doesn't match OSError
    ...
    return 0
# OSError escapes flushPendingBuffer unhandled
```

The old code called `open()` directly in a guarded `try/except OSError` block and returned 0 gracefully. The refactor breaks that - an `OSError` from `open()` (e.g., lock file deleted between `touch()` and `with ctx:`, or a permissions failure) now propagates as an unhandled exception to the caller.

**Fix:** Combine the OSError and RuntimeError catches around `with ctx:`:

```python
flushed = 0
try:
    with acquire_exclusive_lock(FLUSH_LOCK_PATH, timeout=30.0):
        # ... critical section unchanged ...
except OSError as exc:
    print(f"agentic-identity confirm: cannot open lock file: {exc}", file=sys.stderr)
    return 0
except RuntimeError:
    print("agentic-identity confirm: lock timeout after 30s; retry.", file=sys.stderr)
    return 0
```

This eliminates the separate `ctx = ...` assignment, removes the dead `except OSError`, and correctly catches both failure modes.

---

**[Minor] `atomic_write` tmp suffix changed silently**

Old `agentic-migrate` used `config_path.with_suffix(".tmp")` which for `config.json` produces `config.tmp`. The new `atomic_write` uses `path.name + ".tmp"` which produces `config.json.tmp`. Different filename, not mentioned in the PR description. Harmless in practice (the tmp is transient) but worth noting in case any `.gitignore` pattern or cleanup script matches on the old form.

**[Minor] `importlib` loader does not cache in `sys.modules`**

Both `agentic-identity` and `agentic-migrate` use `SourceFileLoader` to import `_lib.py` by path. If both scripts are imported into the same Python process (plausible in a test harness that does `import` rather than subprocess invocation), `_lib.py` is loaded twice under different module identities. The standard guard is to check `sys.modules` before calling `exec_module`. Low severity for standalone CLI invocations; relevant if tests grow to import these scripts directly.

---

The `_lib.py` implementation itself (`atomic_write`, `acquire_exclusive_lock`) is clean. The DRY extraction is the right direction. Once the dead exception handler is fixed, this is ready.

tyson-solara6 (approved):

APPROVE - clean DRY extraction, correct on all critical paths, good test coverage.

**[OK] `acquire_exclusive_lock` fd-close on all paths.** Timeout branch closes fd explicitly before raising `RuntimeError`; normal exit and body exceptions both flow through the inner `finally` (LOCK_UN + close). The outer `except BaseException` guard handles the narrow window between `open()` and first `flock` attempt. No double-close or fd-leak paths found.

**[OK] `atomic_write` tmp-suffix is correct.** Uses `path.parent / (path.name + ".tmp")` - produces `identity.yml.tmp` not `identity.tmp`. The `.tmp` cleanup in `except Exception` fires on write, chmod, or rename failure. Correct.

**[OK] Lock scope in agentic-identity preserved.** The `with ctx:` block encloses the complete critical section: global-log read, `seen_uuids` set build, and the full per-pending-file loop. The dedup improvement (JSON-parse per line) stays inside the lock - #268 dedup correctness is preserved.

**[OK] `agentic-migrate`'s `_acquire_lock` correctly not extracted.** One-shot PID-body stale detection is a different mechanism; extracting it would add complexity with no benefit. The comment at the exclusion site documents this clearly.

**[Minor] Missing exception-during-yield test for `acquire_exclusive_lock`.** If the lock body raises, the `finally` correctly releases (Python contextmanager guarantee), but there is no explicit test for it. The lock contention test demonstrates LOCK_NB behavior; this is the one gap. Low risk for a stdlib-only primitive.

**[Minor] Module manifest docstring says `path.with_suffix('<ext>.tmp')`** - the actual implementation uses `path.parent / (path.name + ".tmp")`. The described behavior is correct but the API description is imprecise. Not a stale-manifest issue, but worth aligning.

**Stacking note.** Out-of-order merge (e.g. #333 before #332) would leave `_lib.py` on main but the callers un-migrated - no test failure, just an incomplete refactor until all three merge. Low risk given clear PR body.

CI: all 13 checks pass. Merge after #331 + #332.

tyson-solara6:
**Automated review pass - 2026-06-29**

No new commits on this branch since our APPROVE on 2026-06-28 (HEAD: `83a1c646`), so no code re-review. However, the branch is now **CONFLICTING** against `main` (mergeStateStatus: DIRTY) - likely due to other PRs merging since our approval.

Please rebase onto `main` and resolve conflicts. Once clean and CI confirms green, this is approvable without further content review - the DRY extraction was already verified.

bloodf:
Rebased onto main, resolved conflicts, and fixed the dead `except OSError` in `flushPendingBuffer` by wrapping `with acquire_exclusive_lock(...)` and catching both `OSError` and `RuntimeError`. Also added `sys.modules` guards for `_lib.py` loading.

All 35 tests pass. Ready for re-review.

tyson-solara6 (approved):
**Re-approving** after rebase - content correct, exception handling fix is sound.

The core extraction (`bin/_lib.py` with `atomic_write` + `acquire_exclusive_lock`) was approved on the prior HEAD. The rebase added one commit: 'fix exception handling in flushPendingBuffer (#333)'.

The fix is correct and conservative: in the `acquire_exclusive_lock` context manager, the added `try: fd.close() / except Exception: pass` in the outer exception handler covers the edge case where `fd` was opened but `flock` had not yet been attempted before an exception was raised. In normal flow this path is unreachable, but the guard is a defensive correctness improvement. No behavior change on the happy path.

Also solid: the `flushPendingBuffer` dedup logic upgrade (O(1) UUID set lookups vs. full-text `in` scan) and `fd.close()` vs `os.close(fd)` drift reconciled onto the file-object pattern.

All 13 CI checks pass.

Issue #266, opened 2026-06-19
Add telemetry-health counter (.agentic/.telemetry-health.json) for silent-fail visibility


bloodf:
## Problem

`hooks/stop-context.js` has ~21 silent-catch blocks on state writes (loop-state, batch-state, events.jsonl, session-log, context.md, capture-gap cursor). The silent-fail discipline is correct - never block session exit. But the discipline produces zero signal when writes fail.

Disk full, permissions drift, corrupted JSON: all fail silently. `bin/agentic-cost` reads these files but has no way to report "N sessions were dropped". `agentic-status` has no field to surface "your session log has not been written in 5 sessions". Users discover data loss only when they run a rollup and the numbers look wrong.

Related to audit index issue: #258. Pairs naturally with issue #5 (both touch stop-context.js read paths).

## Evidence

`hooks/stop-context.js` silent-catch sites on state writes (not the JSON-line-skip or optional-git sites which are correctly silent):

```
:151  writeLoopState          (loop-state.json write)
:197  writeBatchState         (batch-state.json write)
:299  writeSessionTotal       (events.jsonl session_total append)
:326  removeLearningsAgentSession (marker unlink)
:401  appendIdentityNudgeToContextMd (context.md append)
:441  computeSessionTotals    (events.jsonl read)
:493  writeSessionLog         (per-project session-log append)
:531  writeSessionLogGlobal   (global session-log append)
:641  writePendingBuffer      (pending-buffer write)
:913  appendCaptureGapNoticeToContextMd (cursor write)
:1052 run (git status)        (optional - git missing is OK)
:1133 run (wrap coexistence)  (context.md write)
:1183 run (session-log path)  (session-log append)
:1204 run (context.md write)  (primary context.md write)
:1263 run (final state writes) (session-log + capture-gap)
```

## Goal

Add a single append-only failure counter: `.agentic/.telemetry-health.json`. Every silent-catch on a state write increments a per-target counter and records the last error message + timestamp. `bin/agentic-status` reads the file and surfaces a summary.

The counter file itself is also written best-effort behind a silent catch - if it fails, the hook still exits cleanly. No new blocking I/O.

## Repo conventions to match

- Silent-fail discipline: the health counter write MUST NOT throw or block.
- Atomic write pattern (tmp + rename, with cleanup - see issue #6).
- No new deps.
- File lives under `.agentic/` (gitignored via the umbrella) - not committed.

## Steps

[...]

PR #334, opened 2026-06-26
fix(hooks): clean up orphan .tmp files in stop-context atomic-write catch blocks (#262) [hooks-stack 1/3]

### LINKED ISSUE #262: Add .tmp cleanup to atomic-write catch blocks in stop-context.js
## Problem

The atomic-write pattern in `hooks/stop-context.js` writes to a `.tmp` file then renames it into place. On rename failure (or write failure), the `.tmp` file is not cleaned up. Orphaned `.tmp` files accumulate in `.agentic/` across sessions.

Compare to the correct pattern in `bin/agentic-identity:154-162` which DOES unlink the tmp file in its except block.

## Evidence

`hooks/stop-context.js:146-148` (writeLoopState):

```js
const tmpPath = loopStatePath + '.tmp';
fs.writeFileSync(tmpPath, JSON.stringify(loopState, null, 2));
fs.renameSync(tmpPath, loopStatePath);
// ...
} catch (_) {
  // Silent failure - the 10-minute implicit-interrupt heuristic handles missed writes
}
```

No `fs.unlinkSync(tmpPath)` in the catch.

Same pattern at:
- `stop-context.js:194-196` (writeBatchState)
- `stop-context.js:636,640` (writePendingBuffer)
- `stop-context.js:912` (capture-gap cursor update)

Contrast with `bin/agentic-identity:154-162`:

```python
tmp = dest.with_suffix(".yml.tmp")
try:
    tmp.write_text(content, encoding="utf-8")
    os.chmod(tmp, 0o600)
    tmp.rename(dest)
except Exception as exc:
    print(f"agentic-identity: write failed: {exc}", file=sys.stderr)
    tmp.unlink(missing_ok=True)   # <-- cleanup
    return 1
```

## Goal

Add `fs.unlinkSync(tmpPath)` cleanup to each atomic-write catch block in `hooks/stop-context.js`. Suppress unlink errors (the file may not exist if writeFileSync itself failed).

## Repo conventions to match

- Silent-fail discipline: the cleanup itself must not throw. Wrap in try/catch.
- No logging in this issue (that is issue #10).
- Match the Python side's `missing_ok=True` semantics: unlink should not fail if the file is already gone.

## Steps

For each of the 4 atomic-write sites, modify the catch block:

1. **`writeLoopState` (around line 146-153)**:

   Before:
   ```js
   } catch (_) {
     // Silent failure - the 10-minute implicit-interrupt heuristic handles missed writes
   }
   ```

   After:
   ```js
   } catch (_) {
     // Silent failure - the 10-minute implicit-interrupt heuristic handles missed writes.
     // Clean up orphaned .tmp file if it exists.
     try { fs.unlinkSync(tmpPath); } catch (_) { /* tmp may not exist if writeFileSync failed */ }
   }
   ```

   Note: `tmpPath` is declared with `const` inside the `try` block, so it is not in scope in `catch`. Move the `const tmpPath = ...` declaration to **before** the `try` block so the catch can reference it:

   ```js
   const tmpPath = loopStatePath + '.tmp';
   try {
     fs.writeFileSync(tmpPath, JSON.stringify(loopState, null, 2));
     fs.renameSync(tmpPath, loopStatePath);
   } catch (_) {
     try { fs.unlinkSync(tmpPath); } catch (_) { /* not present */ }
   }
   ```

2. Apply the same refactor to:
   - `writeBatchState` (`stop-context.js:190-200`)
   - `writePendingBuffer` (`stop-context.js:634-643`)
   - `appendCaptureGapNoticeToContextMd` cursor update (`stop-context.js:910-913`)

3. **Extend `hooks/tests/test-stop-context-session-log.js`** with a test that:
   - Stages a scenario where `fs.renameSync` throws (mock it).
   - Verifies the `.tmp` file does not exist after the call returns.
   - Verifies no exception propagates (silent-fail preserved).

## Files in scope

Modify:
- `hooks/stop-context.js`
- `hooks/tests/test-stop-context-session-log.js`

- Do NOT change the Python side - it already has the correct pattern.
- Do NOT add logging (issue #10).
- Do NOT refactor the functions into a shared helper (issue #9 covers Python-side consolidation; JS-side is separate).

```
# Tests pass
node hooks/tests/test-stop-context-session-log.js

# Grep: every .tmp write site has a matching unlink in its catch
# Each of these lines should have an "unlinkSync" in the same function's catch:
grep -n "\.tmp'" hooks/stop-context.js
grep -n "unlinkSync" hooks/stop-context.js
# Manually verify the 4 .tmp sites each have a corresponding unlink.

# Stop hook still works
echo '{"cwd":"'"$PWD"'","session_id":"test","transcript":[]}' \
  | node hooks/stop-context.js
ls .agentic/*.tmp 2>/dev/null && echo "FAIL: orphaned tmp files" || echo "PASS: no orphans"
```

## Escape hatches

- If moving `const tmpPath` out of the `try` block triggers a lint error (declared but not used in some path), declare it with `let` initialized to `null` and assign inside the try. The catch then checks `if (tmpPath) try { fs.unlinkSync(tmpPath); } catch (_) {}`.
- If the existing tests do not have a mocking pattern for `fs.renameSync`, use `jest.mock` or `sinon` if already in the devDeps - otherwise inject the mock via Node's `--require` hook or restructure the test to inject a fake `fs` object. If neither is straightforward, skip the specific mock-based test and add a comment in the test file explaining the gap.

## Maintenance note

Reviewers: the scope is small and the pattern is repetitive. Verify all 4 sites are touched, not just the first. The Python side already does this correctly - the JS side is just catching up.

bloodf:
> **Audit batch #258 — hooks-stack 1/3.** Merge order: **#334 → #335 → #336** (all touch `hooks/stop-context.js`). This is the base of the stack — merge first. #335 and #336 rebase onto `main` if this squash-merges (SHA changes).

Atomic-write sites in `stop-context.js` wrote to a `.tmp` then renamed; on failure the `.tmp` was orphaned in `.agentic/`. Closes #262.

- Added a best-effort `try { fs.unlinkSync(tmpPath); } catch (_) {}` to the catch of all 4 atomic-write sites: `writeLoopState`, `writeBatchState`, `writePendingBuffer`, capture-gap cursor.
- Hoisted the path vars out of their try blocks where needed (a `const` declared inside `try` is not in scope in `catch`); `writePendingBuffer` guards its `let pendingTmpFile` against the write-failed-before-assignment case.
- No logging (that is #266), no change to write/rename logic, silent-fail discipline preserved.

Regression test (sub-test 5): pre-plants a stale `.tmp`, forces the catch with corrupt JSON, asserts the `.tmp` is removed. Mutation-checked — removing the unlink fails the test. `test-stop-context-session-log.js` 19/19, `test-stop-context-deferred-wrap.js` 51/51.

### DIFF
--- a/hooks/stop-context.js
+++ b/hooks/stop-context.js
@@ -220,8 +220,8 @@ function writeLoopState(cwd) {
     return;
   }
 
+  const loopStatePath = path.join(cwd, '.agentic', 'loop-state.json');
   try {
-    const loopStatePath = path.join(cwd, '.agentic', 'loop-state.json');
     if (fs.existsSync(loopStatePath)) {
       const loopState = JSON.parse(fs.readFileSync(loopStatePath, 'utf-8'));
       if (loopState.status === 'active') {
@@ -235,6 +235,7 @@ function writeLoopState(cwd) {
     }
   } catch (_) {
     // Silent failure - the 10-minute implicit-interrupt heuristic handles missed writes
+    try { fs.unlinkSync(loopStatePath + '.tmp'); } catch (_e) { /* tmp absent or never created */ }
   }
 }
 
@@ -254,8 +255,8 @@ function writeBatchState(cwd, sessionId) {
     return;
   }
 
+  const batchStatePath = path.join(cwd, '.agentic', 'batch-state.json');
   try {
-    const batchStatePath = path.join(cwd, '.agentic', 'batch-state.json');
     if (!fs.existsSync(batchStatePath)) return;
     const batchState = JSON.parse(fs.readFileSync(batchStatePath, 'utf-8'));
 
@@ -281,6 +282,7 @@ function writeBatchState(cwd, sessionId) {
     fs.renameSync(tmpPath, batchStatePath);
   } catch (_) {
     // Silent failure - best-effort write; resume-time logic tolerates absent mark.
+    try { fs.unlinkSync(batchStatePath + '.tmp'); } catch (_e) { /* tmp absent or never created */ }
   }
 }
 
@@ -652,6 +654,7 @@ function generateUuid() {
  * @param {string|null} sessionId - Current session uuid (uuid v4 generated if null).
  */
 function writePendingBuffer(cwd, sessionId) {
+  let pendingTmpFile = null;
   try {
     const pendingDir = path.join(os.homedir(), '.agentic', 'session-log', '.pending');
     fs.mkdirSync(pendingDir, { recursive: true });
@@ -720,11 +723,14 @@ function writePendingBuffer(cwd, sessionId) {
 
     // Atomic write: tmp + rename
     const outFile = path.join(pendingDir, `${sessionUuid}.json`);
-    const tmpFile = outFile + '.tmp';
-    fs.writeFileSync(tmpFile, JSON.stringify(record, null, 2), 'utf8');
-    fs.renameSync(tmpFile, outFile);
+    pendingTmpFile = outFile + '.tmp';
+    fs.writeFileSync(pendingTmpFile, JSON.stringify(record, null, 2), 'utf8');
+    fs.renameSync(pendingTmpFile, outFile);
   } catch (_) {
     // Silent failure - consistent with all other write paths
+    if (pendingTmpFile) {
+      try { fs.unlinkSync(pendingTmpFile); } catch (_e) { /* tmp absent or never created */ }
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


tyson-solara6 (approved):
**Approving** - all 4 atomic-write sites correctly patched; CI all green.

The pattern is consistent and correct across all sites:

**Sites 1 and 2 (`writeLoopState`, `writeBatchState`):** path vars hoisted out of the try block (making them accessible in `catch`), then best-effort `unlinkSync(path + '.tmp')` in catch. The inner `try/catch` correctly handles ENOENT (silent swallow when no `.tmp` was ever written - e.g., the catch fires on a JSON parse error before any write). This false-positive-unlink-on-parse-error behavior is intentional and safe.

**Site 3 (`writePendingBuffer`):** `let pendingTmpFile = null` before the try is the right call here since `outFile` is computed deep inside the try. The null guard (`if (pendingTmpFile)`) correctly prevents an unlink attempt when the error occurred before the tmp path was assigned (e.g., `mkdirSync` failure). Once `pendingTmpFile` is set (immediately before `writeFileSync`), any subsequent failure correctly unlinks it.

**Site 4 (capture-gap cursor):** unlink added to existing single-line catch, consistent with other sites.

Test sub-test 5 covers the key invariants - success path leaves no `.tmp`, failure path (pre-planted stale `.tmp` + corrupt input) cleans it up. The mutation check (guard removed → test fails) is explicit, which is the right verification approach.

---

**[Nitpick]** Sub-test 5 exercises only Site 1 (`writeLoopState`). Sites 2-4 are not directly tested but the pattern is structurally identical. Not blocking.

PR #335, opened 2026-06-26
perf(hooks): read events.jsonl once per session-end in stop-context run() (#267) [hooks-stack 2/3]

bloodf:
> **Audit batch #258 — hooks-stack 2/3.** Merge order: #334 → **#335** → #336. Stacked on #334 — its diff shows #334 files until it merges. Merge **after #334**; rebase onto `main` if #334 squash-merges.

`stop-context.js` re-read `.agentic/events.jsonl` 3-4 times per session exit. Closes #267.

- One `cachedEventsRaw` read at the top of `run()`, threaded through `scanSessionAggregate`, `writeSessionTotal`, `computeSessionTotals`, `writeSessionLog`, `writePendingBuffer`, and `detectCaptureGap` (in `hooks/lib/capture-gap.js`). Exit paths now read the file once.
- Each consumer keeps a back-compat path: param omitted → it reads the file itself (identical to today). `null` → same as absent-file. The PostToolUse caller of `detectCaptureGap` (2-arg) is unchanged via this path.
- No aggregation logic changed; no logging added (that is #266). The later `session_total` append is never re-consumed — `scanSessionAggregate` filters it out by event type, so the cache being taken before the append is safe.
- JSDoc + module manifests updated for the new param.

Tests: session-log 19/19 (incl #262 tmp cleanup), deferred-wrap 51/51, capture-gap 29/29.

**Stacked on #262** — merge after #262 lands on `main` (rebase if its SHA changes on squash-merge).

Issue #267, opened 2026-06-19
Cache events.jsonl read in stop-context.js run() (re-read 3-4x today)


bloodf:
## Problem

`hooks/stop-context.js` re-reads `.agentic/events.jsonl` 3-4 times per session end:

1. `writeSessionTotal` -> `scanSessionAggregate` (line 215 `readFileSync`)
2. `computeSessionTotals` (line 416, called by `writeSessionLog` and `writePendingBuffer`) -> calls `scanSessionAggregate` internally
3. `detectCaptureGap` (line 717 `readFileSync`)

Each read is a synchronous `fs.readFileSync` of the full file. On a session with 200 events the file is ~200KB - not catastrophic, but read 3-4 times per session end, on every session, with no caching. At 5-10MB files (long-running projects) this becomes user-visible latency.

Related to audit index issue: #258

## Evidence

- `scanSessionAggregate` at `hooks/stop-context.js:215`: `const raw = fs.readFileSync(eventsPath, 'utf8');`
- `computeSessionTotals` at `hooks/stop-context.js:416-444`: calls `scanSessionAggregate(eventsPath, sessionId)`.
- `writeSessionTotal` at `stop-context.js:278-302`: calls `scanSessionAggregate(eventsPath, sessionId)`.
- `writeSessionLog` at `stop-context.js:454-496`: calls `computeSessionTotals(cwd, sessionId)`.
- `writePendingBuffer` at `stop-context.js:569-644`: calls `computeSessionTotals(cwd, sessionId)`.
- `detectCaptureGap` at `stop-context.js:717`: separate `fs.readFileSync(eventsPath, 'utf8')`.

The `run()` function at `stop-context.js:919-1281` invokes multiple of these per exit path. The wrap-coexistence path (line 1133) and the normal exit path (line 1204) both trigger 3+ reads.

## Goal

Read `events.jsonl` **once** at the top of `run()`. Pass the parsed lines (or the raw string) to every downstream function that needs it. Remove the per-function `readFileSync` calls.

Preserve all existing behavior:
- If the file is missing or unreadable, every consumer still gets a null/empty result.
- Each consumer still filters and aggregates independently - only the read is shared.

## Repo conventions to match

- The file is CommonJS (`require`). No ESM conversion.
- JSDoc annotations on every function - update them to reflect the new parameter.
- Silent-fail discipline: if the single read fails, every consumer still silently returns null/empty - do NOT add logging (that is issue #10).
[...]

PR #336, opened 2026-06-26
feat(hooks): add .telemetry-health.json failure counter + agentic-status surface (#266) [hooks-stack 3/3]

### LINKED ISSUE #266: Add telemetry-health counter (.agentic/.telemetry-health.json) for silent-fail visibility

bloodf:
> **Audit batch #258 — hooks-stack 3/3.** Merge order: #334 → #335 → **#336** (last). Stacked on #334+#335 — its diff shows their files until they merge. Merge **after both #334 and #335**; rebase onto `main` if either squash-merges.

Adds a `.agentic/.telemetry-health.json` failure counter for the silent-catch state-write sites in `stop-context.js`, plus a one-line summary in `agentic-status`. Closes #266.

- `recordTelemetryHealth(cwd, target, err)` — best-effort: whole body in a swallow-everything try/catch, atomic tmp+rename with its own .tmp cleanup, can never throw or block session exit (proven by sub-test 7: read-only `.agentic/` → hook still exits 0). Wired into 6 writer catch/success sites mapped to 5 target keys (`loop-state.json`, `batch-state.json`, `events.jsonl`, `session-log`, `context.md`).
- Records `last_success` on success and `failures`/`last_error`/`last_error_ts` on failure, matching the issue's format.
- `agentic-status` gains `_telemetry_health_line` — fail-open: absent or all-zero or unreadable → `telemetry-health: OK`; failures → one-line summary with the worst target.
- File is gitignored (under the `.agentic/` umbrella); no carve-out, not committed.

Per-exit cost is ~5 read-modify-writes of a sub-1KB file — three orders of magnitude smaller than the multi-MB `events.jsonl` reads #267 removed; does not regress that.

Tests: session-log 26/26 (sub-tests 6 forced-failure-increments-counter, 7 read-only-no-crash), deferred-wrap 51/51, capture-gap 29/29, `test_agentic_status.py` 61/61 (8 new health-line cases).

**Stacked on #262 + #267** — merge after both land on `main` (rebase if their SHAs change on squash-merge).

### DIFF
--- a/bin/agentic-status
+++ b/bin/agentic-status
@@ -436,13 +436,13 @@ def _telemetry_health_line(health_path: Path | None = None) -> str:
             data = json.load(fh)
         if not isinstance(data, dict):
             return "telemetry-health: OK (unreadable)"
-        writes = data.get("writes", {})
-        if not isinstance(writes, dict):
+        targets = data.get("targets", {})
+        if not isinstance(targets, dict):
             return "telemetry-health: OK (unreadable)"
 
         total_failures = 0
         failing_targets: list[tuple[str, int, str, str]] = []  # (target, count, msg, ts)
-        for target, info in writes.items():
+        for target, info in targets.items():
             if not isinstance(info, dict):
                 continue
             count = int(info.get("failures", 0) or 0)
--- a/bin/agentic-status
+++ b/bin/agentic-status
@@ -10,16 +10,21 @@ Purpose: Read-only inspection of the agentic-engineering activation
 
 Public API: agentic-status
             Prints a fixed-format multi-line report (status lines with
-            provenance, a profile-behavior explainer, and a how-to-adjust
-            block). No subcommands, no flags. Always exits 0.
+            provenance, a profile-behavior explainer, a telemetry-health
+            summary line, and a how-to-adjust block). No subcommands, no
+            flags. Always exits 0.
             _profile_behavior_block(profile) and _how_to_adjust_block(resolved)
             are the single rendering surface for the explainer text and are
             reused by callers (e.g. init-project shells out to this binary).
+            _telemetry_health_line(health_path) reads .agentic/.telemetry-health.json
+            and returns a one-line OK/failure summary; fail-open (returns OK
+            on any read or parse error).
 
 Upstream deps: Python 3 stdlib only (json, os, sys, pathlib, re).
                Reads ~/.claude/agentic-engineering.json, the project
                root AGENTS.md (resolved through CLAUDE.md @AGENTS.md
-               import if present), and .agentic/.activated.
+               import if present), .agentic/.activated, and
+               .agentic/.telemetry-health.json (fail-open; absent = OK).
 
 Downstream consumers: content/commands/agentic-status.md (command
                       spec); developers running the command
@@ -46,6 +51,7 @@ from pathlib import Path
 
 CONFIG_PATH = Path(os.path.expanduser("~/.claude/agentic-engineering.json"))
 SENTINEL_PATH = Path(".agentic/.activated")
+TELEMETRY_HEALTH_PATH = Path(".agentic/.telemetry-health.json")
 ROLE_MODELS_GLOBAL_PATH = Path(os.path.expanduser("~/.agentic/role-models.yml"))
 ROLE_MODELS_PROJECT_PATH = Path(".agentic/role-models.yml")
 HEALTH_FILE_PATH = Path(".agentic/.telemetry-health.json")
@@ -406,6 +412,62 @@ def _telemetry_health_block() -> list[str]:
     return lines
 
 
+# ANSI escape sequence stripper for exception text rendered by _telemetry_health_line.
+_ANSI = re.compile(r"\x1b\[[0-9;]*[mGKHF]")
+
+
+def _telemetry_health_line(health_path: Path | None = None) -> str:
+    """Return a one-line telemetry-health summary for the current project.
+
+    Reads .agentic/.telemetry-health.json (or the provided path override).
+    Returns 'telemetry-health: OK' when the file is absent or all failure
+    counters are zero. When failures exist, returns a summary of the form:
+      'telemetry-health: N write failures across M targets; last error <target>: <msg> at <ts>'
+    Fail-open: any read or parse error returns 'telemetry-health: OK (unreadable)'.
+
+    Args:
+        health_path: Path to .telemetry-health.json; defaults to TELEMETRY_HEALTH_PATH.
+    """
+    p = health_path if health_path is not None else TELEMETRY_HEALTH_PATH
+    if not p.is_file():
+        return "telemetry-health: OK"
+    try:
+        with p.open("r", encoding="utf-8") as fh:
+            data = json.load(fh)
+        if not isinstance(data, dict):
+            return "telemetry-health: OK (unreadable)"
+        writes = data.get("writes", {})
+        if not isinstance(writes, dict):
+            return "telemetry-health: OK (unreadable)"
+
+        total_failures = 0
+        failing_targets: list[tuple[str, int, str, str]] = []  # (target, count, msg, ts)
+        for target, info in writes.items():
+            if not isinstance(info, dict):
+                continue
+            count = int(info.get("failures", 0) or 0)
+            if count > 0:
+                total_failures += count
+                last_msg = str(info.get("last_error", ""))
+                safe_msg = _ANSI.sub("", last_msg)[:200]
+                ts = str(info.get("last_error_ts", ""))
+                failing_targets.append((target, count, safe_msg, ts))
+
+        if total_failures == 0:
+            return "telemetry-health: OK"
+
+        # Report most-recently-failed target for the one-line summary.
+        failing_targets.sort(key=lambda x: x[3], reverse=True)
+        last_target, _, last_msg, last_ts = failing_targets[0]
+        n_targets = len(failing_targets)
+        return (
+            f"telemetry-health: {total_failures} write failure(s) across "
+            f"{n_targets} target(s); last error {last_target}: {last_msg} at {last_ts}"
+        )
+    except Exception:
+        return "telemetry-health: OK (unreadable)"
+
+
 def _how_to_adjust_block(resolved: dict) -> list[str]:
     """Return the 'How to adjust' explainer lines.
 
@@ -456,6 +518,7 @@ def main(_argv: list[str]) -> int:
     out.append(f"  marker: {res['marker']}")
     out.append(f"  active: {active_display} ({res['active_reason']})")
     out.append(f"  sentinel: {SENTINEL_PATH} ({sentinel_status})")
+    out.append(f"  {_telemetry_health_line()}")
     out.extend(_role_models_status_block())
     out.extend(_telemetry_health_block())
     out.append("")
--- a/bin/tests/test_agentic_status.py
+++ b/bin/tests/test_agentic_status.py
@@ -38,6 +38,7 @@ _profile_behavior_block = _mod._profile_behavior_block
 _role_models_status_block = _mod._role_models_status_block
 _telemetry_health_block = _mod._telemetry_health_block
 _how_to_adjust_block = _mod._how_to_adjust_block
+_telemetry_health_line = _mod._telemetry_health_line
 main = _mod.main
 
 _EXPECTED_VARS = ("PI_HARNESS", "OMP_HARNESS", "OH_MY_PI_HARNESS", "AGENTIC_HARNESS")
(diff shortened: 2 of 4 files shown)


tyson-solara6 (changes requested):
Good structure and excellent test coverage throughout. The fail-open design, atomic tmp+rename discipline, and single-read `cachedEventsRaw` threading are all solid. Two issues need fixing before merge.

---

**[Major] `recordTelemetryHealth` is missing the traversal guard every other write function in this file uses**

Every write function in `stop-context.js` that constructs a path from `cwd` — `writeLoopState`, `writeBatchState`, `writeContextMdOrSpill`, etc. — applies a `path.resolve(cwd) !== cwd` guard before touching the filesystem. `recordTelemetryHealth` does not. It derives its write path directly from the caller-supplied `cwd`:

```js
const healthPath = require('path').join(cwd, '.agentic', '.telemetry-health.json');
```

Current callers all receive a validated `cwd` from `run()`, so this is not actively exploitable today. But `recordTelemetryHealth` is a module-level function that can be called from new sites, and the established file convention is that each write function validates its own `cwd`. Leaving this function without the guard breaks that invariant and creates a latent path-traversal risk.

Fix: add the same guard used elsewhere before the `path.join`:
```js
const resolvedCwd = require('path').resolve(cwd);
if (resolvedCwd !== cwd) return; // traversal component - skip silently
```
(Or use the module-level `path` reference — see Minor item below.)

---

**[Minor] `_telemetry_health_line` renders `last_msg` verbatim to the terminal**

`last_msg` comes from `str(info.get("last_error", ""))` — the raw exception message stored by `recordTelemetryHealth`. On most platforms that message contains a filesystem path, which is fine. However, a crafted `.telemetry-health.json` (writable by anyone with project directory access) could embed ANSI escape sequences that manipulate terminal output when `agentic-status` prints the health line.

Fix: strip ANSI escapes and bound the length before interpolating:
```python
import re
_ANSI = re.compile(r'\x1b\[[0-9;]*[mGKHF]')
safe_msg = _ANSI.sub('', last_msg)[:200]
```

---

**[Minor] Inline `require()` calls inside `recordTelemetryHealth`**

The function calls `require('path').join(...)` and `require('fs').readFileSync(...)` inline. Node module caching means this is functionally identical to using the module-level `path` and `fs` bindings, but it's inconsistent with how every other function in this file uses them. Prefer the module-level variables.

---

Otherwise: the `cachedEventsRaw` threading is correct and complete (all consumers updated, back-compat undefined path preserved), the `pendingTmpFile` hoisting in `writePendingBuffer` is sound, the `.tmp` cleanup additions in catch blocks are correct, and the test suite is comprehensive — particularly sub-test 7 (read-only `.agentic/`) and the multiple-failing-targets sort test. Appreciated.

tyson-solara6 (changes requested):

REQUEST CHANGES - the fail-open design is solid and the test suite is comprehensive, but one Missing invariant (CWD guard) must be addressed before merge, plus one security-adjacent minor.

**[Major] `recordTelemetryHealth` is missing the CWD traversal guard (`hooks/stop-context.js`).**
Every other writer function in this file (`writeLoopState`, `writeBatchState`, `writeContextMdOrSpill`, etc.) opens with a `path.resolve(cwd)` check before deriving a write path from `cwd`. `recordTelemetryHealth` derives `healthPath` directly via `path.join(cwd, '.agentic', '.telemetry-health.json')` without this validation. Current callers all pass a pre-validated `cwd` from `run()`, so it is not exploitable today - but the established module convention is that each writer validates its own input. Breaking this invariant creates a latent path-traversal risk if the function is called from a new site. Add the standard guard before `const healthPath = ...`:
```js
const resolvedCwd = path.resolve(cwd);
if (resolvedCwd !== cwd) return;
```

**[Minor] `_telemetry_health_line` interpolates `last_error` verbatim into terminal output (`bin/agentic-status`).**
`last_msg = str(info.get("last_error", ""))` takes the raw exception message written by the JS hook. The health file is writable by anyone with project directory access. A crafted `.telemetry-health.json` with ANSI escape sequences in `last_error` would manipulate terminal rendering on `agentic-status` output. Sanitize before interpolation: `re.sub(r'\x1b\[[0-9;]*[mGKHF]', '', last_msg)[:200]`.

**[Minor] Inline `require()` calls inside `recordTelemetryHealth` inconsistent with module style.**
`require('path').join(...)` and `require('fs').readFileSync(...)` inline when `path` and `fs` are already module-level bindings. Node caches so behavior is identical, but the inconsistency is noise. Use the module-level bindings.

**[OK] Fail-open guarantee confirmed.** Entire body wrapped in outer `try { } catch (_) { }`. Sub-test 7 (read-only `.agentic/`) proves hook exits 0. Guarantee holds.

**[OK] Race condition on concurrent exits is acceptable.** Read-modify-write is not atomic across processes - lost increments possible under simultaneous exits. Acceptable: this is best-effort observability, not a correctness mechanism.

**[OK] Gitignore, schema documentation, stacking, North Star alignment** all confirmed sound. The `agentic-status` health line directly serves pillar 1 (guard attention - failure summary on demand) and pillar 2 (verifiable outcomes - silent failures now inspectable).

Merge after #334 + #335, after the two issues above are resolved.

tyson-solara6:
Separate from the code review (the traversal-guard Major etc.), one direction-level thought measured against the North Star (`docs/overview/vision.md`, now committed).

The counter + `recordTelemetryHealth` plumbing is a genuine win and I want it in. Silent state-write failures today corrupt the substrate the trust model leans on (team telemetry attribution, cross-session resume, `context.md` continuity) with zero signal, and the fail-open / atomic-write implementation is solid. That advances pillar 2 (verifiable outcomes).

The piece I'd reshape before merge is the operator-facing surface. A constant `Telemetry health: OK / N failures` line is, in the vision doc's terms, surfacing *status* rather than a *decision or work-stoppage* (pillar 1 is the tie-breaker). It also doesn't fully close the loop #266 describes: a real failure still only surfaces when the operator runs `agentic-status` and reads it, which is the same "discover it late" latency, moved one layer up.

Suggested reshape, small on top of what's here: fire a single non-blocking line only when a high-value target (`session-log` / `events.jsonl`) crosses N failures, gated once per session. That matches the pattern we already use for the meta-divergence, provisional-identity, and capture-gap notices, and it turns "status to babysit" into a surfaced work-stoppage. `_telemetry_health_line` is already most of the way there.

Not a correctness blocker, just a direction nudge to keep this on the "don't make the operator watch it" side of the line.

bloodf:
Rebased onto the updated `perf/267-events-cache` branch, resolved conflicts, and addressed the review feedback:
- Added CWD traversal guard to the telemetry-health writer
- Sanitized `last_msg` in `_telemetry_health_line` (ANSI stripped, capped at 200 chars)
- Aligned `agentic-status` health line with the `targets` schema

All tests pass. Ready for re-review.

tyson-solara6 (approved):
**Approving** - Major finding (CWD traversal guard) is resolved; design is solid.

The prior `CHANGES_REQUESTED` raised two issues. Both are addressed:

**[Major - resolved] CWD traversal guard added to `flushHealth`**
```js
const resolvedCwd = path.resolve(cwd);
if (resolvedCwd !== cwd) return; // traversal component - skip silently
```
This is now present in `flushHealth` (the function that writes `.telemetry-health.json` in `stop-context.js`), alongside the existing `if (!cwd) return` null guard. This matches the pattern every other writer in the file uses.

**[Minor - resolved]** The fail-open design and try/catch wrapper are correct.

**New additions are sound:**
- `_telemetry_health_line(health_path)` in `agentic-status`: reads `.telemetry-health.json`, returns `telemetry-health: OK` when absent/all-zeros/unreadable (fail-open on any parse error), or a one-line summary when failures exist. Injected at the right position in the status output.
- Test suite: 8 new cases covering absent file, all-zeros, failures, multiple failing targets, malformed JSON, and default path. Sub-test 7 (read-only `.agentic/` → hook still exits 0) validates the fail-safe behavior end-to-end.
- Per-exit cost (~5 read-modify-writes of a sub-1KB file) is negligible compared to the multi-MB events.jsonl reads #267 removed.

All 13 CI checks pass. Merge after #334 and #335.

PR #394, opened 2026-07-02
fix(tests): DS-59 anchor extract_path_from_notice regex to current notice text (stale since #322)

tyson-solara6:
Fixes [DS-59](https://solara6.atlassian.net/browse/DS-59): hooks/tests/test-version-check-core-repo-dir.sh has failed on clean main since PR #322 added "/pull-and-install (in session)" to the update notice. The test's extract_path_from_notice sed pattern anchored greedily on "or " and captured the intervening text instead of the repo path.

Fix is test-only: re-anchor extraction on the stable "No PATH? " marker that sits directly before the path. Product code (hooks/lib/version-check-core.sh) untouched.

- Pre-fix: 5/10 assertions fail on clean checkout (captured "/pull-and-install (in session). No PATH? <repo>" instead of "<repo>")
- Post-fix: 10/10 pass; test-hooks-staleness-core.sh also pass (8/8)
- Grep confirms no other test carries the same extraction pattern
- Skeptic: round 1 approve with one Minor (comment rationale wording), fixed in follow-up commit

### DIFF
--- a/hooks/tests/test-version-check-core-repo-dir.sh
+++ b/hooks/tests/test-version-check-core-repo-dir.sh
@@ -63,11 +63,11 @@ write_fake_cache() {
 # Extract the repo path from the notice line:
 #   "... Run: agentic-update (shell) or /pull-and-install (in session). No PATH? <PATH>/update.sh"
 #
-# NOTE: do not anchor on "or " - the notice text has multiple "or " phrases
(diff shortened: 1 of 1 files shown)


tyson-solara6:
Superseded by #392, which fixed the same stale extraction regex and merged first (concurrent session). Main is green on this test.

PR #397, opened 2026-07-02
fix(install): DS-58 resolve gitdir for pre-commit hook when .git is a worktree pointer file

tyson-solara6:
Fixes [DS-58](https://solara6.atlassian.net/browse/DS-58): install.sh's pre-commit step ran `ln -s ... $REPO_DIR/.git/hooks/pre-commit`, which fails with "Not a directory" inside a git worktree where .git is a gitdir-pointer file. The identical block was duplicated in three installers (.claude, .cursor, .opencode).

## Fix
- New shared `scripts/lib/precommit.sh` (`install_precommit_hook`), sourced by all three installers the same way they source `identity.sh`. Resolves the real hooks dir via `git -C "$repo_dir" rev-parse --git-path hooks` (relative results normalized to absolute) - hooks live in the common gitdir for every worktree, which is exactly where git reads them. Also correctly honors `core.hooksPath` where the old hardcoded path did not.
- Behavioral parity for normal checkouts verified branch-by-branch against the old inline block (only additive change: `mkdir -p` on the hooks dir).
- Regression test `bin/tests/test_precommit_worktree.sh` (11/11): real `git worktree add` fixture, asserts .git-is-a-file, no "Not a directory", symlink lands in the shared common gitdir. CI-wired into adapter-sync.yml beside the DS-57 test per #393 precedent.

## Review
Skeptic round 1: approve with one Minor (CI wiring for the new test), fixed in the follow-up commit. Empirically verified: worktree resolution to common gitdir, core.hooksPath handling, cwd-independent normalization, non-repo soft-fail.

Known out-of-scope twin: the three uninstall.sh scripts share the hardcoded path but fail silently (no crash) - follow-up ticket to be filed.

Gates: build-all clean across 11 adapters (fix survives rebuild), shellcheck clean, neighboring bin/tests pass.

### DIFF
--- a/.claude/install.sh
+++ b/.claude/install.sh
@@ -42,6 +42,11 @@ export REPO_DIR
   echo "  ! scripts/lib/identity.sh not found - identity setup skipped"
 }
 
+# shellcheck source=scripts/lib/precommit.sh
+[[ -f "$REPO_DIR/scripts/lib/precommit.sh" ]] && . "$REPO_DIR/scripts/lib/precommit.sh" || {
+  echo "  ! scripts/lib/precommit.sh not found - pre-commit hook install skipped"
+}
+
 # ---------------------------------------------------------------------------
 # Activation mode (shared across all adapters)
 #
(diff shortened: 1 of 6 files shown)


tyson-solara6:
Superseded by #396 (concurrent session), which merged the same shared-resolver design first and goes further: symmetric uninstall_precommit_hook with ownership guard across all three adapters, 19-case regression test, CI wiring already in adapter-sync.yml. No unique value remains here.

PR #413, opened 2026-07-03
fix(agentic-team): resolve per-role model from team.yml at dispatch

bloodf:
Fixes the core "cross-harness dispatch only uses Claude models" bug.

## Root cause

`agentic-team dispatch --role <r>` without an explicit `--model` ran the worker on the **harness session default** (e.g. omp → `claude-opus-4-8`), ignoring the model configured for that role in `team.yml`. `_cmd_dispatch` only ever read `args.model`; it never consulted the loaded team config. So a fully populated `team.yml` had no effect unless every caller hand-threaded `--model`.

Proven live before the fix: `dispatch --harness omp --role engineer` (team.yml says `kimi/kimi-k2.7`) → the omp run log showed `"model":"claude-opus-4-8"`.

## Fix

`_cmd_dispatch` resolves the model when `--model` is absent:

```python
resolved_model = args.model or _resolve_role_model(role, harness)
```

`_resolve_role_model` loads the merged global+project `team.yml` (existing `_load_team_config`), looks up `roles[<role>]`, and returns its model **only when the entry harness (or `default_harness`) matches the dispatch harness** — so a row for a different harness never leaks its model. Returns `None` (unchanged: worker uses its session default) when there is no team.yml / no entry / no model / harness mismatch. Explicit `--model` always wins.

## Verification

- Live: role-only `dispatch --harness omp --role engineer` now forwards `kimi/kimi-k2.7` (omp log confirms), not opus.
- Tests: resolve-from-team.yml, harness-mismatch→None, scalar-row via default_harness, absent/no-model→None, argv carries resolved model, explicit-override precedence.
- Gates: check-methodology-drift OK, check-codex-skill-sync OK, `pytest bin/tests/` → **426 passed**. ruff + mypy clean.

### DIFF
--- a/bin/agentic-team
+++ b/bin/agentic-team
@@ -18,7 +18,12 @@ Public API: agentic-team <discover|dispatch|status|collect|configure> [options]
                          spawn a worker (background); prints run-id to stdout.
                          --model forwarded to every harness using its own flag
                          spelling (_MODEL_FLAG_BY_HARNESS: codex/gemini -> -m,
-                         all others -> --model).
+                         all others -> --model). When --model is omitted,
+                         the model is resolved from team.yml
+                         roles[<role>].model (honoring --global-config /
+                         --project-config) IFF the role's harness (or
+                         default_harness) matches the dispatch harness;
+                         otherwise the worker uses its own session default.
             status     <run-id> [--workdir <dir>]  - running|done|failed
             collect    <run-id> [--workdir <dir>]  - final message, demuxed per harness
             configure  [--non-interactive] [--assign role=harness:model ...]
@@ -311,26 +316,42 @@ def _role_entry(spec: object) -> dict:
     return {}
 
 
-def _resolve_role_model(role: str, harness: str) -> str | None:
+def _resolve_role_model(
+    role: str,
+    harness: str,
+    global_path: Path = GLOBAL_TEAM_YML,
+    project_path: Path = PROJECT_TEAM_YML,
+) -> str | None:
     """Resolve the model for *role* from team.yml when no explicit --model.
 
-    Reads the merged global+project team.yml, looks up roles[<role>], and
-    returns its model handle ONLY when the entry's harness matches the harness
-    actually being dispatched to (so a team.yml row for a different harness
-    never leaks its model onto this dispatch). Returns None when there is no
-    team.yml, no entry, no model, or a harness mismatch -- in which case the
-    worker uses its own session default, exactly as before.
+    Reads the merged global+project team.yml (honoring any --global-config /
+    --project-config overrides threaded from main), looks up roles[<role>], and
+    returns its model handle ONLY when the role's resolved harness matches the
+    harness actually being dispatched to. The role's harness is its own entry
+    harness, else default_harness; if NEITHER is set the config is ambiguous
+    and we return None rather than forwarding the model to an arbitrary
+    harness (a team.yml row with no harness and no default_harness must not
+    leak its model). Returns None when there is no team.yml, no entry, no
+    model, or a harness mismatch/ambiguity -- the worker then uses its own
+    session default, exactly as before.
     """
     try:
-        config = _load_team_config()
-    except Exception:
+        config = _load_team_config(global_path=global_path, project_path=project_path)
+    except (OSError, ValueError, ImportError) as exc:  # narrow: I/O + parse
+        print(
+            f"agentic-team: could not load team.yml for model resolution: {exc}",
+            file=sys.stderr,
+        )
+        return None
         return None
     spec = config.get("roles", {}).get(role)
     if not spec:
         return None
     entry = _role_entry(spec)
     entry_harness = entry.get("harness") or config.get("default_harness")
-    if entry_harness and entry_harness != harness:
+    # Require a POSITIVE harness match. If the role has no harness and there is
+    # no default_harness, entry_harness is None (ambiguous) -> do not forward.
+    if entry_harness != harness:
         return None
     model = entry.get("model")
     return model if model else None
@@ -1146,7 +1167,12 @@ def _cmd_dispatch(args: argparse.Namespace) -> int:
     # to the role's model in team.yml (D-1). Without this, a role-only dispatch
     # silently ran on the harness session default (e.g. omp -> claude-opus-4-8),
     # which is why cross-harness dispatch "only used Claude models".
-    resolved_model = getattr(args, "model", None) or _resolve_role_model(role, harness)
+    resolved_model = getattr(args, "model", None) or _resolve_role_model(
+        role,
+        harness,
+        global_path=Path(getattr(args, "global_config", None) or GLOBAL_TEAM_YML),
+        project_path=Path(getattr(args, "project_config", None) or PROJECT_TEAM_YML),
+    )
     argv = _build_worker_argv(harness, augmented_brief, model=resolved_model)
 
     stdout_file = run_dir / "stdout"
--- a/bin/agentic-team
+++ b/bin/agentic-team
@@ -311,6 +311,31 @@ def _role_entry(spec: object) -> dict:
     return {}
 
 
+def _resolve_role_model(role: str, harness: str) -> str | None:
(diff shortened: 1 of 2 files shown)


tyson-solara6 (changes requested):
## Review: Request Changes

Reviewed via skeptic pass (correctness/design/test-coverage) against the diff and the sibling PR #414 in the same subsystem.

- **Major** — `_resolve_role_model` calls `_load_team_config()` with no args, ignoring any `--global-config`/`--project-config` overrides passed to `main()`. A dispatch invoked with explicit config paths can resolve a model from an entirely different (stale/unrelated) `team.yml` than the one `main()` just validated. (`bin/agentic-team:319`)
- **Major** — The new "end-to-end" test (`test_dispatch_forwards_team_yml_model`, `bin/tests/test_agentic_team.py:1675`) never actually calls `_cmd_dispatch`/`main()` — it re-derives the precedence logic inline and calls `_build_worker_argv` directly, so the real integration point (`bin/agentic-team:1099`) has no test coverage.
- **Major** — Untested ambiguous-fallback: when a role entry has no `harness` and `team.yml` has no `default_harness` (a valid, non-erroring config), the model forwards unconditionally to *any* dispatch harness — directly contradicting the stated "never leaks its model" design goal. No test covers this.
- **Major** — Module manifest (header, lines 1-62) wasn't updated to document the new automatic team.yml-based model resolution behavior added by this fix.
- Minor — broad `except Exception: return None` in `_resolve_role_model` swallows config errors silently, compounding the path-mismatch bug above with no diagnostic.

Test coverage is solid for the explicit-harness-match/mismatch and `default_harness` cases, but misses exactly the three scenarios above.

No security concerns — the resolved model flows into a `subprocess` argv list, never shell-interpolated.

**Heads up:** PR #414 (open, same subsystem) defines `_resolve_role_model` at the identical call site with a different signature, and touches the same anchor line in `test_agentic_team.py`. Please coordinate merge order/reconciliation with that PR rather than relying on whichever lands first.

bloodf:
All findings addressed on the latest push.

**Major — config-path override:** `_resolve_role_model` now takes `global_path`/`project_path`, and `_cmd_dispatch` threads `args.global_config`/`args.project_config` into it — so a dispatch run with `--global-config`/`--project-config` resolves from the SAME team.yml `main()` validated, not the default global one.

**Major — untested/real integration point:** replaced the re-derived test with `test_dispatch_forwards_team_yml_model_end_to_end`, which drives the actual `dispatch` subcommand as a subprocess (project team.yml + an argv-recording fake `omp`) and asserts the worker was invoked with `--model kimi/kimi-k2.7` — covering `_cmd_dispatch → _resolve_role_model → _build_worker_argv` end to end.

**Major — ambiguous-fallback leak:** a role with no `harness` and no `default_harness` set `entry_harness=None`, so the old `and entry_harness != harness` guard was skipped and the model forwarded unconditionally. Now a **positive** match is required (`entry_harness != harness → None`). Added `test_resolve_role_model_no_harness_no_default_returns_none`.

**Major — manifest:** updated to document the automatic team.yml model resolution.

**Minor — silent swallow:** narrowed `except Exception` to `(OSError, ValueError, ImportError)` and now emits a stderr diagnostic.

**Re: the #413/#414 signature collision you flagged** — agreed. Proposed merge order: **#413 lands first** (the base resolver), then **#414 rebases** its rotation branch on top (its `_resolve_role_model` becomes this signature + the rotation branch). I will reconcile #414 against this once #413 is approved.

Gates: drift OK, codex-skill-sync OK, `pytest bin/tests/` → 427 passed.

bin/agentic-team
```diff
@@ -305,6 +310,47 @@ def _role_entry(spec: object) -> dict:
...
+    """
+    try:
+        config = _load_team_config(global_path=global_path, project_path=project_path)
+    except (OSError, ValueError, ImportError) as exc:  # narrow: I/O + parse
+        print(
+            f"agentic-team: could not load team.yml for model resolution: {exc}",
+            file=sys.stderr,
+        )
+        return None
+        return None
```
tyson-solara6:
```suggestion
        return None
```
The second `return None` is dead/unreachable code left over from combining the new early-return with the pre-existing except-block return.

tyson-solara6 (approved):
**Automated review.** Verified locally: ran the full `bin/tests/` suite from an isolated scratch copy of the PR head (no working-tree checkout) — all 427 tests pass.

This correctly fixes the bug where `agentic-team dispatch --role <r>` without an explicit `--model` ignored the per-role model configured in `team.yml`. `_resolve_role_model`'s harness-match logic is sound (requires a positive match before forwarding a model, refuses to leak a model when both role harness and `default_harness` are absent), and `--model` still takes precedence when passed explicitly. No blocking findings.

One small nit I'm attaching as a suggestion below: the new early-return in the `except` block leaves the old `return None` right after it as dead/unreachable code.

Two non-blocking notes for awareness (not filed as tickets — see below):
- `_cmd_dispatch` re-parses `team.yml` from disk via `_resolve_role_model` even though `main()` already loaded and validated the same merged config a few lines earlier — redundant I/O and a minor TOCTOU smell. Worth threading the already-loaded config down instead.
- The new `except (OSError, ValueError, ImportError)` branch (a malformed/unparsable `team.yml`) has no direct test coverage.

Also spotted two small pre-existing/out-of-scope items unrelated to this PR's responsibility (not blocking, and not filed as tickets since the Jira integration isn't available in this environment — flagging here so they aren't lost): resolved model strings from `team.yml` reach the worker's argv with no sanity check that they aren't flag-shaped, and `--role` has no `choices=` validation against the known role set (so an unknown role degrades silently rather than erroring). Low priority hardening, not urgent.

Note: this PR currently shows as `CONFLICTING` with `main` — will need a rebase before it can land, independent of this review.

bin/agentic-team
```diff
@@ -311,6 +316,47 @@ def _role_entry(spec: object) -> dict:
...
```
tyson-solara6:
```suggestion

```
Unreachable duplicate of the `return None` on the line above (left over from narrowing the except clause) — safe to delete.

tyson-solara6 (approved):
**Automated review — re-review after new commits.**

New commits since the prior approval are a rebase onto `main` (unrelated merged PRs, no action needed here) plus a small follow-up (`a2139000`) that finishes the nit from the last review. Re-verified the actual resolver logic at the current head:

- Config-path override (`--global-config`/`--project-config` threading into `_resolve_role_model`) — correct.
- Ambiguous-fallback leak fix (positive harness-match requirement) — correct, and the new `test_resolve_role_model_no_harness_no_default_returns_none` covers it.
- New end-to-end test (`test_dispatch_forwards_team_yml_model_end_to_end`) actually drives the `dispatch` subcommand as a subprocess with an argv-recording fake `omp` — good, this replaces the previous test that only re-derived the logic in-process.

No blocking findings. One small leftover from the earlier suggestion: the narrowed `except (OSError, ValueError, ImportError)` block now has its own `return None`, but the original bare `return None` right after it is still there too — dead/unreachable code (harmless, but worth a one-line cleanup, attached below as a suggestion).

Approving. Note this PR previously showed `CONFLICTING` against `main` — confirm it's still mergeable before landing.

tyson-solara6 (approved):
## Automated review — re-review after new commits

**Approving.** The one new commit since the last approval (`d5b05f2`) removes exactly the dead/unreachable `return None` flagged as a leftover nit in the prior round — nothing else changed. Verified line-by-line (12-line diff) and via an independent skeptic pass: no logic, tests, or scope changes; zero regression risk.

**Blocking for merge, not for this review's verdict — DCO check is currently failing on `d5b05f2`.** That commit was made through the GitHub web UI (`committer: GitHub <noreply@github.com>`) and its message has no `Signed-off-by:` trailer, so the required "DCO Signed-off-by check" is red and `mergeStateStatus` is `BLOCKED` even though the code is mergeable. To land this you'll need to add the sign-off — e.g. locally: `git commit --amend -s --no-edit && git push --force-with-lease`, or re-apply the same one-line fix via a normal `git commit -s`. Not adding the `ready-to-merge` label until that's green.

PR #415, opened 2026-07-03
feat(role-models): per-role reviewer models (reviewers.by_role)

bloodf:
Lets the adversarial reviewer model be chosen per **authored role**, not just globally.

## Motivation

`reviewers.pool` / `by_task` pick reviewer models globally for skeptic + security-auditor. There was no way to say "review engineer-authored code on gpt-5.5, but architect-authored design on opus."

## Design

New `reviewers.by_role` map, resolved **first**:

```yaml
reviewers:
  strategy: distinct-from-author
  by_role:
    engineer: cx/gpt-5.5
    architect: cc/claude-opus-4-8
  pool: [cx/gpt-5.5, glm/glm-5.2]
  fallback: cx/gpt-5.5
```

Pure resolver `resolve_reviewer_model(authored_role, author_model, reviewers, task_kind=None)` in `bin/_role_spec.py` codifies the precedence (previously prose-only):

```
by_role[role] -> by_task[kind]/by_task[default] -> pool[*] -> fallback
```

distinct-from-author is enforced at every step — a candidate equal to the author model is skipped and resolution falls through. Returns `None` when nothing resolves (reviewer uses its session default, unchanged).

## Verification

- Tests: by_role hit; miss→pool; by_role==author→fallthrough; mapping form; by_task after by_role miss (+default); pool→fallback; absent/empty→None.
- Gates: check-methodology-drift OK, check-codex-skill-sync OK, `pytest bin/tests/` → **427 passed**. ruff + mypy clean.


tyson-solara6 (changes requested):
## Review: Request Changes

- **Major** — `resolve_reviewer_model` ignores the `reviewers.strategy` field entirely. For `strategy: round-robin` it silently degrades to `distinct-from-author` behavior (always returns the first pool entry that isn't the author) — no actual rotation is implemented. No test exercises `strategy: round-robin` or `strategy: by-task`.
- **Major** — Stale module manifest (`bin/_role_spec.py` header) — doesn't list the new `resolve_reviewer_model` public function.
- **Major** — Undocumented interaction with PR #413/#414's team.yml-driven model resolution in the same `bin/` area: `_resolve_role_model` (team.yml, dispatch-time) has no distinct-from-author guarantee at all, so cross-harness-dispatching `skeptic`/`security-auditor` via `agentic-team dispatch` could still resolve to the author's own model, silently bypassing the exact guarantee `reviewers.by_role`/`pool` exists to provide. Not addressed in any of the three PRs.
- Minor — no validation of `authored_role`/`task_kind` against known enums (typos silently fall through to pool/fallback); a few test gaps (empty pool, non-string author, strategy interaction); `resolve_reviewer_model` currently has zero call sites (matches the pre-existing pattern where `pool`/`by_task` also have no runtime consumer today, so not a new regression, but worth noting the schema currently has no enforceable runtime effect).

Test coverage is good for the happy-path precedence chain (7 new tests, all passing, docs fully synced across all adapter mirrors) but doesn't touch `strategy` at all, which is the crux of the first Major finding. No security concerns — no code-execution/shell-interpolation surface in this diff.

Given PR #413/#414 are open in the same subsystem, recommend explicitly reconciling all three before any of them merge — right now there are two independent, inconsistent answers to "which model does this role use," with no stated precedence between them.

bloodf:
Fixed on the latest push.

**Major — strategy ignored:** `resolve_reviewer_model` now honors `reviewers.strategy` (default `distinct-from-author`):
- `by-task`: `by_task[task_kind]` → `by_task['default']` → distinct pool/fallback.
- `round-robin`: rotates the pool from a caller-supplied `rotation_index` (mod len), **skipping the author's own model**, then fallback. New `rotation_index` param carries the durable counter (the conductor/agentic-team supplies it); `None` degrades to distinct-from-author ordering.
- `by_role[<role>]` is still checked FIRST regardless of strategy (the per-role override).
distinct-from-author guarantee is applied at every step across all strategies.

**Major — manifest:** `bin/_role_spec.py` header now lists `resolve_reviewer_model` (+ KNOWN_HARNESSES/KNOWN_ROLES).

**Minor:** the `role-models.md` round-robin note is aligned to the implementation — it now **skips the author** for safety rather than "ignoring author identity."

**Re: the #413/#414/#415 cross-PR interaction you flagged** (team.yml `_resolve_role_model` has no distinct-from-author guarantee, so a cross-harness `skeptic`/`security-auditor` dispatch could resolve to the author's model): agreed this is the real reconciliation point. `reviewers.by_role`/`pool` (this PR, role-models.yml layer) and team.yml `roles[*]` are two different layers; wiring `resolve_reviewer_model` into the dispatch path for reviewer roles is the follow-up that ties them together. This PR makes the reviewer resolver correct and testable so that wiring has something correct to call; I'll do the dispatch-side wiring as a focused follow-up once #413/#414 land (it depends on both).

Tests: by-task (+default+no-kind), round-robin rotates 0,1,2,0, round-robin skips author, None-index degrades, by_role overrides strategy, default unchanged. Gates: drift OK, codex-skill-sync OK, `pytest bin/tests/` → 433 passed.

tyson-solara6 (changes requested):

The `resolve_reviewer_model` resolver in `bin/_role_spec.py` is a good idea and the documented happy paths are well tested (all 38 tests in the changed test file pass), but the PR as submitted has real defects:

**Blocking:**
- `bin/_role_spec.py:175-211` — Major. A full duplicate of the resolver's logic (a second `_model_of` redefinition, re-derivation of `author`, steps 1-4) is left in the function body, unreachable after the real implementation's `return None` at line 173. Looks like a merge artifact from combining two commits — the rewrite should have replaced this block, not been prepended to it. Please delete lines 174-211.
- `bin/_role_spec.py:112-116` (`_model_of`) — Major. Any malformed (non-`str`/non-`dict`) entry in `pool`, `by_role`, `by_task`, or `fallback` (e.g. a YAML typo like `by_role: {engineer: 5}`) raises an uncaught `TypeError` from `normalize_role_spec`, contradicting the function's own docstring contract ("returns None, never raises"). Verified directly: `resolve_reviewer_model('engineer','author',{'pool':[123,'cx/gpt-5.5']})` raises `TypeError`. Please wrap the `normalize_role_spec` call in a try/except (or pre-filter non-str/non-dict entries) so malformed config entries are skipped rather than crashing resolution.
- `bin/_role_spec.py:150-155` vs `content/references/role-models.md:97` (and its 5 mirrors) — Major. The documented precedence doesn't match the shipped code for two of the three strategies: the `by-task` doc bullet says nothing about falling through to `pool`/`fallback`, but the code does; the `distinct-from-author` bullet says nothing about consulting `by_task`, but the code does. Since the PR description states this resolver "codifies the precedence (previously prose-only)," code and prose should agree — please either narrow the code to match the docs, or update all six doc copies to describe the actual behavior, and add tests for the previously-undocumented fallthrough paths.

- Module manifest overstates `resolve_reviewer_model`'s current callers — it's not yet wired into `agentic-configure`/`agentic-team` (only `normalize_role_spec` is imported there today). Worth a one-line clarification so a future reader doesn't assume it's already load-bearing.
- An invalid/typo `strategy` value silently falls into `distinct-from-author` with no warning — worth a docstring note.
- A couple of test assertions have autoformatter-artifact line breaks (`bin/tests/test__role_spec.py:363-368`) — cosmetic.

**Not blocking, flagged for awareness** (pre-existing, out of scope — not filed as tickets since Jira integration isn't available here): the shared `role-models.md` preamble ("consulted ONLY on Pi/omp harnesses") is shipped verbatim inside the Codex/Gemini/etc. mirrors, i.e. a Codex doc that states Codex ignores it — confusing but predates this PR.

**Test coverage:** the 38 tests cover the documented happy paths well, but there's zero coverage for malformed config (the crash above) or for the undocumented by-task→pool→fallback fallthrough — both gaps line up exactly with the blocking findings.

tyson-solara6 (approved):
**Automated review — re-review after new commit.**

The single fix commit (`3be562f`) correctly resolves both blocking findings from the prior review:

- The dead duplicate `_model_of` redefinition and unreachable steps 1-4 (the merge-artifact block) are gone; the real implementation at the top of the function is now the only one.
- `_model_of` wraps `normalize_role_spec(spec)` in `try/except (TypeError, ValueError)` and validates the result is a `dict` before calling `.get("model")` — malformed `pool`/`by_role`/`by_task` entries (e.g. `{'pool': [123, 'cx/gpt-5.5']}`) now return `None` instead of crashing.

Verified the logic reads correctly against the described PoC (`resolve_reviewer_model('engineer','author',{'pool':[123,'cx/gpt-5.5']})` → `None` rather than a raised `TypeError`).

Minor, non-blocking: `except (TypeError, ValueError) as exc:` binds `exc` but never uses it — harmless, could drop the `as exc` binding. Not worth blocking on.

No new issues found. Approving.

PR #423, opened 2026-07-07
docs(ds55): define absent-key defaults for all under-specified config settings

tyson-solara6:
## Summary

Audit and clarify six user-facing settings that previously showed `required`, `optional`, or `derived from GitHub login` instead of an explicit absent-key default.

## Changes

**`docs/configuration-reference.md`**
- `developer_id`: changed from `required if file exists` → `none (absent file = no attribution)`; added prose explaining `agentic-identity auto` as the on-ramp
- `team.yml enabled`: changed from `required` → `false (absent file = feature off)`
- `team.yml default_harness`: changed from `optional` → `none (absent = no harness fallback)` with explanation
- `team.yml roles`: changed from `optional` → `none (absent = empty map, no per-role routing)`

**`content/references/risk-config-and-tiers.md`**
- `model_profile`: added `absent-key default: "default"` inline to the prose (the seeded template already had this; the reference doc did not)

**`content/references/cross-harness-teams.md`**
- Schema table: added a `Default (absent key)` column to all fields; clarified that absent file = feature off, absent `default_harness` = native spawn fallback, absent `roles` = empty map

**`docs/index.html`**
- Identity card: corrected `Default: derived from GitHub login` → `Default: none (absent file = no attribution; run agentic-identity auto to derive from GitHub login)`

**`README.md`**
- `model_profile` entry: added `default "default" (absent key resolves to "default")`

**Adapter copies** (`.codex`, `.cursor`, `.gemini`, `.hermes`) auto-updated by pre-commit.

## No behavioral changes
Docs-only. No adapter rebuilds, no logic touched, no seeded template changes needed (template already has `model_profile: "default"`).

Closes DS-55

PR #420, opened 2026-07-06
DS-68: compress force-loaded methodology resident set 29.5% (124KB -> 87KB) for weaker-model consumers

tyson-solara6:
## Summary

- Compresses the force-loaded methodology resident set (built `METHODOLOGY.md` + `rules/code-standards.md` + `rules/conventions.md`) from **123,942 B (~31k tokens) to 87,323 B (~21.8k tokens) - a 29.5% reduction** - by relocating detail into the trigger-loaded `content/references/` tier. Primary beneficiaries are weaker-model consumers (Haiku/Sonnet tiers, Kimi, Gemini, Cursor-hosted) that pay the resident-context cost hardest.
- Structure over terseness: no resident rule was densified or deleted. Detail moved behind pointers extending the existing `12-protocol-details.md` read-on-trigger pattern; worked examples were relocated, never dropped; every relocated rule keeps its bolded lead phrase byte-for-byte (machine-verified).
- Zero dropped rules, machine-proven: new `scripts/extract-rule-inventory.py` (Counter-based multiplicity - catches deletion of one of N duplicate copies) reports **0 UNMAPPED anchors / 0 unaccounted lines** against the pre-edit snapshot (1,644 anchors / 3,558 lines), with a 31-entry audited mapping (26% Skeptic spot-checked).
- Behavioral gate (Haiku, SWE-bench-lite probe, pre-registered noise envelope): **operator-graded, pending** - per the RUNBOOK contract (`.agentic/ds68/RUNBOOK.md` §5) G3 never auto-passes or auto-fails, so meeting the numeric arms does not stamp a PASS. Measured arms: correctness floor 3/3 ok reps at pass_fail 1.0; cost median $0.5553, within the $0.5646 band (envelope median 0.53974 + 0.025 noise floor) but +2.9% vs the pre-edit resident median - inside the noise floor yet opposite the RUNBOOK's expected decrease. The pass/fail grade is the operator's to assign; recorded here as pending the operator's G3 grade (flagged, not hidden).
- New CI ratchet `resident-budget` fails any PR that regrows the resident set past **88,323 B** (measured actual + 1,000 headroom).
- `scripts/build-methodology.sh` now strips per-file HTML-comment manifest headers from the built output (mirrors the existing `.claude/build.sh` SKILL.md strip).
- All 11 adapters rebuilt; `methodology-drift` baseline regenerated; build idempotent.

## Per-file resident changes (source bytes)

| File | Before | After |
|---|---|---|
[...]

PR #431, opened 2026-07-08
feat: dormant invoke-first install + supervised multi-model teams + TUI installer

bloodf:
## Summary

Two pillars in one branch:

**Zero-tax install (dormant by default).** Installing DinoStack today force-loads ~124KB of methodology into every session of every project (Claude `@`-imports; Codex/Gemini/Kimi full-file symlinks), and hooks fire everywhere regardless of whether the project uses DinoStack. This PR makes DinoStack dormant by default across all 11 harnesses: only a small stub is resident until explicitly activated via a new `/ds` command, `.agentic/` marker auto-detection, or an opt-in allowlist. Hooks become self-gating (pure-stat guards, <10ms, fail-active on error) so a dormant project pays zero context/side-effect cost.

**Reliable multi-model external-harness teams.** `agentic-team` dispatch previously had watchdog coverage only for cursor-agent, a decorative (unwired) `timeout_seconds`, no retry/failover, and silent crash detection. This PR generalizes supervision to all 9 dispatch harnesses (`bin/_supervise.py`), wires `team.yml`'s `dispatch.{timeout_seconds,stall_seconds,retries,failover}` block, adds a same-harness-retry → rotation-model → native-claude failover chain, and a new `agentic-team status`/`--json` for live visibility.

**Plus:** a pure-bash TUI installer (`scripts/install-tui.sh`) for harness multi-select, mode+tier, team/model seeding, and profile targets — zero deps, bash 3.2 compatible, falls through silently to the existing flag-based flow outside an interactive TTY.

Supersedes and absorbs 7 open PRs: #413, #414, #415, #416, #417, #422, #423. (#420, resident-set compression, is excluded — it's a draft and will rebase onto this after merge.)

## Commits

1. `feat(agentic-team): resolve per-role/rotation models, add supervised failover + status, activation core` — bin/ portion of units 0-4, 9a-9d, 10
2. `feat(hooks): self-gating activation guards, exit dormant in <10ms with zero side effects` — unit 10, hooks/ portion
3. `feat(methodology): 3-tier system, dormant stub content, absent-key doc defaults` — units 6, 8, 10 (content/docs), 14, 15
4. `feat(adapters): dormant/resident wiring, tier support, TUI-composable flags across all 11 harnesses` — units 5, 6, 10, 11 (adapter portion + generated artifacts)
5. `feat(scripts): pure-bash TUI installer, profile-aware installs, dormancy/stub build libs` — units 5, 6, 7, 11 (scripts/ portion)

## Test plan

- [x] `pytest bin/tests/` — 510 passed
- [x] `ruff check bin/` — clean
- [x] `mypy bin/` — clean
- [x] `scripts/check-methodology-drift.sh` — full/medium/minimal all OK
- [x] `scripts/check-codex-skill-sync.sh` — OK
- [x] `scripts/check-stub-budget.sh` — OK (1078 ≤ 2048 bytes)
- [x] `scripts/build-all.sh` run twice — byte-identical (idempotent)
- [x] `bin/tests/test_activation.sh` — fake-$HOME dormant/resident matrix, all pass
- [x] `hooks/tests/test_activation_guard.{py,js,sh}` — guard fail-active + timing contract
- [x] `bin/tests/test_install_tui.sh` — TUI flag composition, non-TTY fallthrough
- [x] Zero `.rej`/`.orig` artifacts across the tree
- [ ] Manual smoke: one tiny-brief dispatch per available external harness
- [ ] Manual smoke: `/ds activate` → `/ds status` → `/ds deactivate` round-trip in a scratch project

## North Star alignment

Advances **Guard operator attention**, **Verifiable outcomes autonomously**, and **Universality** without regressing the others. Dormant-by-default means a project that has not opted in pays zero context and zero hook side-effects (self-gating pure-stat guards, <10 ms, fail-active) — nothing to watch and nothing to babysit — directly serving the attention test. The `agentic-team` supervision/failover work drives external-harness runs to a checkable `done`/`failed` outcome with `status --json` visibility, serving verifiable autonomy. Activation resolves via runtime markers (`.agentic/active`), auto-detection, or an opt-in allowlist — never a hardcoded operator, tracker, or workspace — so it passes the portability test.

Trade-off: more machinery (activation guards, `bin/_supervise.py`, the retry→rotation→native failover chain) in exchange for lower default cost and higher reliability. This is accepted because the default path gets *quieter* (attention test) and the added concurrency/IO paths (`_resolve_run_dir`, `atomic_write`) are now validated and race-free, keeping outcomes verifiable.


tyson-solara6 (changes requested):

**Verdict: Request changes.**

This is a large, brand-new PR (305 files, ~54k lines). Given the size, review was scoped to the hand-written, behavior-bearing code (new Python modules, install/bootstrap scripts, enforcement-hook wiring) rather than the adapter-mirror files, per standard practice for this kind of change. Both a correctness pass and a dedicated security audit were run.

## Security findings

1. **Critical — unauthenticated kill-switch for all six enforcement hooks** (`hooks/lib/activation.py`). `is_active()`'s activation precedence puts an explicit tombstone file, `<cwd>/.agentic/dormant`, ahead of "`.agentic/` dir exists = active" — verified directly in the source: rule 3 (`dormant` file exists → `False`) is checked and returned *before* rule 4 (`.agentic/` dir exists → `True`). This PR wires `is_active(cwd)` as the first gate in all six enforcement hooks (`enforce-no-abdication.py`, `enforce-orchestrator-singularity.py`, `enforce-tier.py`, `enforce-askuserquestion-default.py`, `enforce-planning-artifact-spawn.py`, `enforce-background-spawn.py`). The check is pure `os.path.exists` — no signature, no ownership check, no confirmation, no log entry. Any process with ordinary write access to the repo (a compromised dependency, a malicious skill, or — concretely — a prompt-injected instruction telling the agent to "clean up" or "touch `.agentic/dormant`") can silently disable the entire mechanical guardrail layer for the rest of the session with a single zero-byte file write. This directly undermines the methodology's "hooks mechanically enforce X" guarantee. Recommend requiring an authenticated write path for the tombstone (e.g. only via a controlled `/ds deactivate` command path with some integrity marker) or, at minimum, emitting a loud, unsuppressable notice the first time any hook observes a fresh dormant transition.

2. **Medium — path traversal via unvalidated scripted-install label** (`scripts/install-tui.sh`). In `INSTALL_TUI_SCRIPT` mode, `_ask_multiselect` splits the answer-file token on commas into `SELECTED` with no validation against `H_LABELS` (confirmed directly in source — the `else` branch of the scripted case just does `for tok in ${ans//,/ }; do SELECTED+=("$tok"); done`). That value flows into `adapter_script()` (`printf '%s/.%s/install.sh' "$REPO_DIR" "$1"`) and then `COMPOSED=(bash "$script")`, executed directly. A label like `../../../tmp/evil` traverses outside the intended `.{label}/install.sh` path and executes whatever script exists there. Exploitability depends on who controls the `INSTALL_TUI_SCRIPT` answer file (usually a CI-local trust boundary), but if that file can ever be sourced from an external or PR-provided location, this becomes a real supply-chain vector. Fix: validate scripted harness tokens against the `H_LABELS` allowlist before use (the interactive path already does equivalent validation for tenant names).

3. Low/informational: `bootstrap.sh` clones over HTTPS with no commit/tag pinning or checksum — standard for this class of installer, not blocking.

Positive controls noted: `bin/_lib.py`'s `atomic_write` defaults to `0o600`; the interactive tenant-name path validates against `^[a-zA-Z0-9_-]+$`; the new Python modules use stdlib-only JSON parsing (no `yaml.load`, no `pickle`, no `subprocess(shell=True)`, no `eval`).

## Correctness findings

4. **Major — race condition in failover sentinel liveness** (`bin/agentic-team`, `_spawn_attempt`/`_run_chain`, ~lines 1299-1395). `run_dir/"exit"` is written only once the whole failover chain terminates, not per attempt, while `run_dir/"pid"` is overwritten per attempt. Between one attempt dying and the next attempt's `Popen`, the pid file holds a stale dead PID with no `exit` file present. `_is_live_readonly()` (used by `agentic-team collect`'s sibling-gated `.agentic/teamrun/.active` cleanup) treats "no exit file + dead pid" as not-live, so a concurrent `collect` on a sibling run can prematurely unlink the shared `.active` sentinel while this run is still retrying/failing over — letting the background-spawn-suppression hook stop suppressing native spawns mid-supervision. Suggest gating `_is_live_readonly` on an explicit in-progress state in `status.json` (already written per-attempt) instead of raw pid liveness across attempt boundaries. Related: the comment at line ~1402 claims the `.active` sentinel is "refreshed after each attempt," but the write actually happens once after the loop — the documented mitigation isn't implemented.

## Minor / nits (not blocking, worth cleaning up in the same PR since they're small)
- `bin/_supervise.py` docstring says agentic-team integration "happens in a later wave" — already false, it's wired in this PR.
- `bin/agentic-update:576` — stray leftover word "ponytail:" in a comment, should be "note:".
- Test gaps: `test_activation.py` doesn't exercise the 5th precedence layer (`allowlist`); `test_install_tui.sh` only covers `DRY_RUN=1` (no mocked-failure or malformed-answer-file path); no dedicated unit test file for `bin/_role_spec.py`'s public surface.

## Out of scope (pre-existing, not from this PR)
- `scripts/install-profiles.sh` was in scope for review but is untouched by this PR at head — not actually new.

Given the Critical security finding (guardrail bypass via unauthenticated dormant marker) and the Major correctness finding (sentinel race), please address at least those two before merge. The new dormancy/supervision primitives themselves (`_activation.py`, `_supervise.py`, `_role_spec.py`) are otherwise clean, well-tested, and free of injection/unsafe-deserialization issues.

tyson-solara6 (approved):
Automated re-review (agentic-engineering PR-review pass) — re-reviewing only the commits pushed since my prior review (`a6dd1a247`..`f7fce74b7`), since the base of the PR was already fully audited.

**Verdict: Approve.**

## Prior findings — verification of fixes

All three items I raised in the previous review pass have been independently re-verified (security-auditor + a dedicated correctness check) against the current head:

1. **Critical — unauthenticated dormant-tombstone kill-switch**: RESOLVED. `hooks/lib/activation.py` now emits a one-time, unsuppressable `AGENTIC-ENGINEERING DORMANT: ...` notice to stderr the first time a project is observed dormant via the tombstone, tracked via a marker file under `~/.agentic/` (outside the project tree, keyed by `sha256(realpath(cwd))`) — a repo-write-only attacker can't predict, delete, or pre-fake that marker. Exception handling is double-guarded so a marker-write failure can never flip the guard to fail-active or crash `is_active()`. This was the "at minimum" remediation I suggested and it's implemented correctly. (Minor nit, not blocking: `_dormant_notice_path`'s exception fallback returns one shared path for *any* project whose `realpath()` call raises, which could theoretically cause one project's already-fired notice to suppress a different project's first notice in that rare error path — worth a one-line comment/tracking note.)

2. **Medium — path traversal via unvalidated scripted-install label**: RESOLVED. `scripts/install-tui.sh`'s scripted-answer branch now validates every comma-split token against the hardcoded `H_LABELS` allowlist before use, dropping and warning on unknown labels (`../../../tmp/evil` and friends). Confirmed the allowlist itself is never attacker-influenced, and traced downstream consumption (`adapter_script`, `COMPOSED=(bash "$script")`) to confirm no other path reaches these labels unvalidated. A new test (Test 7) exercises the traversal string directly. No bypass found.

3. **Major — race condition in failover sentinel liveness**: RESOLVED. `_is_live_readonly()` now gates on `status.json`'s `state` (running/retrying/failed_over vs done/failed) instead of raw pid+exit-file, with the legacy check only reachable for pre-migration run dirs (verified fresh runs always write status.json before the pid file, so the fallback branch is unreachable in practice). The `.active` sentinel refresh now genuinely happens per-attempt inside the loop, matching what the code comment already claimed. New tests deliberately pair in-progress/terminal states against *contradicting* pids, so they'd fail against the old pid-only logic — not vacuous.

## New finding (not blocking, recommend a fast-follow)

While re-verifying #3, found an adjacent gap in the same subsystem: `_run_status()` (used by `agentic-team status` and by `_cmd_collect`'s own-run "is this run still running" gate) was **not** updated to the same status.json-aware logic — it still trusts raw pid+exit-file only. During the narrow transition window between one failed attempt and the next attempt's `Popen` (status.json already says `retrying`/`failed_over`, previous pid dead, no exit file yet), calling `agentic-team collect <run-id>` on that *same* run (not a sibling) will hit `_run_status`'s "pid dead, no exit file" branch, which writes `exit=1` and reports `"failed"` — even though the run is about to retry/failover and may still succeed. The exit file self-corrects if/when a later attempt finishes, but a caller polling `collect` during that window gets a premature/incorrect failure. This is the same bug class as the already-fixed sibling-gating race, just left in the "check my own run" path instead of the "check a sibling" path. Given the narrow timing window, self-correcting exit file, and that this doesn't affect security, I'm not blocking merge on it — but recommend applying the same `read_status()`-first logic to `_run_status()` in a quick follow-up, since it's the same file/PR that just fixed the sibling-side equivalent.

## Other
- The other two minor nits from the prior review (`bin/agentic-update` stray "ponytail" comment, stale `_supervise.py` docstring) are also fixed.
- CI is fully green (18/18 checks) at the current head.
- Note: this PR currently shows a merge conflict with `main` (`mergeable: CONFLICTING`) — rebase/resolve before merge; not something this review can fix.

Given all originally-blocking findings are resolved with test coverage, and the one new item found is narrow and non-security, approving.

tyson-solara6 (changes requested):
## Automated review — requesting changes

Reviewed with a general-correctness pass plus a dedicated security audit of the install/hook/dispatch surface (the two riskiest areas of this diff — install-time and hook-execution code). Given the diff size (286 files, ~76k lines), review focused on the canonical source tree (`bin/`, `hooks/`, `scripts/`, `content/`) and trusted CI's `check-adapter-sync`/`check-drift`/`check-codex-skill-sync` (all green) for the ~250 generated per-harness mirror files rather than re-reading them by hand.

### Correctness

- **Major — verify before merge:** `bin/agentic-team` removes `_cursor_watchdog` (the unconditional 300s kill of `cursor-agent`, explicitly documented as a hang workaround) in favor of generic stall-based detection in `_supervise.py` (kill only after 300s of *no new stdout/stderr bytes*). If the original hang emits any periodic output (spinner/heartbeat) while otherwise stuck, the new logic won't detect it — the process could now run up to the new 1800s hard timeout instead of being killed at 300s. Please confirm this against the actual hang symptom that motivated the original watchdog, or keep a tighter cursor-agent-specific stall floor.
- **Minor/Major:** `_cmd_dispatch`'s coercion of `retries`/`stall_seconds` (`int(str(retries_setting))`) has no try/except, and `_validate_config` never validates the `dispatch:` block's types. A malformed `team.yml` (e.g. `retries: 1.5` or a garbage string) crashes `dispatch` with an unhandled `ValueError` instead of a clean config-parse error.
- **Minor:** `hooks/skill-auto-load-check.sh` shells out to `python3` just to read `cwd`, which breaks the "pure-stat guard, <10ms" characterization claimed for the dormant-path guards elsewhere in this PR. Other `.sh` guards extract `cwd` with shell-native string ops — worth matching that pattern here.
- **Minor:** `_is_active()` bootstrap (~20 lines) is duplicated verbatim across 5 `hooks/enforce-*.py` files — acceptable if this mirrors an existing self-containment convention for standalone hook scripts, but flagging for awareness.
- **Minor:** `activation.py`'s dormant notice re-fires on every call until it successfully writes its tracking marker (spam risk if `~/.agentic` is unwritable), and only the Python guard implements the notice at all — the `.sh`/`.js` guards at the same layer are silent, so notice frequency depends on which hook happens to fire first.
- **Minor (latent):** `_supervise.py`'s `supervise()` can report `state="killed"` even when the post-kill `proc.wait(timeout=5.0)` times out (unkillable/D-state process) — status.json would claim killed while the process may still be alive.
- **Minor (test gap):** no test covers malformed `dispatch.retries`/`stall_seconds`, or the "kill signaled but process didn't actually die" path above.

### Security (dedicated audit — no Critical/High findings)

- **Medium:** The unsuppressable dormant notice fires for the explicit `.agentic/dormant` tombstone, but the far more common "never activated" case (layer 6, plain `False`) returns silently with zero output across all enforcement hooks. A user on dormant-mode install working in a fresh, unactivated repo gets no indication that guardrails aren't running.
- **Medium:** The new retry → rotation-model → native-claude failover chain reuses the same brief across harnesses on failure. Since several harnesses run with all interactive permission gates disabled, a single adversarial/prompt-injected brief that fails on harness A now gets additional unattended executions on harness B/C with no additional consent. Consider capping automatic failover to harnesses at the same or stricter sandbox posture, or requiring explicit opt-in for cross-harness failover.
- **Minor:** `_rotation_cursor_next`'s filename sanitization (`role.replace(...)`) doesn't fully match the completeness implied by its "same pattern as `_make_run_id()`" comment — low exploitability since `role` is drawn from a known/trusted set, but worth tightening.
- Positive: no `shell=True`/`eval`/string-built shell commands anywhere in the reviewed surface; all subprocess calls use argv lists; TUI installer inputs are enum/regex-validated; activation-guard fail-active semantics verified correct in all three (`.py`/`.js`/`.sh`) implementations; no secrets/tokens found in the diff.

### Cross-PR conflict — please resolve before merging any of the three

Confirmed via file-list diff: **100% of #414's 9 changed files are also touched here**, implementing the identical round-robin model-rotation feature in `bin/agentic-team`. **204 of #422's 286 changed files overlap with this PR's 286**, covering the same tier-system scaffolding (`agents-medium/*`, `agents-minimal/*`, `install.sh`, `build.sh`). This isn't adjacent-file overlap — it's the same code regions. Merging #414 or #422 independently while this is open risks direct conflicts or silently-diverging duplicate implementations. Recommend deciding merge order explicitly (this PR first with #414/#422 closed or rebased, or vice versa) rather than merging any of the three in isolation. Note: independent review of #414 found a **Critical** bug in the shared rotation-cursor logic (cursor doesn't actually persist across separate dispatches) — if this PR carries the same code path, please verify it isn't inherited here too.

No pre-existing out-of-scope issues were found with enough confidence to file as separate tickets.

tyson-solara6 (changes requested):
## Automated review — requesting changes

Re-reviewed at head `5ee9d069` after the fix commits pushed since the last (2026-07-08) round, with a correctness pass plus a dedicated security spot-check (mirroring the previous round's structure).

### Prior findings — status

- **Major — cursor-agent watchdog gap**: **Fixed.** New `_supervised_timeout_seconds()` caps `cursor-agent` dispatches at `min(timeout_seconds, 300.0)` regardless of `team.yml`. Traced `supervise()`'s `check_progress()` directly: the wall-clock `timeout_seconds` check is independent of and evaluated ahead of the stall/heartbeat check, so a cursor-agent process emitting periodic output while stuck is still killed at 300s rather than floating up to the new 1800s ceiling. New test (`test_cmd_dispatch_cursor_agent_timeout_capped`) verifies the value is actually threaded through even when `team.yml` requests 3600s.
- **Minor/Major — unhandled `ValueError` on malformed `dispatch:` config**: **Fixed.** `_validate_config` now validates all five `dispatch:` fields with correct type coercion (including explicitly rejecting `bool`/`float` for `retries`), gated in `main()` before dispatch runs; `_cmd_dispatch` also wraps coercion in try/except as defense-in-depth. Covered by new parametrized tests across all 5 fields.
- **Minor — `_supervise.py` could report `state="killed"` for an unkillable/D-state process**: **Fixed.** New `"unkillable"` status, correctly wired into both `_is_live_readonly` and `_run_status` as a terminal state.
- **Minor — `python3` shell-out in `skill-auto-load-check.sh` broke the "<10ms pure-stat guard" claim**: **Fixed**, with one narrow non-blocking caveat (see Security section below).

### New finding — dangling section anchor (Major)

`content/agents-medium/architect.md` and its `.claude/` mirror reference `content/references-medium/delegation-detail.md §Named agents for spawn protocol` — that file exists (143 lines) but has no `Named agents` heading anywhere (its actual sections are Open Questions and Deferred Defaults, Worker Autonomy Contract, Stop-Frequency as Planning Signal, Common Rationalizations to Reject, Investigator-Before-Architect Rules, Learnings Pipeline, Worker Preamble and Execution Contract Template, Digest-Return Discipline). This is the same class of dangling-anchor defect flagged and required to be fixed in the sibling PR #422 (which touches ~204 of the same files) — for consistency this needs the same treatment here: either add a `## Named agents` heading with the relevant content, or repoint the reference. (Note: this PR's *other* two medium-tier reference fixes — `skeptic.md` → `content/references-medium/skeptic-protocol.md`, and `debugger.md`'s self-contained note about `05-qa-gate.md` — are both correctly done and don't have this problem; only the `architect.md` reference is affected.)

### Security spot-check (no Critical/High; all non-blocking)

- Both previously-reported Medium findings remain **open and unchanged** by this round's commits (neither regressed nor fixed): (a) the far more common "never activated" case still returns silently with zero output across enforcement hooks (only the explicit `.agentic/dormant` tombstone case got a rate-limited notice this round); (b) the cross-harness failover chain still reuses the same brief verbatim across every harness/model attempt with no consent gating between sandbox postures.
- Two new narrow, non-blocking findings in the new sed-based `cwd` extraction (`hooks/skill-auto-load-check.sh` + `.codex/` mirror): it's not JSON-structure-aware, so a later/nested `"cwd"` key in the payload silently wins over the intended top-level value; and the documented "unparseable cwd → fail-ACTIVE" guarantee only holds for the fully-empty-extraction case — a non-empty-but-mangled extraction (e.g. a path containing a literal embedded quote) can silently resolve to fail-dormant instead. The realistic prompt-injection vector (forging a `cwd` value via injected text in another field) does **not** work, since JSON string-escaping breaks the regex match — verified directly.
- Informational: `hooks/enforce-no-abdication.py` was named in a commit message as one of the hooks wired with the shared activation guard, but has no `_is_active`/activation import at all — it uses its own independent gate that doesn't respect an explicit `.agentic/dormant` tombstone the way the other five enforcement hooks now do.

None of the security items are blocking (consistent with the prior round's non-blocking Medium classification), but please fix the new Major dangling-anchor finding before merge.

tyson-solara6:
**North Star alignment check (direction signal, not a verdict change):** MISALIGNED — the two known non-blocking security findings both cut against the North Star: the silent "never activated" case means the operator isn't told enforcement is off (violates "surface decisions and work-stoppages, not status"), and the cross-harness failover chain re-running the same brief on a weaker-sandboxed harness with no added consent gate trades verifiable autonomy for unattended risk escalation. This is a direction flag for the operator to weigh, independent of the code-correctness findings in the review above.

bloodf:
Pushed a fix for the dangling-anchor finding in `content/agents-medium/architect.md`: it now points to `content/references-medium/agent-team.md` §Spawning.\n\nPlease re-review the current head `d75f8c32`.

tyson-solara6 (changes requested):
## Automated review — requesting changes

Re-reviewed at head `d75f8c32a4` after the fix commit pushed since the last (2026-07-08) round, plus a dedicated follow-up security pass on the supervision/dispatch/hooks/installer surface.

### Prior finding — status

The Major dangling-anchor finding from the last round is **fixed** and independently re-verified: `content/agents-medium/architect.md` (and its `.claude/` mirror) now reads `content/references-medium/agent-team.md §Spawning for named-agent spawn protocol and composer flows` — that heading exists at line 128 in this PR's own head version of the file. Clean, no dangling reference.

The four previously-flagged non-blocking security items remain open and unchanged (not re-scored here): (a) enforcement hooks return silently with zero output in the common "never activated" case, (b) the cross-harness failover chain reuses the same brief verbatim with no consent gating, (c) narrow non-JSON-aware `cwd` extraction in the sed-based hook, (d) `hooks/enforce-no-abdication.py` doesn't use the shared activation guard.

### New finding — High

**Activation-guard `cwd` exact-match creates a systemic bypass of five enforcement hooks inside the mandated worktree-isolation execution path.** `hooks/lib/activation.py` (`is_active`) and the new `_is_active()` wrapper this PR adds to `hooks/enforce-tier.py`, `enforce-background-spawn.py`, `enforce-orchestrator-singularity.py`, `enforce-planning-artifact-spawn.py`, and `enforce-askuserquestion-default.py` resolve activation by checking only the literal `cwd` for `.agentic/{active,active.session,dormant}` or an exact realpath match in `~/.agentic/activation.list` — no ancestor-directory walk.

The methodology mandates every `engineer`/`qa-engineer`/`release-orchestrator` spawn run inside `.agentic/worktrees/<branch>/`. Since `.agentic/` is gitignored, a fresh worktree has no `.agentic/active` marker of its own and is never added to the allowlist. If a subagent's tool-call `cwd` reflects its actual worktree directory (the documented isolation model), `is_active()` returns `False` for every such call — silently disabling all five guards for that subagent. This is most severe for `enforce-orchestrator-singularity.py`, whose entire purpose is blocking a subagent from spawning another subagent: the guard goes dark in precisely the context it exists to police. It would also silently disable the Tier-3 model-enforcement backstop for security/auth/crypto spawns issued from a worktree.

Flagging with appropriate confidence: I couldn't directly verify Claude Code's exact hook-payload `cwd` semantics for subagent-issued tool calls from the codebase alone — worth confirming empirically before treating this as closed, but the code path as written does not defend against it. Suggested fix: walk up from `cwd` to the nearest ancestor `.agentic/{active,dormant}` marker (bounded at `.git` root or filesystem root), or propagate an activation marker/env var into new worktrees at creation time.

### New finding — Medium

**Path traversal in `agentic-team status`/`collect` run-id resolution.** `_resolve_run_dir()` in `bin/agentic-team` joins the CLI-supplied `run_id` directly into `<workdir>/.agentic/teamrun/<run_id>` with no rejection of `..` or `/`, unlike `_make_run_id`/`_rotation_cursor_next`, which do sanitize the `role` string for the same class of risk (there's even an inline comment there naming the exact threat). A crafted `run_id` can resolve outside the intended directory. Impact is bounded today (same-privilege caller), but it's a real missing validation gate — please mirror the existing sanitization pattern.

### New finding — Low (non-blocking)

**Predictable `.tmp` sibling name in `atomic_write`.** `bin/_lib.py`'s `atomic_write` always writes to `<path>.tmp` with no unique suffix or symlink pre-check. Two concurrent writers (plausible given `_supervise.py`'s per-attempt status polling) can race, and a local same-user TOCTOU could redirect the write via a pre-planted symlink. No privilege boundary crossed, so this is low priority — a `tempfile.mkstemp` in the same directory would close it out.

### Positive controls worth noting

`bin/agentic-team` uses `yaml.safe_load` exclusively despite `team.yml` being fully attacker-shapeable; all subprocess spawns use list-form argv with no `shell=True`; `_make_run_id`/`_rotation_cursor_next` already sanitize path components with an explicit threat comment; the new `copilot` harness's unrestricted-filesystem risk is explicitly gated behind an opt-in env var with a clear warning; and the dormant-notice marker is deliberately stored outside the repo so a repo-local actor can't suppress it. Good adversarial thinking already applied in several places.

Please address the High activation-guard bypass before merge — it undermines the enforcement hooks this PR itself extends, in exactly the isolated-worktree context they're supposed to cover. The Medium path-traversal item should also be fixed; the Low item is optional.

bloodf:
Pushed `2aa3b55d` addressing the 2026-07-09 findings:

- High: `hooks/lib/activation.py is_active()` now walks from cwd up to the outermost `.git` root (inclusive), inheriting the project root's markers/allowlist — closes the worktree-isolation bypass of the five enforce-* guards. Bound at the outermost git root so it never escapes into `~/.agentic`. 6 tests added.
- Medium: `bin/agentic-team _resolve_run_dir()` rejects `..`, path separators, absolute paths and NUL with a `commonpath` containment check (rejected ids hit the existing "unknown run-id" branch). Test added.
- Low: `bin/_lib.py atomic_write` uses `tempfile.mkstemp` + `os.replace`; `mode=None` preserves the destination's existing perms (race-free) or 0o644 for new files. Tests incl. a pre-planted-symlink TOCTOU check.

Full suite 516 passed; activation guard + matrix tests green. Please re-review at `2aa3b55d`.

tyson-solara6 (approved):
## Automated review — approving

Re-reviewed at head `2aa3b55d6b` after the fix commit addressing the High/Medium/Low findings from the last security-focused round. Verified adversarially (tried to break each fix, not just read the commit message):

- **High (activation-guard bypass) — FIXED.** `hooks/lib/activation.py:104-229` now walks from `cwd` up to the outermost `.git`-bearing ancestor and evaluates activation/allowlist state there, correctly walking past a worktree's own `.git` *file* to the main checkout's `.git` *directory*. All 5 `enforce-*.py` hooks (`enforce-tier`, `enforce-background-spawn`, `enforce-orchestrator-singularity`, `enforce-planning-artifact-spawn`, `enforce-askuserquestion-default`) load this same shared module — no stale duplicate logic anywhere. The 6 new tests in `bin/tests/test_activation.py:265-347` directly exercise the worktree-inherits-marker scenario the finding was about, using a hermetic fake-`$HOME`/fake-`.git` harness against the real module (not a mock).
- **Medium (path traversal in `agentic-team` run-id resolution) — FIXED.** `bin/agentic-team:1504-1527` (`_resolve_run_dir`) rejects all 4 named bypass classes (separator, `.`/`..`, absolute path, NUL byte) before any filesystem call, plus a `commonpath` containment check as defense-in-depth. Covered by `bin/tests/test_agentic_team.py:358-374`.
- **Low (predictable `.tmp` sibling in `atomic_write`) — FIXED.** `bin/_lib.py:116-159` now uses `tempfile.mkstemp` + `os.replace`; perms-preservation for existing files and the symlink-TOCTOU close are both backed by real regression tests, not just assertions.

CI is green across all 19 checks at this head.

### Non-blocking follow-up (not introduced by this commit, doesn't reopen the fixed vulnerability)
`hooks/lib/activation.js` and `hooks/lib/activation.sh` still implement the old exact-`cwd`-match logic with no ancestor walk. Confirmed this doesn't reopen the High finding — none of the 5 security-enforcement guards route through the JS/Shell copies, only the Python module. Worth a fast-follow to port the same fix for consistency across the hook trio. (Would normally file a tracking ticket for this, but Jira/mcp-atlassian tooling isn't available in this run — flagging here instead so it isn't lost.)

Nice adversarial security work in this round — the fixes hold up under an independent attempt to break them.

tyson-solara6 (approved):
## Automated review — approving (re-review at new head)

Re-reviewed at head `c5539c7535` after two new commits landed since the last APPROVED (security-focused) review (`2aa3b55d6b`, 2026-07-09 13:43 UTC).

**Delta since last approval:** a merge-forward of `origin/main` (`3d6ce9b07a`, same 54 generic infra/doc files as the sibling PR #422's merge — no logic content) plus the same mechanical `fix(adapter-sync)` mirror-regen commit (`c5539c7535`, 2 files, pure build output/prose).

Given the last approval specifically verified three security fixes (High: activation-guard worktree-bypass in `hooks/lib/activation.py`; Medium: path traversal in `bin/agentic-team`'s `_resolve_run_dir`; Low: predictable tmp-file in `bin/_lib.py`'s `atomic_write`), I had this re-checked adversarially with extra scrutiny on exactly those files:
- The merge commit's 54 changed files do not include `hooks/lib/activation.{py,js,sh}`, any `enforce-*.py` hook, `bin/agentic-team`, or `bin/_lib.py`.
- Diffing the last-approved commit directly against the current head for those specific files returns **empty** — they are byte-identical to what was approved. The three fixes are intact, untouched by the merge or the mirror-regen commit.

All 19 CI checks pass, including CodeQL across all three languages.

**North Star alignment (direction signal, not a verdict change):** an automated alignment pass flagged this **MISALIGNED** — bundling 7 superseded PRs (#413–#417, #422, #423) into one 288-file / ~54,667-line branch is itself a large attention/verifiability tax on whoever has to review or reason about it, even though the end-state (dormant-by-default install, supervised failover) clearly advances the North Star. Worth the operator's attention as a process concern about PR sizing going forward, not a reason to hold this one back given how thoroughly it's already been reviewed.

**Outstanding item — please smoke-test before merging:** the PR's own test plan lists two unchecked manual verification steps: one live dispatch per external harness, and an `/ds activate` → `/ds status` → `/ds deactivate` round-trip in a scratch project. All automated gates (pytest, ruff, mypy, drift/sync checks, CodeQL) pass, and the code-level review is clean, but an automated reviewer can't perform these two live/interactive checks. Recommend running them before merge given this touches activation logic across all 11 harnesses.

tyson-solara6 (changes requested):
**Automated review** (skeptic-style correctness pass + security audit, run in parallel — both independently converged on the same root cause below).

## Verdict: Request changes

### Critical/High — the activation-guard fix this PR ships is incomplete across languages, reopening the exact bug it claims to close
Commit `cd7c158` fixes a real bug: `hooks/lib/activation.py`'s `is_active()` checked only the literal cwd, so subagents running in the mandated `isolation: "worktree"` path (`.agentic/worktrees/<branch>/`, no local `.agentic` marker) silently resolved to dormant, disabling every Python `enforce-*` guard. The fix adds an ancestor-walk (`_iter_roots`/`_git_bound`) that climbs from cwd to the git root checking for `.agentic/{active,dormant}` at each level, nearest-first (verified via `bin/tests/test_activation.py::TestHookActivationAncestorWalk`, 7 passing).

Two problems with this, found independently by both reviewers:

1. **The walk exists only in `activation.py`.** `hooks/lib/activation.sh` (sourced by `session-start-wrap.sh`, `skill-auto-load-check.sh`) and `hooks/lib/activation.js` (required by `stop-context.js`, `session-end-wrap.js`, `pre-tool-use-spawn-emit.js`, `post-tool-use-capture-nudge.js`, `wrap-daemon.js`) — plus `bin/_activation.py::resolve_state` (used by `/ds status`) — still do exact-cwd checks only. Concrete failure: an engineer subagent spawned per the methodology's own worktree-isolation mandate runs `stop-context.js` inside its worktree; `isActive()` finds no local marker, returns `false`, and the hook silently no-ops — dropping `.agentic/events.jsonl` telemetry for that subagent even though the true project root is ACTIVE. `check-adapter-sync`/`check-drift` (both green) don't catch this because they check generated-artifact byte-identity, not behavioral parity across the hand-maintained `.py`/`.js`/`.sh` siblings.

2. **Nearest-root-wins cuts the other way too.** Since a subagent worktree carries its own `.git` file, the Python ancestor walk treats it as a valid candidate root. Any actor with write access to that worktree — which is exactly what an `engineer` subagent has, by design — can `touch .agentic/dormant` inside its own worktree and force `is_active()` to return `false` for every tool call scoped there, disabling `enforce-background-spawn.py`, `enforce-askuserquestion-default.py`, `enforce-no-abdication.py`, `enforce-orchestrator-singularity.py`, and `enforce-tier.py` for itself — even though the real project root has an explicit `.agentic/active` marker. The new test suite covers a tombstone at the project root propagating *down* into a worktree, but never tests one planted *inside* the worktree overriding an active root marker, so this gap shipped untested.

**Fix:** port the ancestor-walk into `activation.sh`, `activation.js`, and `bin/_activation.py::resolve_state` for consistency; and for enforcement-hook consumers specifically, either resolve from the outermost git root only, or ignore `.agentic/dormant` tombstones found strictly inside `.agentic/worktrees/**` (that path is exclusively conductor/subagent-managed scratch space, never an operator-facing project boundary).

### Major — a regression test for this PR's own notice-spam fix never actually runs
`bin/tests/test_activation.py::test_dormant_notice_does_not_spam_on_repeated_marker_failure` is defined nested inside `if __name__ == "__main__": unittest.main()`. Under `pytest` the module import never reaches that block so the method is never collected; under direct `python3 file.py`, `unittest.main()` exits before the nested `def` is reached. It provides zero real coverage for the fix it was written to verify.

### Major — headline "supervised multi-model teams" feature is untested
`_is_live_readonly` (`bin/agentic-team:1550`) implements the new `status.json`-driven liveness states (`running/retrying/failed_over` → live; `done/failed/killed/unkillable` → not live). The only test touching this function exercises solely the legacy PID/exit-file fallback branch — no test writes a `status.json` exhibiting any of the new states.

### Scope note
18 commits / 292 files / 5 largely-independent subsystems (install, hooks/activation, methodology tiers, adapter mirrors, agentic-team dispatch) bundled into one PR. Two real, independently-verified bugs surfaced from spot-checking just two of those subsystems — recommend splitting PRs of this shape going forward for reviewability.

Positive controls confirmed: TUI harness-label allowlist validation (with test coverage), list-form subprocess/array invocation throughout (no shell=True/eval/argv injection), `_resolve_run_dir` path-traversal hardening, rotation-cursor path sanitization, and the `atomic_write` TOCTOU/symlink-race fix (`mkstemp` instead of predictable `.tmp` sibling) are all solid.

tyson-solara6 (changes requested):
Automated review (security-auditor + adversarial code review dispatched in parallel).

**Verdict: Request changes** - mainly on test coverage for the headline reliability feature, not a design or security blocker.

## Security review
No Critical or Major findings. Two Minor hardening suggestions:
- `.github/workflows/adapter-sync.yml` / `shell-syntax.yml` pin `actions/checkout@v7` by mutable tag rather than commit SHA (CWE-829). Low severity here (first-party action, `pull_request` trigger, `contents: read` only, no secrets), but worth pinning to a SHA.
- `bin/agentic-team` `_build_worker_argv` forwards `team.yml` model/role strings straight into subprocess argv without checking for a leading `-`/`--`. `team.yml` is locally operator-authored today, so this isn't currently attacker-controlled, but a malformed entry (e.g. `model: --dangerous-flag`) could be misread as a flag by the target harness CLI. Consider validating the string against an allowed charset before adding to argv.

Notably solid: no `curl|bash` of unpinned remote code, no shell-injection surface (argv lists throughout, no `shell=True`), explicit path-traversal defenses in `_resolve_run_dir`/`_rotation_cursor_next`, and conventional env-var-only credential handling.

## Adversarial / correctness review
- **Major (test gap):** every dispatch test in `bin/tests/test_agentic_team.py` explicitly sets `retries: 0, failover: false` for determinism, so there's no test that actually exercises a failover attempt - the reliable-multi-model-teams feature this PR is built around. Recommend at least one integration test that forces a first-attempt failure (e.g. a fake harness binary exiting 1) and asserts the chain advances, `status.json` records `retrying`/`failed_over`, and the final state is correct. `_build_failover_chain` itself also has no direct unit test.
- **Minor (DRY):** `bin/agentic-team` `_cmd_dispatch` parses `team.yml` up to three times per dispatch (once in `main()`, once in `_cmd_dispatch`, once inside `_resolve_role_model`'s own `_load_team_config` call), with divergent error handling between the sites. Consider threading the already-loaded config through.
- **Minor:** `dispatch.retries` validation only checks the value is an integer, not non-negative; a negative value silently clamps to a single attempt via `max(1, 1 + retries)` with no warning.
- **Minor (cross-language drift):** `hooks/lib/activation.py`'s dormant-notice stderr message has no equivalent in `activation.js` or `activation.sh` - Node/shell-triggered hooks silently no-op with no user-visible dormancy signal, while Python hooks do surface it. Worth tracking as the three ports continue to diverge.
- **Minor:** `bin/_supervise.py` stall detection relies on stdout/stderr file mtime as the liveness heartbeat; a harness that block-buffers stdout under a non-tty pipe could go quiet on disk longer than `stall_seconds` and get falsely killed. Worth a doc note or explicit unbuffered flag per harness.

No pre-existing out-of-scope issues identified - everything reviewed is new code from this PR, so no follow-up tickets filed.

Happy to re-review once the failover-chain test lands.

bloodf:
Review fixes are in `eaa48849` (+ `04d78aba` drift-baseline regen); CI green (19 checks):

- **Ancestor-walk activation fix (Critical/High):** the fix previously existed only in `activation.py`. Ported to `hooks/lib/activation.sh`, `hooks/lib/activation.js`, and `bin/_activation.py` — all four language implementations now walk `_iter_roots`/`_git_bound` from cwd to git root.
- **Worktree-zone tombstone guard:** candidate roots at/below `<outermost-git-root>/.agentic/worktrees/` are subagent scratch space — markers there are ignored and the walk continues; the project root's markers always decide. Prevents a subagent disabling `enforce-*` hooks via its own dormant tombstone.
- Fixed a nested test (`if __name__` inside a test) and added `_is_live_readonly` tests including status-beats-dead-pid and malformed-json fallback.

Local: 141 passed +11 subtests; the full activation matrix passes across sh/js/py. Both reviewer-found bugs (walk missing in sh/js, tombstone semantics) fixed with parametrized tests. Ready for re-review.

On the scope note (18 commits / 292 files / 5 subsystems): acknowledged — the canonical-PR split vs #422 is a separate operator decision and I'll act on it once you pick A or B.

bloodf:
Rebased onto current `main` (conflicts were the tier-format `.methodology-baseline.sha256` and generated `.hermes/SKILL.md` — regenerated all three tier hashes + rebuilt all 11 adapters; drift/adapter-sync green) and landed the failover test coverage in `9f93ac09`:

- **`_build_failover_chain` unit tests (4):** pin the chain shape — retries-only when `failover=false`, single-attempt floor at `retries=0`, other-models-then-terminal-`(claude, None)` when `failover=true`, and no redundant claude terminal for a claude-harness role.
- **e2e chain advancement:** a fake `omp` binary exits 1 on attempt 1 then 0 on attempt 2; the test asserts the chain advances, `status.json` ends `done` at attempt 2 with a `retrying` transition recorded, and the exit file reflects the successful retry. This exercises the reliable-multi-model-teams path that every prior dispatch test skipped via `retries:0/failover:false`.

Full `test_agentic_team.py` green (116 passed). The four Minors (config parsed up to 3x per dispatch; negative-retries clamp; cross-language dormant-notice drift; mtime-heartbeat false-kill) are left as noted — hardening/DRY, not correctness blockers.

tyson-solara6 (changes requested):

Reviewed the incremental diff since our last review round (`eaa48849` → `9f93ac09`), since the full PR diff is ~80K lines across regenerated per-harness mirrors.

**Correctness/security — 1 Major / 1 High (same underlying issue, converging findings):**

1. **Cross-harness failover has no consent gate (High).** `bin/agentic-team:_build_failover_chain` defaults `dispatch.failover=True` and, on exhausted same-harness retries, appends a terminal `("claude", None)` fallback and can route through other harnesses mid-chain — re-running the same brief on a harness with a different (often weaker) sandboxing posture, with no operator confirmation. Worse than the design-note version of this finding suggested: `opencode` dispatches get `--dangerously-skip-permissions` with zero gate (only `copilot` got an opt-in env-var gate, `AGENTIC_TEAM_ALLOW_COPILOT`), and `worker_env = dict(os.environ)` hands the dispatched subprocess (codex/gemini/claude/opencode/copilot) the conductor's full environment — every credential/token included. This is a previously-flagged concern that is still open in live code, not just a design gap. Recommend an explicit `dispatch.failover_confirm` (or equivalent) gate before silently changing harness/model mid-run, applied uniformly (not just to copilot), plus scoping `worker_env` to an explicit allowlist instead of the full environment.

2. **Silent "never activated" case remains open (Medium).** `hooks/enforce-background-spawn.py` exits 0 with zero stdout/stderr when the project was never activated — all enforcement (background-spawn, team-routing, sentinel) silently no-ops. The dormant-notice mechanism added in this PR (`_emit_dormant_notice`) only covers the explicit-deactivation tombstone case (a project that was activated then deactivated); a project that simply never opted in still gets no signal at all. This was one of the two originally-flagged security concerns and is only partially addressed.

**Minor:**
- A code comment cites "the RISK-ACCEPTED note in cross-harness-teams.md" for the copilot gate, but that doc no longer contains any mention of copilot (the doc was trimmed from 9 to 7 harnesses in this PR) — the rationale an operator would go read doesn't exist anymore.
- `abdication_guard_enabled` default flipped `true`→`false` repo-wide. Intentional and documented, but worth flagging explicitly to whoever makes the canonical-PR call, since it's a meaningful behavior change.

**North Star alignment:** MISALIGNED — both previously-flagged direction concerns (no consent gate on cross-harness/model failover; silent-dormant visibility) persist in substance, even though ex-post status logging was added for the failover case. This is a direction flag for the operator, not a verdict override, but it reinforces the code-level findings above.

**Note on the merge-order hold:** #422 and #431 have a genuine 3-way merge conflict (verified via `git merge-tree`, 15 conflicting files) and an active operator hold remains in effect — do not merge either until the canonical-PR decision is made, independent of this review's verdict.

Requesting changes on the failover consent gap (High) and the still-open silent-dormant case (Medium).

bloodf:
Rebased onto current `main` (conflict was the tier-aware `.claude/build.sh` rewrite — took main's version and regenerated all 11 adapters + tier baselines; drift/adapter-sync green) and addressed the High/Medium/Minor findings in `0e631bf3`:

- **High — cross-harness failover consent gate:** `dispatch.failover` now defaults to **false**. Failover re-runs a brief on a *different* harness (different sandbox posture, different provider creds), so it's an explicit `team.yml` opt-in (`dispatch.failover: true`). Same-harness `retries` carry no cross-harness escalation and stay on by default.
- **High — env leak:** dispatched workers no longer inherit `dict(os.environ)`. New `_scoped_worker_env()` forwards an allowlist — shell/locale basics, recognized provider-auth vars (`API_KEY`/token/base-url markers), harness/provider prefixes — and drops everything else (verified: `AWS_SECRET_ACCESS_KEY`/`STRIPE_SECRET`/`MY_DATABASE_PASSWORD` no longer leak). `AGENTIC_TEAM_ENV_PASSTHROUGH` is the escape hatch. Also removed a dead full-environ build in `_cmd_dispatch`.
- **Medium — silent never-activated case:** a project that never opted in now emits a distinct `AGENTIC-ENGINEERING INACTIVE` notice once (pointing at `/ds activate`), instead of the prior silent no-op. Markers are kind-scoped so the INACTIVE and tombstone-DORMANT notices fire once each independently.
- **Minor — stale doc citation:** restored a `Cross-harness escalation (RISK-ACCEPTED)` section to `cross-harness-teams.md` (opencode `--dangerously-skip-permissions`, the copilot `AGENTIC_TEAM_ALLOW_COPILOT` gate, failover-off default, env allowlist); the two code comments citing that note resolve again.

Tests: env-allowlist (secret-drop + passthrough + disable-flag), failover-default-off, inactive-notice-once. `test_agentic_team.py` (120) and all activation-guard suites green.

**Minor — `abdication_guard_enabled` default flip (`true`→`false`):** left as-is and flagging for your explicit call — it's an intentional, documented behavior change, not something I want to silently revert or lock in as part of this review pass. Say the word if you want it reverted here.

Re the merge-order hold vs #422: both are rebased and independently green now; leaving the canonical-PR / keep-vs-close decision to you.

tyson-solara6 (changes requested):

The activation-guard ancestor-walk work (subject of earlier Critical/High rounds) is now solid — all four implementations (`hooks/lib/activation.{py,js,sh}`, `bin/_activation.py`) are parity-tested via `bin/tests/test_activation.py`'s cross-implementation matrix, wired into CI. A parallel security audit also passed this PR with no Critical/High findings, and confirms a real prior credential-over-exposure bug (full env forwarded to dispatched subprocesses) is fixed here via a new allowlist (`_scoped_worker_env`). Three issues still block approval:

**Major — TUI installer mutates state before the final confirm**
`scripts/install-tui.sh:383-393` vs `:454` — `bin/agentic-team configure --non-interactive` runs as soon as `TEAM_DEFAULT`/`TEAM_ASSIGNS` is set, **before** the "Proceed with install? [y/N]" prompt at line 454. If the operator declines, the script prints "cancelled" and exits as if nothing happened, but `team.yml` has already been mutated. Move the team-config call into the confirmed execution path (after line 457), or gate it on the same yes/no.

**Major — new test suites for this PR's own headline features aren't wired into CI**
`bin/tests/test_install_tui.sh` (165 lines, 7 scenarios) and `hooks/tests/test_activation_guard.{py,js,sh}` (~370 lines) are only syntax-linted, never executed by any workflow — confirmed by grepping all 14 workflow files. The PR's own test-plan checklist lists these as manually-verified, which reads as CI coverage but isn't. Note the manual suite also doesn't exercise the state-consistency bug above (its decline-confirm test never combines a real team assignment with a decline).

**Major (security-adjacent) — permission-bypass gating fix is incomplete**
`bin/agentic-team:1095-1125` — a prior round flagged unattended `--dangerously-skip-permissions`/`--yolo` dispatch with no consent gate. This commit adds a gate for copilot (`AGENTIC_TEAM_ALLOW_COPILOT`) and fixes `failover` to default `False`, but `opencode` and `kimi` still unconditionally inject their own bypass flags with no equivalent `AGENTIC_TEAM_ALLOW_OPENCODE`/`_KIMI` gate. Same risk class, inconsistently applied. (Independently flagged by the security-auditor pass as Medium/risk-accepted-but-worth-tracking — both agents agree this is the same underlying gap.)

**Minor**
- `bin/agentic-team:1352-1360` — copilot's consent message still claims the worker "inherits the full parent environment," stale now that `_scoped_worker_env` applies uniformly.
- `scripts/install-tui.sh:345` — `IFS=';' read` only splits into 2 fields; a malformed 3+-segment answer silently glomms the excess onto the second field with no validation.

**Design note (non-blocking, echoing a prior round):** this PR bundles three independently-shippable features (dormant install, supervised multi-model dispatch, TUI installer) into one 292-file/~53K-line branch, and shares commit ancestry with the currently-open #422 and #443. Not a reason to hold this specific round given how many rounds it's already been through, but worth declaring merge order across the three before any of them lands.

tyson-solara6:

tyson-solara6:
Automated review check-in: no new commits since our last review (head SHA unchanged at `0e631bf`, CHANGES_REQUESTED). No re-review performed this run.

tyson-solara6:
Automated review check-in: no new commits since our last review (head SHA unchanged at `0e631bf`, CHANGES_REQUESTED verdict stands). Still showing merge conflicts (mergeable: CONFLICTING) against `main`, and the DCO Signed-off-by check is still failing.

**North Star alignment check (new this pass, direction signal only):** ALIGNED - dormant-by-default install and self-gating hooks directly cut attention/context tax and friction, while generalized multi-harness supervision/failover and `agentic-team status` add autonomous, checkable outcome verification, with no evidence of increased operator babysitting from the described changes.

tyson-solara6:
Automated review check-in: no new commits since our last review (head SHA unchanged at `0e631bf`, CHANGES_REQUESTED verdict stands). `DCO Signed-off-by check` is still failing on this PR, and it still shows merge conflicts (mergeable: CONFLICTING, mergeStateStatus: DIRTY) against `main`. No re-review performed this run.

tyson-solara6:
Automated review check-in: no new commits since our last review (head SHA unchanged, still `0e631bf3`; last verdict: CHANGES_REQUESTED on 2026-07-15). CI is green. Note: this PR is now reporting a merge conflict against `main` (`mergeable_state: dirty`) — will need a rebase. North Star alignment concern noted in a prior comment (supervised multi-model teams / setup ceremony) still stands as a direction flag, not a blocker.

tyson-solara6:
Automated review check-in: no new commits since our last review (head SHA unchanged at `0e631bf`, CHANGES_REQUESTED verdict stands). Still showing merge conflicts (mergeable: CONFLICTING) against `main`, and the DCO Signed-off-by check is still failing. North Star alignment concern noted in a prior comment (silent never-activated case + unconsented cross-harness failover) still stands as a direction flag, not a blocker - MISALIGNED. Jira auto-filing unavailable in this environment - no new follow-up tickets filed this run.

tyson-solara6:
Automated review check-in: no new commits since our last review (head SHA unchanged at `0e631bf`, CHANGES_REQUESTED verdict stands). Still showing merge conflicts (mergeable: CONFLICTING, mergeStateStatus: DIRTY) against `main`, and the `DCO Signed-off-by check` is still failing. North Star alignment concern noted in a prior comment (silent never-activated case + unconsented cross-harness failover) still stands as a direction flag - MISALIGNED, not a merge blocker. Jira auto-filing still unavailable in this environment - no new follow-up tickets filed this run.

tyson-solara6:
Automated review check-in: no new commits since our last review (head SHA unchanged, still `0e631bf3`). CI has a real failure - DCO Signed-off-by check is failing on this PR specifically - and it is also reporting a merge conflict against `main` (mergeable: CONFLICTING, mergeStateStatus: DIRTY). North Star alignment: MISALIGNED (unchanged from prior pass) - flagged as a direction signal for the operator, not a merge blocker. Verdict remains changes-requested pending the DCO fix, rebase, and prior review feedback. No Jira follow-ups filed this run (mcp-atlassian tools unavailable).

tyson-solara6:
Automated review check-in: no new commits since our last review (head SHA unchanged, still `0e631bf`). Standing verdict remains CHANGES REQUESTED. Also note this PR now has merge conflicts against `main` (mergeStateStatus: DIRTY) and the DCO Signed-off-by check is failing.

tyson-solara6:
Automated review check-in: no new commits since our last review (head `0e631bf`, changes requested 2026-07-15). Outstanding blockers unchanged: `DCO Signed-off-by check` still failing, branch still `CONFLICTING` with `main`, plus the previously requested changes. No new formal review posted since nothing has changed.

North Star alignment (re-assessed this run): **ALIGNED** - dormant-by-default install plus self-gating, sub-10ms hooks reduces resident footprint and ceremony without weakening verifiability (Low Friction pillar). Note: an earlier automated pass (2026-07-08) flagged this MISALIGNED; on re-assessment against the same vision doc the direction reads as aligned. Flagging the discrepancy for the operator's awareness rather than asserting either judgment as final.

tyson-solara6:
Automated review check-in: no new commits since our last review (head `0e631bf`, CHANGES_REQUESTED verdict from 2026-07-15 stands). Outstanding blockers unchanged: `DCO Signed-off-by check` failing, branch `CONFLICTING` with `main`. North Star alignment: ALIGNED (per prior re-assessment). No new formal review posted since nothing has changed.

tyson-solara6:
Automated review check-in: no new commits since our last review (head unchanged at `0e631bf`, last verdict CHANGES_REQUESTED 2026-07-15). Outstanding blockers unchanged: `DCO Signed-off-by check` still failing, branch still `CONFLICTING` with `main`. North Star alignment: on the most recent re-assessment this reads ALIGNED (dormant-by-default install plus self-gating reduces resident footprint/ceremony without weakening verifiability); an earlier pass had flagged MISALIGNED on the same vision doc - noting the discrepancy for the operator rather than asserting either as final. No new formal review posted since nothing has changed. Jira auto-filing unavailable in this environment - no new follow-up tickets filed this run.

tyson-solara6:
Automated review check-in: no new commits since our last review (head unchanged at `0e631bf`, last verdict CHANGES_REQUESTED 2026-07-15). PR is `mergeable: CONFLICTING` against `main` (mergeStateStatus: DIRTY) and the DCO check is failing. No new formal review posted since nothing in the diff has changed; prior requested changes still stand.

tyson-solara6:
Automated review check-in: no new commits since our last review (head unchanged at `0e631bf`, last verdict CHANGES_REQUESTED 2026-07-15). PR is `mergeable: CONFLICTING` against `main` (mergeStateStatus: DIRTY) - needs a rebase before it can be reconsidered. No new formal review posted since nothing in the diff has changed; prior requested changes still stand.

tyson-solara6:
Automated review check-in: no new commits since our last review (head `0e631bf` unchanged, our last verdict was CHANGES_REQUESTED). `DCO Signed-off-by check` is still failing and the PR is CONFLICTING with the base branch - both need to be resolved before re-review.

tyson-solara6:
Automated review check-in: no new commits since our last review (head unchanged at `0e631bf`, verdict remains CHANGES_REQUESTED). North Star alignment concern flagged 2026-07-08 (adds capability/attention tax without proportional autonomy/verifiability gain) still stands unchanged - direction signal, not a verdict change. CI checks passing, but the PR is CONFLICTING with main.

tyson-solara6:
Automated review check-in: no new commits since our last review (head unchanged at `0e631bf`, last verdict CHANGES_REQUESTED 2026-07-15). CI checks all passing. PR remains `mergeable: CONFLICTING` against `main` - needs a rebase before it can be reconsidered or move to the ready-to-merge queue. No new formal review posted since nothing in the diff has changed.

tyson-solara6:
Automated review check-in: no new commits since our last review (head unchanged at `0e631bf`). Our last verdict (CHANGES_REQUESTED) still stands. CI checks all passing, but the branch has a merge conflict with `main` (mergeStateStatus: DIRTY/CONFLICTING) - not eligible for the ready-to-merge queue until rebased and outstanding changes are addressed.

tyson-solara6:
Automated review check-in: no new commits since our last review (head `0e631bf`, unchanged). Prior verdict (CHANGES_REQUESTED) stands. PR remains CONFLICTING with base - not mergeable regardless of review verdict.

tyson-solara6:
Automated review check-in: no new commits since our last review (head `0e631bf`, unchanged). Prior verdict (CHANGES_REQUESTED) stands. PR remains CONFLICTING with base - not mergeable regardless of review verdict.

tyson-solara6:
Automated review check-in: no new commits since our last review (head `0e631bf`, unchanged). Prior verdict stands (CHANGES REQUESTED). PR remains CONFLICTING with base - not eligible for the ready-to-merge queue until rebased and the outstanding review comments are addressed.

tyson-solara6:
Automated review check-in: no new commits since our last review (head `0e631bf`, unchanged). Prior verdict stands (CHANGES REQUESTED). PR remains CONFLICTING with base - not eligible for the ready-to-merge queue.

tyson-solara6:
Automated review check-in: no new commits since our last review (head `0e631bf`, unchanged). Prior verdict stands (CHANGES REQUESTED). PR remains CONFLICTING with base - not eligible for the ready-to-merge queue.

tyson-solara6:
Automated review check-in: no new commits since our last review (head `0e631bf`, unchanged). Last formal verdict stands: **request changes** (2026-07-15 - activation-guard ancestor-walk concerns). Branch is now **CONFLICTING** with base (mergeStateStatus DIRTY); the previously-noted DCO Signed-off-by check failure should also be reconfirmed once rebased.

tyson-solara6:
Automated review check-in: no new commits since our last review (head `0e631bf`, CHANGES_REQUESTED stands). PR remains CONFLICTING with base — please rebase when ready for re-review.

PR #390, opened 2026-07-02
fix(abdication-guard): enable by default across all adapters and templates (DS-47)

tyson-solara6:
## Summary

Fixes DS-47: *Abdication Guard Ships Disabled by Default — Architect Silent-Stop Not Auto-Fixed*

The abdication guard Stop hook was merged in PR #305 (hook implementation) and made fail-open safe in PR #339 (counter write failure = allow stop), but shipped with `abdication_guard_enabled: false` — requiring every project to explicitly opt in. Teams that don't discover the toggle get no protection from conductor abdication.

## Changes

- **10 template `config.json` files** — flip `abdication_guard_enabled: false → true` (canonical `content/templates` + all per-adapter: `.claude`, `.codex`, `.cursor`, `.gemini`, `.kimi`, `.omp`, `.openclaw`, `.opencode`, `.pi`)
- **All init-project adapters** — update seed block `false → true` (claude, codex, cursor, gemini, github, hermes, openclaw, opencode)
- **All prose** — update `'default `false` (opt-in)'` → `'default `true` (opt-out)'` in:
  - `content/commands/init-project.md`
  - `content/sections/02-delegation.md`
  - `content/references/conventions-detail.md` + `risk-config-and-tiers.md` (+ all adapter mirrors)
  - `docs/safe-configuration.md`, `docs/configuration-reference.md`, `docs/components.md`, `README.md`
- **Hook docstring** — update `hooks/enforce-no-abdication.py` to reflect new default

## Safety

The fail-open behavior added in PR #339 means a corrupt/missing config.json exits 0 (allows the stop) — so enabling by default cannot brick a session. Projects wanting opt-out can set `abdication_guard_enabled: false` in `.agentic/config.json` or use `AE_ABDICATION_GUARD_DISABLE=1` per-session.

PR #432, opened 2026-07-08
fix(abdication-guard): enable by default across all adapters and templates (DS-47)

tyson-solara6:
Supersedes #390 (same change, rebased onto current main; the original branch could not be force-pushed from this session). Original description: see #390.

Changes vs #390 beyond the rebase:
- All commits carry DCO sign-off
- Adapters rebuilt via `scripts/build-all.sh`; methodology baseline regenerated (content/sections/02-delegation.md is in the diff)
- Restores unrelated content the original DS-47 commit had silently dropped from a stale snapshot: the `goal-condition-evaluator` agent row/count, `feedback-triage` command count, the `docs/codex-permissions.md` README cross-ref, and `opencode, copilot` in the cross-harness dispatch list (verified line-by-line against main; caught by Skeptic-reviewed diff comparison)
- Semantic merge verified against today's DS-55 defaults docs and #415/#416

Safety notes from the original PR still apply: the hook is fail-open (corrupt/missing config allows stop); opt-out via `abdication_guard_enabled: false` or `AE_ABDICATION_GUARD_DISABLE=1`. Note: a config.json that exists but omits the key still resolves off (hook uses `is not True`); templates ship the key explicitly so deployment default is on.


tyson-solara6:
Automated check-in (self-authored, no formal verdict — self-review is blocked anyway).

Found one real discrepancy between the doc/comment changes and the actual runtime logic in `hooks/enforce-no-abdication.py`:

The PR updates the comments at lines ~41-42 and ~401-402 to say "Guard is on by default (`abdication_guard_enabled` defaults to `true`)", and flips the seeded template default in every `.agentic/config.json` template to `true`. But the actual enforcement check is unchanged:

```python
if config.get("abdication_guard_enabled") is not True:
    sys.exit(0)
except Exception:
    sys.exit(0)
```

This still fails open (guard OFF) whenever:
- `.agentic/config.json` is missing entirely, or
- the file exists but doesn't have the `abdication_guard_enabled` key yet (i.e. every project scaffolded before this PR).

So "enable by default" only actually takes effect for brand-new projects that get the key seeded via `/init-project` (or anyone who explicitly adds `"abdication_guard_enabled": true`). Every existing project's config predates this key, so the guard stays off for them despite what the updated comments/docs now claim.

If the intent is genuinely "on by default everywhere," the check should probably become something like `config.get("abdication_guard_enabled", True) is not True` (defaulting the *absent-key* case to enabled, while still failing open on read/parse errors for safety). If the intent is narrower — "only new/`init-project`-scaffolded projects get it on by default, existing projects need to add the key" — then the comments and `docs/safe-configuration.md` / `docs/components.md` wording ("on by default", "Guard is on by default") should say so explicitly rather than reading as universal.

Not fixing this myself since it's a design decision (which semantics you actually want), not a typo. Flagging before merge.

PR #442, opened 2026-07-09
feat(identity): per-profile (per-tenant) developer identity scope

bloodf:
## Summary

Add a third identity scope, `profile`, co-located with the active harness profile's config dir (`<config-dir>/identity.yml`). Per-tenant installs now attribute telemetry per profile instead of collapsing into one global handle.

- `bin/agentic-identity`: 4-tier to 6-tier resolution (project > profile > global, confirmed-then-provisional); `--scope profile` on init/auto/confirm/show; `--profile-dir` override; active config dir detected via `AGENTIC_CONFIG_DIR` / `CLAUDE_CONFIG_DIR` / `CODEX_HOME` (values outside `$HOME` are rejected). Back-compatible: with no env set, behavior is byte-identical to the prior 4-tier.
- `hooks/stop-context.js`: mirror 6-tier `getIdentity()` with the same env detection.
- `scripts/lib/identity.sh`: honor `AE_IDENTITY_SCOPE` (default global) so the helper can write profile-scoped identity via `--profile-dir "$AE_CONFIG_DIR"`.
- `scripts/install-profiles.sh`: set `AE_IDENTITY_SCOPE=profile` for per-profile installs, so each tenant config dir gets its own `identity.yml`.
- `bin/tests/test_agentic_identity.py`: 8 new tests (6-tier ordering, env precedence, outside-$HOME rejection, `--profile-dir` override, back-compat). 18/18 pass; `test_install_profiles.sh` + `test_install_profile_config_dir.sh` pass.

## North Star alignment

- **Guard operator attention.** No prompt at first use; identity auto-derives from `gh api user`. Attention stays on the work, not on a meta-dialog about who you are.
- **Verifiable outcomes autonomously.** All 18 identity tests pass deterministically under fake `$HOME`. L1-L6 + J1-J6 + K1-K4 cover tier ordering, env precedence, symlink escape rejection, profile-dir mkdir guards, and pending-buffer flush filtering by config dir. Each test pins a specific assertion that would fail on the corresponding regression.
- **Universality.** Same identity resolution across Claude Code, Codex, and any harness that sets `AGENTIC_CONFIG_DIR` / `CLAUDE_CONFIG_DIR` / `CODEX_HOME`. Per-profile scope is opt-in (`--scope profile`); the existing global/project scopes are unchanged for callers that don't set the new env var.

Trade-off: more resolution tiers (4 to 6) in exchange for per-tenant telemetry isolation. Behavior is byte-identical when no `AE_IDENTITY_SCOPE` / `--scope profile` is set, so existing callers see no change.

### DIFF
--- a/.github/workflows/wrap-lock-tests.yml
+++ b/.github/workflows/wrap-lock-tests.yml
@@ -19,3 +19,5 @@ jobs:
         run: node hooks/tests/test-wrap-acquire-lock.js
       - name: wrap-lock release test
         run: node hooks/tests/test-wrap-release-lock.js
+      - name: stop-context identity resolution test
+        run: node hooks/tests/test-stop-context-identity.js
--- a/bin/agentic-identity
+++ b/bin/agentic-identity
@@ -19,7 +19,8 @@ Public API:
   Helpers: _read_identity(path=None), _write_identity(lines, target_path=None),
            _project_identity_path(cwd=None), _profile_identity_path(profile_dir=None),
            _profile_config_dir(), _resolve_effective_identity(cwd),
-           flushPendingBuffer(confirmed_dev_id, repo_root_filter=None).
+           flushPendingBuffer(confirmed_dev_id, repo_root_filter=None,
+                              profile_dir_filter=None).
 
 Data model:
   Global:  ~/.agentic/identity.yml
@@ -64,6 +65,10 @@ Failure modes: exits 1 on validation error, 2 on file-exists-without-force.
                does not match the filter are skipped (left in the buffer, not
                unlinked); records with no/empty repo_root are also skipped under
                a filter (conservative).
+               flushPendingBuffer with profile_dir_filter: config_dir-tagged
+               records flush only under a matching profile_dir_filter; a None
+               filter excludes all tagged records (they await their own
+               profile's confirm).
 
 Performance: Standard. File I/O only; subprocess for gh CLI in auto subcommand.
 """
@@ -264,7 +269,11 @@ def _write_identity(lines: list[str], target_path: Path | None = None) -> int:
     return 0
 
 
-def flushPendingBuffer(confirmed_dev_id: str, repo_root_filter: str | None = None) -> int:
+def flushPendingBuffer(
+    confirmed_dev_id: str,
+    repo_root_filter: str | None = None,
+    profile_dir_filter: str | None = None,
+) -> int:
     """Flush ~/.agentic/session-log/.pending/*.json to global + per-project logs.
 
     Race-safe via fcntl.flock exclusive lock on .flush.lock for the whole loop.
@@ -272,12 +281,21 @@ def flushPendingBuffer(confirmed_dev_id: str, repo_root_filter: str | None = Non
     Returns number of records flushed (0 on no pending or timeout).
 
     When repo_root_filter is None (default): current behavior exactly - all
-    pending records are attributed to confirmed_dev_id.
+    untagged pending records are attributed to confirmed_dev_id.
 
     When repo_root_filter is set (e.g. project-scope init/confirm): only records
     whose repo_root exactly matches repo_root_filter are flushed. Records whose
     repo_root does not match, or whose repo_root is absent/empty, are left in the
     buffer (not unlinked, not attributed) for later global flush.
+
+    profile_dir_filter partitions the buffer per profile config dir
+    (cross-tenant leakage guard; records are tagged with config_dir by
+    writePendingBuffer in hooks/stop-context.js when a profile is active):
+      - profile_dir_filter set (profile-scope init/confirm): only records whose
+        config_dir field equals it are flushed; all others stay in the buffer.
+      - profile_dir_filter None (global/project scope): records carrying a
+        config_dir tag are EXCLUDED (they await their own profile's confirm);
+        untagged records behave exactly as before, incl. repo_root_filter logic.
     """
     # Ensure lock file and parent dirs exist.
     FLUSH_LOCK_PATH.parent.mkdir(parents=True, exist_ok=True)
@@ -324,6 +342,15 @@ def flushPendingBuffer(confirmed_dev_id: str, repo_root_filter: str | None = Non
                     # Skip malformed records silently.
                     continue
 
+                # profile partition: tagged records flush only under their own
+                # profile's filter; a None filter (global/project) excludes them.
+                record_config_dir = record.get("config_dir", "")
+                if profile_dir_filter is not None:
+                    if record_config_dir != profile_dir_filter:
+                        continue
+                elif record_config_dir:
+                    continue
+
                 # repo_root_filter: skip records that don't match (leave in buffer).
                 if repo_root_filter is not None:
                     record_repo_root = record.get("repo_root", "")
@@ -344,8 +371,9 @@ def flushPendingBuffer(confirmed_dev_id: str, repo_root_filter: str | None = Non
 
                 # Build attributed line in canonical session-total shape to match
                 # writeSessionLog / writeSessionLogGlobal in hooks/stop-context.js.
-                # Fields NOT in canonical shape (schema_version, repo_root) are used
-                # only for routing/bookkeeping above and must NOT appear in the line.
+                # Fields NOT in canonical shape (schema_version, repo_root,
+                # config_dir) are used only for routing/bookkeeping above and
+                # must NOT appear in the line.
                 attributed = {
                     "ts": record.get("ts"),
                     "phase": "session_end",
@@ -486,10 +514,10 @@ def cmd_init(args: argparse.Namespace) -> int:
 
         print(f"Identity set (profile scope): developer_id={handle}, created_at={created_at}")
 
-        # Profile scope is persona-bound (not repo-bound): flush without a
-        # repo_root_filter, mirroring the global path.
+        # Profile scope flushes only records tagged with this profile's own
+        # config dir (cross-tenant leakage guard); no repo_root_filter.
         if is_overwriting_provisional or is_fresh:
-            flushPendingBuffer(handle)
+            flushPendingBuffer(handle, profile_dir_filter=str(target_path.parent))
 
         return 0
 
@@ -739,8 +767,9 @@ def cmd_confirm(args: argparse.Namespace) -> int:
         if rc != 0:
             return rc
 
-        # Profile scope is persona-bound (not repo-bound): flush without a filter.
-        flushPendingBuffer(dev_id)
+        # Profile scope flushes only records tagged with this profile's own
+        # config dir (cross-tenant leakage guard); no repo_root_filter.
+        flushPendingBuffer(dev_id, profile_dir_filter=str(target_path.parent))
         print(f"Identity confirmed (profile scope): developer_id={dev_id}")
         return 0
 
--- a/bin/agentic-identity
+++ b/bin/agentic-identity
@@ -455,6 +455,32 @@ def flushPendingBuffer(
     return flushed
 
 
+def _profile_target_or_err(args: argparse.Namespace) -> Path | None:
+    """Resolve the profile-scope identity.yml target; print an error and return None on failure.
+
+    Distinguishes a provided-but-rejected --profile-dir (outside $HOME) from an
+    omitted flag with no qualifying env var. Shared by cmd_init/cmd_auto/cmd_confirm.
+    """
+    profile_dir = getattr(args, "profile_dir", None)
+    target = _profile_identity_path(profile_dir)
+    if target is not None:
+        return target
+    if profile_dir is not None:
+        print(
+            f"agentic-identity: --profile-dir '{profile_dir}' rejected: "
+            "must resolve under $HOME.",
+            file=sys.stderr,
+        )
+    else:
+        print(
+            "agentic-identity: cannot determine profile config dir. Set one of "
+            + ", ".join(PROFILE_CONFIG_DIR_ENV)
+            + " (under $HOME), or pass --profile-dir <dir>.",
+            file=sys.stderr,
+        )
+    return None
+
+
 def _require_repo_root(cwd: Path) -> int:
     """Return 1 (and print error) if cwd has neither .agentic/ nor .git/; else 0."""
     if not (cwd / ".agentic").exists() and not (cwd / ".git").exists():
@@ -480,14 +506,8 @@ def cmd_init(args: argparse.Namespace) -> int:
 
     scope = getattr(args, "scope", "global")
     if scope == "profile":
-        target_path = _profile_identity_path(getattr(args, "profile_dir", None))
+        target_path = _profile_target_or_err(args)
         if target_path is None:
-            print(
-                "agentic-identity: cannot determine profile config dir. Set one of "
-                + ", ".join(PROFILE_CONFIG_DIR_ENV)
-                + " (under $HOME), or pass --profile-dir <dir>.",
-                file=sys.stderr,
-            )
             return 1
         existing = _read_identity(target_path) if target_path.is_file() else None
         is_overwriting_provisional = existing is not None and existing.get("provisional", False)
@@ -627,14 +647,8 @@ def cmd_auto(args: argparse.Namespace) -> int:
 
     scope = getattr(args, "scope", "global")
     if scope == "profile":
-        target_path = _profile_identity_path(getattr(args, "profile_dir", None))
+        target_path = _profile_target_or_err(args)
         if target_path is None:
-            print(
-                "agentic-identity: cannot determine profile config dir. Set one of "
-                + ", ".join(PROFILE_CONFIG_DIR_ENV)
-                + " (under $HOME), or pass --profile-dir <dir>.",
-                file=sys.stderr,
-            )
             return 1
         existing = _read_identity(target_path)
         if existing is not None:
@@ -737,14 +751,8 @@ def cmd_confirm(args: argparse.Namespace) -> int:
     """Confirm provisional identity: strip provisional/derived_from, flush pending buffer."""
     scope = getattr(args, "scope", "global")
     if scope == "profile":
-        target_path = _profile_identity_path(getattr(args, "profile_dir", None))
+        target_path = _profile_target_or_err(args)
         if target_path is None:
-            print(
-                "agentic-identity: cannot determine profile config dir. Set one of "
-                + ", ".join(PROFILE_CONFIG_DIR_ENV)
-                + " (under $HOME), or pass --profile-dir <dir>.",
-                file=sys.stderr,
-            )
             return 1
         identity = _read_identity(target_path)
         if identity is None:
@@ -852,8 +860,16 @@ def cmd_show(args: argparse.Namespace) -> int:
         return 0
 
     if scope == "profile":
-        target_path = _profile_identity_path(getattr(args, "profile_dir", None))
+        profile_dir = getattr(args, "profile_dir", None)
+        target_path = _profile_identity_path(profile_dir)
         if target_path is None:
+            if profile_dir is not None:
+                print(
+                    f"agentic-identity: --profile-dir '{profile_dir}' rejected: "
+                    "must resolve under $HOME.",
+                    file=sys.stderr,
+                )
+                return 1
             print("No profile scope (no config-dir env set and no --profile-dir).")
             return 0
         identity = _read_identity(target_path)
--- a/bin/agentic-identity
+++ b/bin/agentic-identity
@@ -1,32 +1,37 @@
 #!/usr/bin/env python3
 """
 Purpose: CLI for managing per-developer identity files used by the Stop hook
-         to write per-developer session telemetry. Supports global identity
-         (~/.agentic/identity.yml) and project-local identity
-         (<repo>/.agentic/identity.yml), with a 4-tier confirmation-first
-         resolution: project-confirmed > global-confirmed > project-provisional
-         > global-provisional > none.
+         to write per-developer session telemetry. Supports three scopes -
+         global (~/.agentic/identity.yml), profile (<config-dir>/identity.yml,
+         co-located with the active harness profile's agentic-engineering.json),
+         and project (<repo>/.agentic/identity.yml) - with a 6-tier
+         confirmation-first resolution: project-confirmed > profile-confirmed >
+         global-confirmed > project-provisional > profile-provisional >
+         global-provisional > none.
 
 Public API:
-  agentic-identity init <handle> [--display-name "<name>"] [--force] [--scope {global,project}]
-  agentic-identity show [--scope {global,project,effective}]
-  agentic-identity auto [--force] [--scope {global,project}]
-  agentic-identity confirm [--scope {global,project}]
+  agentic-identity init <handle> [--display-name "<name>"] [--force] [--scope {global,profile,project}] [--profile-dir <dir>]
+  agentic-identity show [--scope {global,profile,project,effective}] [--profile-dir <dir>]
+  agentic-identity auto [--force] [--scope {global,profile,project}] [--profile-dir <dir>]
+  agentic-identity confirm [--scope {global,profile,project}] [--profile-dir <dir>]
   agentic-identity --help
 
   Helpers: _read_identity(path=None), _write_identity(lines, target_path=None),
-           _project_identity_path(cwd=None), _resolve_effective_identity(cwd),
+           _project_identity_path(cwd=None), _profile_identity_path(profile_dir=None),
+           _profile_config_dir(), _resolve_effective_identity(cwd),
            flushPendingBuffer(confirmed_dev_id, repo_root_filter=None).
 
 Data model:
   Global:  ~/.agentic/identity.yml
+  Profile: <config-dir>/identity.yml  (config dir = AGENTIC_CONFIG_DIR, else
+           CLAUDE_CONFIG_DIR, else CODEX_HOME; only values under $HOME qualify)
   Project: <repo>/.agentic/identity.yml  (gitignored via .agentic/* umbrella)
   Fields: developer_id, display_name (optional), created_at,
           provisional (optional, absent=confirmed), derived_from (optional).
   ~/.agentic/session-log/.pending/<uuid>.json: pre-attribution session records;
     flushed to global + per-project logs on confirm/init.
-  Internal keys _scope ("project"/"global") and _confirmed (bool) are tagged
-    on resolved-identity dicts; never written to any file or log.
+  Internal keys _scope ("project"/"profile"/"global") and _confirmed (bool)
+    are tagged on resolved-identity dicts; never written to any file or log.
 
 Upstream deps: Python 3 stdlib only (argparse, glob, json, os, re, signal,
                subprocess, sys, datetime, pathlib, time).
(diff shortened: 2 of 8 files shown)


tyson-solara6 (changes requested):
Automated review (skeptic + security-auditor, adversarial pass on the diff).

**Verdict: Request changes.** Two independent reviewers (correctness skeptic + security-auditor) independently found the same core issue, which is a strong signal it's real.

### Blocking (Major)
1. **Symlink bypass in `hooks/stop-context.js`'s `$HOME` containment check** (`_profileIdentityPath()`). `path.resolve()` in Node is purely lexical and does not follow symlinks, while the Python sibling (`bin/agentic-identity`'s `_is_under_home`) uses `Path.resolve()`, which *does* follow symlinks via the OS. A symlink under `$HOME` pointing outside `$HOME` (e.g. `~/.claude-alt -> /tmp/evil`) is correctly rejected by the Python CLI but silently accepted by the JS hook - the opposite of the PR's explicit claim that "values outside `$HOME` are rejected," and the two "mirrored" implementations disagree on a security-relevant check. The project already has a tested convention for exactly this threat (`.claude/install.sh` refuses symlinked `--config-dir`, covered by `test_install_profile_config_dir.sh`); the new env-derived path in `stop-context.js` doesn't meet that bar. Fix: resolve symlinks (`fs.realpathSync`, with an ENOENT fallback) before the containment check.

2. **Cross-tenant pending-buffer leakage defeats the PR's core purpose.** `PENDING_DIR`/`GLOBAL_SESSION_LOG_DIR`/`FLUSH_LOCK_PATH` in `bin/agentic-identity` remain hardcoded to a single machine-wide `~/.agentic/session-log/...` location - not partitioned per profile. `cmd_init`/`cmd_confirm`'s profile branches call `flushPendingBuffer()` with no `repo_root_filter` or profile discriminator, and `hooks/stop-context.js`'s pending records carry no tag for which profile/config-dir was active. Net effect: on a machine with two tenants (the PR's stated use case), if tenant A buffers sessions unconfirmed and tenant B later confirms their own profile identity, tenant A's buffered records get flushed and attributed to tenant B. This is normal multi-tenant usage, not an attack - and no test exercises `flushPendingBuffer` under profile scope.

3. **`scripts/lib/identity.sh` Branch 3 breaks profile-scope idempotency detection.** The "detect existing identity" branch hand-builds `--scope "$show_scope"` and never appends the `--profile-dir` flag that the init/confirm branches correctly pass. Since `AE_CONFIG_DIR` is never exported and no profile env var is set by `install-profiles.sh`, every re-run of a profile install will report "no identity" even right after `init` wrote one - not data-corrupting, but a reproducible regression in the exact feature this PR adds.

4. **Zero test coverage for the JS half of the "mirrored" 6-tier implementation.** All 8 new tests are Python-only (`bin/tests/test_agentic_identity.py`); no `hooks/tests/test-stop-context-*.js` test touches `getIdentity`/profile/env-detection - which is exactly where finding #1's divergence lives.

### Minor (non-blocking, worth a follow-up)
- TOCTOU window between the `_is_under_home` containment check and actual path use in `bin/agentic-identity` (low real-world impact - requires existing write access under the victim's own `$HOME`).
- Misleading error message: "cannot determine profile config dir..." fires identically whether `--profile-dir` was omitted or was passed-but-rejected for resolving outside `$HOME`.
- `test_env_detection_precedence` never verifies `CODEX_HOME` resolves correctly in isolation (only in combination with the others).

### Verified sound
6-tier precedence logic itself is correct and agrees between the two implementations; back-compat (no env var set -> byte-identical to prior 4-tier behavior) holds; literal `../` traversal is handled correctly in both languages; no command-injection vectors in `identity.sh`/`install-profiles.sh`.

Note: spotted one pre-existing, out-of-scope issue in code this PR touches - `_write_identity`'s `dest.parent.mkdir(mode=0o700, ...)` only applies the mode on creation, so a pre-existing (looser-permissioned) harness config dir keeps its existing mode. This isn't new in this PR (same pattern used by global/project scope already) and isn't this PR's job to fix. Could not auto-file a follow-up ticket - the Jira MCP tools were unavailable in this run - flagging here so it doesn't get lost.

bloodf:
## Review fixes pushed (`0bc05b0c..003536df`)

All findings from the automated review addressed. Finding → commit map:

| Finding | Fix | Commit |
| --- | --- | --- |
| **Major 1** — symlink bypass in `stop-context.js` `_profileIdentityPath()` | Containment now realpath-based on both sides (`fs.realpathSync` on candidate and on `$HOME`; ENOENT skips to the next env candidate) — JS and Python now agree. TOCTOU window documented per the minor note. | `2dbb8357` |
| **Major 2** — cross-tenant pending-buffer leakage | Pending records now tagged with the active profile `config_dir` (additive optional field, schema v1); `flushPendingBuffer` gained `profile_dir_filter` — profile confirms flush only own-tagged records, global/project flushes exclude tagged ones, untagged legacy records behave exactly as before. Canonical jsonl lines strip the tag. K-tests cover own-tag flush / global-skip / legacy paths. | `067cfdb4` |
| **Major 3** — `identity.sh` Branch 3 missing `--profile-dir` | Detect-existing `show` call now pins `--profile-dir "$AE_CONFIG_DIR"` under profile scope, mirroring the init branches; profile re-runs detect the existing identity (new Test 8 in `test_install_profile_config_dir.sh`). | `afd72556` |
| **Major 4** — zero JS coverage for the 6-tier mirror | New `hooks/tests/test-stop-context-identity.js` (22 asserts): J1–J8 mirrors, `CODEX_HOME`-alone precedence, symlink-escape rejection (fails against the pre-fix hook — genuine regression proof), and the `config_dir` tag. Wired into `wrap-lock-tests.yml`. | `6868e377` |
| Minor — misleading "cannot determine profile config dir" | Consolidated into `_profile_target_or_err`; a passed-but-rejected `--profile-dir` now gets a distinct "must resolve under $HOME" message (also in `show`). | `835921ae` |
| Minor — `CODEX_HOME` never tested in isolation | Third sub-check added to `test_env_detection_precedence`. | `835921ae` |
| Minor — TOCTOU window | Documented at the containment checks in both languages (accepted: attacker needs write access inside the victim's own `$HOME`). | `2dbb8357` |

Adversarial re-review (round 2) caught one consequence of the symlink fix itself: the JS writer realpaths the `config_dir` tag but the Python flush filter didn't, silently orphaning records for symlinked-`~/.config` (Stow/chezmoi) setups. Fixed by realpathing both sides of the compare, with a round-trip regression test (K4) proven to fail on the pre-fix code. | `003536df`

Verified hermetically (`env -u AGENTIC_CONFIG_DIR -u CLAUDE_CONFIG_DIR -u CODEX_HOME`): pytest 22 passed + standalone mode green; both node hook suites green (22 + 25 asserts); `test_install_profiles.sh` + `test_install_profile_config_dir.sh` green; J8 back-compat (no env vars → prior 4-tier behavior) unchanged.

tyson-solara6 (changes requested):
Automated review (security-auditor + qa-engineer). Adversarial skeptic review was attempted but the agent declined to proceed on this ad-hoc external PR (missing formal architect-plan/Brief inputs it requires) - the two findings below stand on their own merit independent of that.

**Verdict: Request changes.**

### Major (test coverage)
1. **`cmd_auto --scope profile` and `cmd_confirm --scope profile` have zero test coverage at any level.** The PR body advertises `--scope profile` on `init`/`auto`/`confirm`/`show`, but every new test calls internal helpers (`_resolve_effective_identity`, `_profile_identity_path`, `flushPendingBuffer`) directly - never the `cmd_auto`/`cmd_confirm` CLI functions with `scope="profile"`. Neither the "already-confirmed, no `--force`" rejection path in `cmd_confirm`, nor the provisional-write path in `cmd_auto`, nor the `flushPendingBuffer(dev_id, profile_dir_filter=...)` call inside `cmd_confirm`'s profile branch has any test. A wiring bug in either (wrong `getattr` dest, wrong path passed to the flush filter) would go completely undetected. (`bin/agentic-identity` ~L657-694 `cmd_auto`, ~L761-798 `cmd_confirm`)

2. **Real, untested behavioral divergence between the Python CLI and the Node hook mirror.** `_profile_config_dir()` (Python) and `_profileConfigDir()` (`hooks/stop-context.js`) are documented as exact mirrors of the same 6-tier env-detection logic, but they disagree on one concrete case: when the highest-precedence env var (e.g. `AGENTIC_CONFIG_DIR`) points at a directory that **doesn't exist yet**, Python's `_is_under_home` (via `Path.resolve()`, which doesn't require existence) accepts it and never falls through to check `CLAUDE_CONFIG_DIR`/`CODEX_HOME`; the JS version explicitly does `fs.realpathSync` and `continue`s to the next candidate on `ENOENT`. Concretely: `AGENTIC_CONFIG_DIR` set to a not-yet-created path + `CLAUDE_CONFIG_DIR` pointing at a valid, confirmed profile → the Python CLI reports "no profile identity" while the Node hook resolves the `CLAUDE_CONFIG_DIR` identity. Every existing fixture in both suites creates the candidate directory before testing it, so this is invisible to CI today.

### Medium (security/robustness)
3. **Unhandled crash when a profile config-dir resolves to a non-directory or unwritable path.** `_write_identity`'s `dest.parent.mkdir(mode=0o700, parents=True, exist_ok=True)` sits outside the surrounding `try/except` (`bin/agentic-identity:263`), reached from `cmd_init`/`cmd_auto`/`cmd_confirm` whenever `--scope profile` is used. For global/project scope `dest.parent` was always a fixed, trusted path; profile scope makes it derive from an env var that can point at, e.g., an existing regular file under `$HOME` - reproduced live: `FileExistsError [Errno 17]` propagates all the way out of `main()` as an uncaught Python traceback. This is new attack surface this PR introduces (bounded to a DoS of the CLI, not persistent compromise). Recommend wrapping the `mkdir` in the same `try/except` as `atomic_write` and returning a clean error like `_rejected_profile_dir_msg`.

### Verified sound (no action needed)
The core "values outside `$HOME` are rejected" property genuinely holds in both the Python and Node implementations - independently reproduced against symlink-escape and `../..` traversal payloads, not just trusted from docstrings/tests. Env-var precedence order is identical and equally strict in both languages (no weakest-link asymmetry). `flushPendingBuffer`'s per-profile partitioning correctly prevents cross-tenant flush leakage (K1-K4 tests pass). No shell/command injection surface. Identity files remain `0600` under `0700` parents.

### Minor / non-blocking
- No Python-side regression test for the symlink-escape containment check specifically (JS has a dedicated `[SYM]` test; Python's `_is_under_home` is almost certainly equally safe via `.resolve()`, but a future "simplification" to a lexical check would only be caught on the JS side).
- Missing `isinstance(str)` guard on `record.get("config_dir", "")` before `os.path.realpath()` in the pending-buffer flush path (`bin/agentic-identity:352-357`) - a malformed pending JSON file with a non-string `config_dir` raises an uncaught `TypeError`. Lower priority than #3 since it requires local write access to `.pending/`.
- PR body claims "8 new tests... 18/18 pass"; PR head actually has 12 new test functions and 22/22 total passing (confirmed by direct execution) - the suite genuinely passes, just a stale count in the description.
- `check-vision-alignment` CI check is failing because the PR body has no filled-in "## North Star alignment" section - that check is explicitly advisory (its own error message says so), not a merge blocker, but worth adding.
- (Out-of-scope, pre-existing, not filed as a ticket since Jira access is unavailable in this review environment) Neither `bin/tests/test_install_profile_config_dir.sh` nor `bin/tests/test_install_profiles.sh` appear to be wired into any CI workflow - only `test_codex_hooks_resolve.sh` is under `bin/tests/*.sh`. This predates this PR; flagging so it doesn't silently bit-rot, but it's not this PR's responsibility to fix.

TOCTOU between the containment check and file use is a documented, accepted residual risk in both languages (not a new finding here).

bloodf:
## Round-2 review fixes pushed (`003536df..4b80a4d7`)

Thanks for the second pass. All findings addressed:

| Finding | Fix | Commit |
| --- | --- | --- |
| **Major** — `cmd_auto`/`cmd_confirm --scope profile` had zero test coverage | New CLI-level tests drive the real binary via subprocess under a fake `$HOME`: (a) `cmd_confirm --scope profile` already-confirmed without `--force` → rejection (rc 2, file unchanged); (b) `cmd_auto --scope profile` → provisional write under the profile dir; (c) the `flushPendingBuffer(dev_id, profile_dir_filter=…)` call inside `cmd_confirm`'s profile branch flushes the tagged pending record end-to-end (a wrong getattr dest or flush-filter path fails the assert). | `4b80a4d7` |
| **Major** — Python/JS divergence when the highest-precedence env var points at a not-yet-created dir | The Node hook now matches the Python rule: on `realpathSync` ENOENT it falls back to the lexical path + the same `$HOME` containment and returns it (stops the scan) instead of continuing to a lower-precedence var. Verified against the reviewer's exact scenario plus a boundary matrix — both mirrors now agree on the active profile in every case, and the existing symlink-escape rejection is unweakened. | `c34a6055` |
| **Med** — `_write_identity` `mkdir` outside try/except → uncaught traceback when the profile dir parent is a regular file | `mkdir` moved inside the try; `OSError` → clean exit 1 with `cannot create profile dir …`. Test asserts rc 1 + message + no traceback. | `90d66ce8` |
| **Minor** — no Python symlink-escape containment test | Added, mirroring the JS `[SYM]` test. | `4b80a4d7` |
| **Minor** — non-string `config_dir` in a malformed pending record → `TypeError` on `realpath` | `isinstance(str)` guard before the compare; test with a non-string `config_dir` flushes cleanly. | `90d66ce8` |

On the advisory `check-vision-alignment`: happy to add a `## North Star alignment` section to the PR body if you'd like it green — flagging rather than editing the PR description unprompted.

Verified hermetically (`env -u AGENTIC_CONFIG_DIR -u CLAUDE_CONFIG_DIR -u CODEX_HOME`): pytest 30 passed + standalone green; node identity 24 + session-log 25 green; both install shell suites green; `node --check` + `bash -n` clean. K4 realpath round-trip and J8 4-tier back-compat still green.

tyson-solara6 (approved):
**Automated review (skeptic + security-auditor)**

This is the 3rd review round. Verified the two prior CHANGES_REQUESTED rounds were addressed with real fixes backed by regression tests proven to fail pre-fix:

- Symlink bypass in `hooks/stop-context.js` `_profileConfigDir()` — fixed via `_realpathNonStrict` (`c34a6055`, `2dbb8357`), covered by the `[SYM]` test.
- Cross-tenant pending-buffer leakage — fixed via `profile_dir_filter` in `flushPendingBuffer` (`bin/agentic-identity:214-272`), covered by `test_cli_confirm_profile_flushes_tagged_pending` and JS `[J4]`.
- `scripts/lib/identity.sh` Branch 3 not passing `--profile-dir` — fixed (`afd72556`), covered by `test_install_profile_config_dir.sh` Test 8.
- JS/Python parity gaps on the 6-tier resolution logic and env-precedence edge cases — fixed, with dedicated regression tests (K4, ENOENT/`[ENOENT]`).

Independently re-derived the 6-tier identity resolution logic in both `bin/agentic-identity` and `hooks/stop-context.js` — they agree. Security audit found no Critical/High/Medium issues: tenant-name whitelisting, quoted shell interpolation, symlink/TOCTOU guards, and cross-tenant flush partitioning are all sound and tested.

**Minor, non-blocking:**
1. `check-vision-alignment` CI check is failing only because the PR body has no `## North Star alignment` section — per the check's own text this is advisory, not a merge blocker. Worth adding for hygiene.
2. Security-auditor note: `bin/agentic-identity` `_write_identity()`'s `dest.parent.mkdir(mode=0o700, parents=True, exist_ok=True)` only applies `0700` on *creation* — a pre-existing, looser-permissioned profile config dir (e.g. an existing `0755` dir) keeps its permissions, so `identity.yml`'s parent dir may be more permissive than intended on shared boxes (the file itself is correctly `0600`). This was already flagged and explicitly deferred as out-of-scope in round 1 — still true. This is normally the kind of pre-existing/out-of-scope finding we'd file as a tracked follow-up ticket, but the Jira integration (mcp-atlassian) isn't available in this run's environment, so flagging it here instead: consider `os.chmod(dest.parent, 0o700)` defensively even when the dir pre-exists.

**Alignment:** ALIGNED with the North Star — extends identity/telemetry scoping for per-tenant installs while staying byte-identical by default (no new ceremony, no operator attention tax).

Approving — no Critical/Major findings remain open.

tyson-solara6 (changes requested):
## Automated review

Reviewed via security-auditor + qa-engineer (a third, correctness-focused skeptic pass stalled on a live `gh` call after several minutes and was dropped per prior experience with this agent on ad-hoc reviews — the findings below stand on their own).

### Security (security-auditor)
No Critical/Major findings. The `$HOME`-containment logic is sound: `_is_under_home()` in `bin/agentic-identity` and `_profileConfigDir()` in `hooks/stop-context.js` both realpath-resolve before the containment check, and the cross-tenant flush partition in `flushPendingBuffer` correctly realpaths both sides of the `config_dir` comparison. `config_dir` is confirmed stripped before it reaches the committed session-log line — no path/PII leak. Two Minor notes, non-blocking:
- `_profile_config_dir`/`_profile_identity_path` (bin/agentic-identity) return the *unresolved* `candidate` Path after the containment check passes, while the JS side always returns a fully-realpath'd string — the two implementations aren't symmetric at this layer (only the flush-comparison code papers over it). Consider returning `candidate.resolve()` in both Python helpers.
- The containment check and the later mkdir/write are separate filesystem operations (documented TOCTOU) — acceptable for this single-user local-tool threat model, flagging for visibility only.

### Test coverage (qa-engineer, static review — no UI/runtime surface here)
Partial gaps worth closing before merge:
- `bin/tests/test_agentic_identity.py`: no test drives a successful `init --scope profile` happy path (exit 0, provisional-false write) — only the mkdir-failure branch is covered for `init`.
- `cmd_show --scope profile` has **zero** command-level tests — only the underlying path helpers are unit-tested, not `cmd_show`'s three print paths (no-config-dir / no-identity / success).
- No test verifies that `bin/agentic-identity` (Python) and `hooks/stop-context.js` (JS) agree on the same fixture — the two suites assert matching *intent* in parallel, hand-mirrored suites, but nothing diffs their actual output on identical inputs. A precedence or containment change on one side not mirrored in the other would pass CI silently.
- No `init --scope profile --force` overwrite-of-confirmed test.

Symlink-escape, ENOENT-scan-stop, and non-string-`config_dir` guards *are* well covered on both sides — good adversarial coverage where it exists, the gaps above are specifically the newer `init`/`show` command surface.

### Verdict
Requesting changes on test coverage rather than correctness — the resolution logic itself looks sound per the security pass, but this is identity/telemetry infrastructure with broad blast radius (used by the whole methodology), and the missing happy-path/cross-language tests are exactly the kind of gap that lets a future regression on one side (Python vs JS) ship silently. Adding the four tests noted above should be enough to flip this to approve.

No pre-existing/out-of-scope issues were spotted in this PR — no Jira follow-ups filed for it.

tyson-solara6:
## Follow-up: additional Major finding from a deeper correctness pass

A third review pass (correctness-focused, running the actual resolution functions rather than just reading them) finished after my initial review above and surfaced a real bug that the security/coverage passes missed. Confirmed independently before posting:

**JS and Python config-dir resolution diverge on tilde-prefixed env values** — directly contradicting this PR's own parity claim.

- `bin/agentic-identity`'s `_profile_config_dir()` does `Path(os.path.expanduser(raw))` before the `$HOME`-containment check.
- `hooks/stop-context.js`'s `_profileConfigDir()` does `path.resolve(raw)` with **no** tilde expansion.

So `AGENTIC_CONFIG_DIR=~/.claude-tenant` resolves to `$HOME/.claude-tenant` in Python but to `<cwd>/~/.claude-tenant` (a literal `~` directory under cwd) in Node. Verified live:
- Python: `os.path.expanduser('~/.claude-tenant')` → `/Users/…/.claude-tenant`
- Node: `path.resolve('~/.claude-tenant')` → `<cwd>/~/.claude-tenant`

Both values can independently satisfy each side's own `$HOME`-prefix containment check (when cwd is under `$HOME`), so neither falls back to global — the two implementations silently read/write two *different* `identity.yml` files for the identical env var. That's exactly the "JS and Python disagree on where telemetry lands" failure mode this PR is supposed to prevent. The comment in `hooks/stop-context.js` (~line 1336, "so the two sides never disagree on which profile is active") is factually incorrect once a tilde-spelled value is used — a realistic case for CI/systemd-style env maps that don't shell-expand `~`. Neither test suite (`bin/tests/test_agentic_identity.py`, `hooks/tests/test-stop-context-identity.js`) exercises a tilde-prefixed value, so it's untested as well as unfixed.

**Fix:** expand `~` in `hooks/stop-context.js`'s `_profileConfigDir()` the same way the Python side does (or drop the Python-side expansion — pick one), plus a regression test in both suites using a literal `~`-prefixed env value.

This strengthens rather than changes the request-changes verdict above — please treat it as a blocking correctness item alongside the test-coverage gaps already noted.

bloodf:
Addressed the review-coverage findings in `1b95cc8c`:

- **`.resolve()` symmetry (Minor):** `_profile_config_dir` and `_profile_identity_path` now return resolved paths, matching the JS `_profileConfigDir`/`_realpathNonStrict` output — the two sides agree byte-for-byte on the active config dir instead of only the flush-comparison code papering over it.
- **init --scope profile happy path (L5):** new test asserts a non-provisional write + exit 0.
- **init --scope profile --force (L6):** overwrite-of-confirmed test (rejected without `--force`, succeeds with it).
- **cmd_show --scope profile (L7):** command-level coverage of all three print paths (no config-dir / no-identity / success).
- **cross-language agreement (X1):** one identical fixture (`AGENTIC_CONFIG_DIR` -> under-$HOME profile, confirmed profile beating confirmed global) run through both `agentic-identity show --scope effective` (Python) and `stop-context.js` (JS); the test fails if either side resolves a different winner.

Full Python suite (30 tests) and JS suite (24) both green.

tyson-solara6 (changes requested):

**Security:** clean — no Critical/High/Medium findings. Strong controls: tenant/profile names validated against a strict whitelist before path construction, symlink guards (CWE-59), consistent `$HOME`-containment checks on both the Python and JS sides with a cross-language parity test, handle format validated before being written into YAML, identity files written `0600`/dirs `0700`, no secrets handled. One accepted Low (TOCTOU on the containment check, requiring the attacker to already have write access inside the victim's own `$HOME` — equivalent trust level to already owning the account).

**Correctness — 1 Major finding, confirmed still open at the current head commit:**

The tilde-prefix divergence bug flagged in the 2026-07-14T21:12:02Z follow-up comment is **not fixed**. `hooks/stop-context.js:_profileConfigDir` still does `path.resolve(raw)` with no `~` expansion, while `bin/agentic-identity:_profile_config_dir` still does `Path(os.path.expanduser(raw))`. The only commit since that follow-up comment (`1b95cc8c7`) touched only `bin/agentic-identity` and its Python test — `hooks/stop-context.js` was not modified at all, so it structurally could not have addressed this. With `AGENTIC_CONFIG_DIR=~/.claude-tenant`, Python resolves to `$HOME/.claude-tenant` while Node resolves to `<cwd>/~/.claude-tenant` — when cwd is under `$HOME`, both independently pass their own containment check, so the two languages silently read/write different `identity.yml` files for the same env var. That's exactly the cross-language telemetry misattribution this PR exists to prevent, and neither test suite (`test_agentic_identity.py`, `test-stop-context-identity.js`) exercises a tilde-prefixed value.

**North Star alignment:** ALIGNED — opt-in per-profile scope, no added friction, back-compat preserved.

Requesting changes until `hooks/stop-context.js`'s `_profileConfigDir` gets tilde expansion matching the Python side (or vice versa), plus a regression test using a literal `~`-prefixed env value in both suites.

bloodf:
Fixed the tilde-expansion divergence (Major) in `dea3a93c`:

`hooks/stop-context.js:_profileConfigDir` now expands a leading `~` (bare `~` and `~/rest` -> `$HOME`) via a new `_expandUser()` applied before `path.resolve`, matching the Python side's `os.path.expanduser`. With `AGENTIC_CONFIG_DIR=~/.claude-tenant` both languages now resolve to `$HOME/.claude-tenant` instead of Node landing on `<cwd>/~/.claude-tenant`.

Regression tests using a literal `~`-prefixed env value added to both suites (they fail on the pre-fix code — Node would fall back to the global identity): `test_tilde_prefixed_env_expanded` in `test_agentic_identity.py` and the `[TILDE]` case in `test-stop-context-identity.js`. Both suites green (Python 31, JS 26).

tyson-solara6 (approved):
## Automated review — APPROVE

This is the 6th review round on this PR; the current head commit fixes the sole remaining Major from the last request-changes round (cross-language tilde-prefix divergence between `bin/agentic-identity`'s `_profile_config_dir` and `hooks/stop-context.js`'s `_profileConfigDir`), with a matching `_expandUser()` helper and dedicated regression tests added to both suites. I independently re-verified the prior rounds' Critical/Major findings (symlink-escape containment, cross-tenant pending-buffer leakage, `identity.sh` missing `--profile-dir`, JS coverage gaps) are actually fixed by reading the diffs, not just trusting the comment trail. All CI passes except the repo-wide DCO check.

A parallel security audit also reached APPROVE, with one **Major design note worth flagging even though it doesn't block**: the "cross-tenant leakage guard" this PR adds is attribution-only — it's a self-reported `config_dir` string compared at flush time, with no cryptographic binding or filesystem-level separation between profiles sharing one `$HOME`. Any process that can write into the shared `.pending/` directory (which any co-tenant profile running as the same OS user can do) could forge a pending record's `config_dir` and get it misattributed on the next `confirm`. Given the actual threat model here (cooperating profiles under one shared-user `$HOME`, not genuinely hostile multi-user isolation), this isn't a blocking vulnerability, but the commit-message language ("so one tenant's confirm can no longer attribute another tenant's sessions") somewhat overstates the guarantee — worth a doc clarification or a follow-up hardening pass if genuine hostile-co-tenant isolation is ever a goal.

**Minor (non-blocking):**
- `hooks/stop-context.js` `_expandUser()` only handles bare `~`/`~/rest`; Python's `os.path.expanduser()` also expands `~otheruser/...` via a pwd lookup. Narrow edge case (specific env-var spelling), doesn't bypass either side's `$HOME` containment check.
- Documented-and-accepted TOCTOU in both languages' containment checks (resolve-then-use) — already scoped in-code as same-user-only risk, no action needed.
- Out-of-scope, previously deferred across earlier rounds: `_write_identity()`'s `mkdir(mode=0o700)` only applies permissions on creation, so a pre-existing looser-permissioned config dir keeps its permissions. Noting again since it keeps resurfacing without being tracked anywhere; no Jira integration available in this environment to file it formally this run.

Nice hardening also confirmed in this diff: `_write_identity`'s mkdir is now wrapped in try/except (was an uncaught crash path), and the pending-buffer flush now coerces non-string `config_dir` values before comparison (was a `TypeError` risk on malformed JSON).

tyson-solara6:

tyson-solara6 (changes requested):
Requesting changes. The 6-tier resolver parity work between `bin/agentic-identity` and `hooks/stop-context.js` is solid (mirrored symlink/tilde/ENOENT regression tests are exactly right), but there are four Major issues, one of them a silent telemetry drop.

Major:

1. Silent telemetry drop - pending records are tagged on env presence, not on profile scope actually winning (`hooks/stop-context.js` `writePendingBuffer`). Every pending record gets `config_dir` whenever `AGENTIC_CONFIG_DIR`/`CLAUDE_CONFIG_DIR`/`CODEX_HOME` is set under `$HOME`, regardless of whether the resolved identity is profile-scoped. `flushPendingBuffer`'s global-confirm path then excludes all config_dir-tagged records. Net effect: a user with a plain global identity who merely has one of those env vars set (CODEX_HOME especially) buffers telemetry that `agentic-identity confirm` never flushes, until cap-100 eviction drops it - breaking the "telemetry is buffered, not lost" guarantee. Fix direction: tag only when the winning identity is profile-scoped, or have the global flush drain tagged records when no profile identity exists at that dir.

2. Merge conflict: the PR edits `.github/workflows/wrap-lock-tests.yml`, which main deleted in #461 in favor of glob auto-discovery (hence the DIRTY merge state). The edit is obsolete - after rebase, `hooks/tests/test-stop-context-identity.js` is already picked up by the `hooks-tests.yml` `hooks/tests/test-*.js` glob. Drop the workflow edit and rebase.

3. Doc-sync: README.md:301 still documents the 4-tier precedence list; the new profile tiers and `--scope profile` are undocumented in the identity section. Per the repo doc-sync obligation, the stale public list is in-scope for this PR. README:274's "no sessions lost" claim is also falsifiable while item 1 stands.

4. The conductor-side identity read was not updated: `content/sections/01-activation-preflight.md` and `content/references/events-log.md` still describe a two-file (project + global) 4-tier read. On a profile-scoped install, the conductor's first-turn identity notice resolves a different identity than the hook attributes telemetry to. Either extend the preflight or document the divergence explicitly (note: touching content/ also triggers adapter rebuilds).

Minor: `bin/tests/test_install_profile_config_dir.sh` is not run by any CI workflow (`bin-tests.yml` is pytest-only), so the new install-scenario assertions do not gate.

Verified clean: DS-66 `_lib.py` pattern not reintroduced; bin/hook resolver parity including `$HOME` containment via realpath on both sides; back-compat when no env is set; manifests updated; DCO on every commit; DRY helpers extracted.

tyson-solara6:
Superseded by #583, rebuilt from current main with complete regression coverage and fresh Tier-3 correctness/security sign-off.

PR #447, opened 2026-07-10
docs(ds52): add story-size preflight and token-reduction callout to /ticket-triage

tyson-solara6:
### What changed

Added a **Story-size preflight** block to `/ticket-triage` Phase 1, running once immediately after all ticket metadata is collected.

**Trigger:** any ticket with a `story_points` or Linear `estimate` >= 5 (Jira's `timeestimate` field is seconds-denominated and display-only; it never triggers this preflight). Tickets marked `terminal: true` or `in_progress: true` are excluded from the scan - they are deferred or removed from lane assignment regardless of size.

**On detection (per oversized ticket):**
- Prints a `⚠ [TICKET-ID] Est: N pts - large story` warning with a concrete split recommendation (decompose into <= 3-point sub-tickets, re-triage the sub-set)
- Sets a `context_risk: high` flag on the entry (display-only; no lane distribution rule consumes it)
- If any ticket has `context_risk: high`: appends a one-time token-reduction reminder pointing operators to `ctx_*` context-mode tools (~98% context savings vs raw shell output, see `content/rules/code-standards.md` §Context Window Management) and the `/wrap -> /clear -> re-invoke` mid-session reset pattern

**Silent on clean sets:** no output when no ticket exceeds the threshold.

**Artifact:** added a `⚠` column to the Phase 4a per-ticket summary table skeleton so `⚠ large` surfaces visually in triage artifacts for oversized tickets.

## North Star alignment

- Pillar(s) advanced (see docs/overview/vision.md): **Guard operator attention.** The preflight is silent on clean sets - it prints nothing unless a ticket actually needs decomposing, so it does not tax attention in the common case. When it does fire, it surfaces one concrete decision (decompose or proceed) rather than a status dump.
- Trade-off or pillar this could regress, if any: adds one more recurring operator-facing output type to a core command. Mitigated by the silent-on-clean-sets behavior above; a set with no oversized tickets sees zero additional output.

### Files changed

`content/commands/ticket-triage.md`, rebuilt across all 11 adapters: full copies regenerated for `.claude`, `.codex`, `.cursor`, `.gemini`, `.opencode`, `.github/prompts`, `.hermes`, `.openclaw`; `.kimi`, `.omp`, `.pi` symlink/read `content/` directly and show no diff; `.copilot` regenerated with no drift.

### Testing

Docs-only change - no behavioral logic, no state writes, no adapter build scripts modified.

### DIFF
--- a/.claude/commands/ticket-triage.md
+++ b/.claude/commands/ticket-triage.md
@@ -118,7 +118,7 @@ Conductor-direct (no subagent). For each entry in `normalized_input.entries[]`, 
 - **Jira:** `mcp__mcp-atlassian__jira_get_issue` - capture `priority`, `status`, `story_points` (or `timeestimate`), `labels`, `components`, `assignee`, and `issuelinks` (blocks / is-blocked-by / relates-to).
 - **Linear:** `mcp__linear__get_issue` - capture `priority`, `state`, `estimate`, `labels`, `assignee`, and relations (blocks / blocked-by / related).
 
-The captured estimate (`story_points` / `timeestimate` / Linear `estimate`) populates the display-only "Est" column in the Phase 4a per-ticket summary table AND triggers the story-size preflight below.
+The captured estimate (`story_points` / `timeestimate` / Linear `estimate`) populates the display-only "Est" column in the Phase 4a per-ticket summary table; no distribution rule consumes it, and estimate-aware lane sizing is a deferred default. Only `story_points` and Linear `estimate` trigger the story-size preflight below - `timeestimate` is Jira's time-tracking field (denominated in seconds), display-only, and is never compared against the preflight threshold.
 
 **Soft-fail per ticket:** on any fetch error, mark `fetch_failed: true` on that entry and proceed. Fetch-failed tickets are treated as independent (no known deps, no known metadata) in all downstream phases.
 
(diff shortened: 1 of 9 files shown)


tyson-solara6:
**Automated review note** (self-authored PR — no formal verdict per review policy; posting findings from an adversarial pass before merge).

**Major** — `content/commands/ticket-triage.md:127-159` (and all 5 mirrors) places the "story-size preflight" in Phase 1 ("runs once, immediately after all metadata is collected"), but its token-reduction reminder fires "if any `context_risk: high` ticket **is lane-assigned** (not deferred)" — and lane assignment doesn't exist until Phase 3 runs, several phases later. As written, an agent executing Phase 1 can't evaluate that condition yet. Either move the reminder to Phase 3/4a, or say explicitly the callout is deferred/printed later even though documented in Phase 1.

**Major** — line 125 says the estimate trigger is `story_points` / `timeestimate` / Linear `estimate`, but line 129's actual check only tests `story_points` or Linear `estimate` — `timeestimate` is silently dropped, and there's no time-to-points conversion defined anywhere. A ticket estimated only via `timeestimate` is promised a check that never fires.

Minor: 4 em dashes introduced on lines 127/132/142/146 — worth a second look since it runs against the repo's own stated global no-em-dash rule; line 147 uses "→" while the rest of the file uses ASCII `->`; the `[DS-XX]` placeholder on line 132 borrows this repo's own ticket prefix in a doc meant to run against arbitrary Jira/Linear projects (suggest `[<ticket_id>]`).

Cross-file consistency: verified all 6 copies (`content/`, `.codex/`, `.cursor/`, `.claude/`, `.opencode/`, `.gemini/`) are byte-identical for the new block — no drift.

Out-of-scope (pre-existing, not filed as a ticket — no Jira access in this run): the pre-existing "Est" column already merges `story_points`/`timeestimate` into one display value with no unit normalization (same ambiguity as the Major above, predates this diff); Phase 0's "large-list gate" and Phase 2b's `HEURISTIC_ONLY` gate both hardcode the threshold `20` independently.

tyson-solara6:
## Automated review — advisory (own PR, no formal verdict per self-review rules)

Investigated the 3 failing CI checks on this PR — all have real, confirmed causes:

1. **`check-adapter-sync` (real bug):** the PR description says "5 adapter copies" but 3 more adapters mirror `ticket-triage` and were missed — `.github/prompts/ticket-triage.prompt.md`, `.hermes/SKILL.md`, and `.openclaw/skills/ticket-triage/SKILL.md` are all still on the old content (no Story-size preflight block, no `⚠` column). Confirmed by rebuilding each adapter transform locally from `content/commands/ticket-triage.md` and diffing — the drift is exactly the same content already applied to the other 5 mirrors. Fix: run `bash scripts/build-all.sh` and commit the regenerated output for those 3 files.
2. **`check-vision-alignment` (confirmed real):** the PR body genuinely has no `## North Star alignment` heading — not empty, just absent. This is advisory-only (not a merge blocker per the check's own error message), but easy to fix by editing the PR description to add the section.
3. **DCO check:** commit `8233722` has no `Signed-off-by` trailer — mechanical, needs `git commit --amend -s` + a re-push. Left for manual handling (no force-push automated here).

The content that *was* regenerated (the 5 mirrors + `content/`) is internally consistent — thresholds (≥5pts warn, 3-4pts no-op, ≤2pts/absent no-op) match everywhere I checked.

**Pre-existing, out-of-scope bug spotted (not introduced by this PR):** `.gemini/commands/ticket-triage.toml`'s generated `description` field renders as `<!--` — the `.gemini` build script's description-extractor grabs the first non-header line of the file body, and `content/commands/ticket-triage.md` opens with an HTML-comment manifest block, so it captures the comment delimiter literally instead of skipping it. This affects every command with a leading HTML-comment manifest, not just this one. Would normally file this as a follow-up ticket, but the Jira MCP integration is unavailable in this run's environment — flagging here so it isn't lost; worth filing manually against `.gemini/build.sh`.

tyson-solara6:
Automated review note (informational only - self-review is blocked on our own PRs, so no formal verdict here).

The story-size preflight logic and docs prose are sound and internally consistent. Flagging the current CI failures since they look like real gaps, not flakes:

- `check-adapter-sync`: the PR updates `content/commands/ticket-triage.md` plus 5 adapter copies, but 3 more generated files that also derive from that same source are now stale - `.github/prompts/ticket-triage.prompt.md`, `.hermes/SKILL.md`, and `.openclaw/skills/ticket-triage/SKILL.md`. The description's "rebuilt 5 adapter copies" undercounts the full build matrix (11 scripts per `adapter-sync.yml`) - worth running the remaining build scripts and pushing a follow-up commit.
- `check-vision-alignment`: failing because the PR body has no "## North Star alignment" section at all (this touches `content/*`, which triggers the requirement) - a one-line N/A would clear it.
- DCO failure is the expected missing sign-off trailer on the commit.

No pre-existing out-of-scope issues found - the sync gap above is caused by this PR's own change, not pre-existing.

PR #448, opened 2026-07-11
fix(build): glob content/templates/.agentic seeds so new templates propagate to adapters

tristanlee85:
## Summary
Four adapter `build.sh` scripts (`.claude`, `.cursor`, `.codex`, `.gemini`) hardcoded a `config.json learnings.md` list when hardlinking `.agentic` seed templates into adapter template dirs. The tracked seed `content/templates/.agentic/skill-candidates.md` was never propagated, so `bin/agentic-migrate` v4 scaffolding hit its missing-template branch, returned exit 3, and never stamped `scaffolding_version` - a silent break for every project on a machine.

Replace each hardcoded loop with a glob over `content/templates/.agentic/*` guarded by `[[ -f ]] || continue`, preserving each script's own hardlink + inode-equality semantics and destination variable (two variants: inline `ln`+inode in `.claude/build.sh`; the `hardlink_from_content` helper in the other three). Future seed templates now auto-propagate, so this bug class cannot recur.

## Changes
- 4 build scripts: hardcoded template list -> glob.
- 4 regenerated adapter artifacts: `skill-candidates.md` now committed under `.claude`/`.codex`/`.cursor`/`.gemini` `templates/.agentic/` (adapter-sync CI only diffs tracked files, so the seed must be committed or the fix is inert for bare-pull consumers).

## North Star alignment
Advances **Low friction (P3)** and **Works for everyone (P4)**: removes a silent, per-machine breakage in v4 scaffolding; new seed templates now propagate to every adapter automatically, so no maintainer must remember to update four hardcoded lists - portable by construction. Also advances **Verifiable outcomes (P2)**: eliminates a silent-failure class (errored-but-no-stderr, exit 3 with no version stamp), so migration cleanly applies (exit 0) or is a real, visible error.

Trade-off (named plainly): the glob trades an explicit two-item allow-list for "propagate whatever is in the dir" (Skeptic Minor). Low-risk - `agentic-migrate` only seeds manifest-listed files, and `content/templates/.agentic/` is curated to seeds only. No pillar materially regressed; the attention test improves (one fewer silent break for the operator to notice and hand-fix).

## Review
- Fresh independent Skeptic: semantics preserved per-variant, glob-safe under `set -euo pipefail` (no `nullglob`/`dotglob`), `hardlink_from_content(src,dst)` signature confirmed, 4-adapter scope correct (the other 7 adapters have no such loop). Functionally verified by running builds in an isolated worktree.
- Skeptic Major ("regenerated artifacts must be committed") resolved: `skill-candidates.md` is tracked in all 4 adapter dirs in this PR; `check-adapter-sync` passes.

## Test plan
- [x] `bash -n` on all four scripts passes
- [x] Full `build-all` regenerates cleanly; `git status` shows only the 4 scripts + 4 new seeds, no unrelated churn
- [x] CI: adapter-sync, methodology-drift, codex-skill-sync, DCO green

### DIFF
--- a/.claude/build.sh
+++ b/.claude/build.sh
@@ -78,10 +78,11 @@ else
   ln "$SCAFFOLDING_SRC" "$SCAFFOLDING_DST"
 fi
 
-# templates/: hardlink .agentic seed files
+# templates/: hardlink .agentic seed files (glob so new templates auto-propagate)
 mkdir -p "$SKILL_DST/templates/.agentic"
-for tmpl_name in config.json learnings.md; do
-  TMPL_SRC="$CONTENT/templates/.agentic/$tmpl_name"
+for TMPL_SRC in "$CONTENT"/templates/.agentic/*; do
+  [[ -f "$TMPL_SRC" ]] || continue
+  tmpl_name="$(basename "$TMPL_SRC")"
   TMPL_DST="$SKILL_DST/templates/.agentic/$tmpl_name"
   if [[ -e "$TMPL_DST" ]] && [[ "$(get_inode "$TMPL_SRC")" == "$(get_inode "$TMPL_DST")" ]]; then
     :
--- a/.claude/skills/agentic-engineering/templates/.agentic/skill-candidates.md
+++ b/.claude/skills/agentic-engineering/templates/.agentic/skill-candidates.md
@@ -1,0 +1,34 @@
+# Skill Candidates
+
+Recurring workflow friction detected across sessions. Each entry represents a
+domain that has accumulated enough occurrences to warrant considering a new
+skill. Detection signal: `tool_failure_workaround` events (by `domain_tag`) and
+`.agentic/learnings.md` entries (by `Domain`), threshold >= 3 across sessions.
+
+Classification rules and routing taxonomy are in
+`content/commands/skill-candidates.md`. The detector is in
+`hooks/lib/skill-candidate-detector.js`.
+
+## Entry format
+
+Each candidate uses this exact format. The `## <domain>` heading is the unique
+key (no separate id field). `Status` is set to `open` by the detector; a human
+edits it to `dismissed` to suppress future session-start notices.
+
+```markdown
+## <domain>
+**Count:** <lifetime occurrence count>
+**Suggested artifact:** command | named-agent | preset | lint-rule
+**First seen:** YYYY-MM-DD
+**Last seen:** YYYY-MM-DD
+**Status:** open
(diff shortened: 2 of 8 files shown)


tyson-solara6 (approved):
**Automated review (skeptic)**

Verified the fix directly: `content/templates/.agentic/` currently contains 3 seed files (`config.json`, `learnings.md`, `skill-candidates.md`). The old hardcoded loop only copied the first two, silently dropping `skill-candidates.md`; the new glob (`"$CONTENT"/templates/.agentic/*` with an `[[ -f ]] || continue` guard) picks up all three. Checked all 4 touched `build.sh` files (`.claude`, `.codex`, `.cursor`, `.gemini`) — consistent with each script's existing hardlink/dedup convention, no semantic drift. Also checked the other 7 adapter directories in the repo (`.copilot`, `.kimi`, `.pi`, `.omp`, `.openclaw`, `.opencode`, `.hermes`) for the same pattern — none reference `templates/.agentic`, so there's no completeness gap; this PR's scope is correct and complete.

**Minor, non-blocking:** the bare `*` glob (no `dotglob`) would silently re-skip any *future* seed template whose filename starts with a dot — the same silent-drop failure class this PR is fixing, just for a different filename shape. Not urgent since today's 3 seed files are all non-dotfiles, but worth a follow-up if dotfile-named seeds are ever added.

**Alignment:** ALIGNED — build-script hygiene that keeps template propagation correct, no operator attention added.

Small, correct, well-scoped fix. Approving.

tyson-solara6 (approved):

Verified:
- Deterministic, safe under `set -euo pipefail`: no-nullglob empty-dir case is skipped by the `[[ -f ]]` guard; quoting handles spaces; output is order-independent (fixed dest paths, not concatenation), so no adapter-sync flapping.
- All 11 adapters accounted for: the 4 hardlink adapters (.claude/.codex/.cursor/.gemini) had the bug and are fixed; the 5 dir-symlink adapters already auto-propagate; .copilot/.hermes ship no .agentic templates.
- Regenerated `skill-candidates.md` seeds committed to all 4 adapter dirs, byte-identical to source.
- DCO clean on every commit; no stale counts in README/CONTRIBUTING/SKILL.

Minor (non-blocking):
- The glob has no `dotglob` and `[[ -f ]]` skips dirs, so a future hidden seed (`.agentic/.something`) or nested seed subdir would silently not propagate on the 4 hardlink adapters while the symlink adapters would pick it up. Fine for today's flat curated seed set, but worth a comment or `shopt -s dotglob` if the seed set ever grows.

PR #453, opened 2026-07-15
fix(hooks): bounded stdin guard across all 7 hooks + real Cursor stop-context port

tristanlee85:
Fixes the Cursor "taking longer than normal to close" hang caused by the AE stop hook, and gives Cursor real stop-context parity with the Codex/Gemini/Copilot adapters.

## Root cause

Two independent defects in the Cursor `stop` hook wiring, verified by live reproduction:

1. `hooks/stop-context.js` blocked forever on `fs.readFileSync(0)` when the spawning harness never closes stdin - Cursor's composer session-end path does exactly that, so Cursor shutdown waited on the hook. The same blocking-read pattern existed in 6 more hooks (codex/gemini/copilot ports, capture-nudge, session-end-wrap, spawn-emit).
2. `.cursor/hooks.json` registered a relative `node hooks/stop-context.js` (MODULE_NOT_FOUND outside the AE workspace) pointing at a Claude-payload-shaped script that no-ops on Cursor's schema - all cost, no benefit.

## What this PR does

- **`hooks/lib/stdin-guard.js` (new):** shared bounded-stdin reader. First-byte timeout 750ms, early-completion-by-parse (a complete JSON payload resolves immediately - no EOF needed), inactivity timeout 5s re-armed per chunk, `unref()` on every resolution path so a held-open pipe can never hold the process. Never rejects, never truncates a slow-but-legitimate payload (proven by chunked-delivery regression tests).
- **All 7 stdin-reading hooks hardened** with the guard: `stop-context.js` (minimal 4-edit diff, twelve write paths untouched), the codex/gemini/copilot ports (+ regenerated `.github/hooks` build copy), `post-tool-use-capture-nudge.js`, `session-end-wrap.js` (call site restructured to `.finally()`-gated exit so deferred-wrap finalization still runs), `pre-tool-use-spawn-emit.js`.
- **`.cursor/hooks/stop-context-cursor.js` (new):** real Cursor port. Maps the official documented stop payload (`workspace_roots[0]` -> cwd with canonicalization, `conversation_id`, `transcript_path` with a 256 KB stat-gate) and writes `<cwd>/.agentic/context.md` for cross-harness continuity. Always exits 0 with `{}` on stdout; never emits `followup_message` (Cursor auto-submit landmine); never exits 2.
- **`.cursor/install.sh` converge:** the stop entry is rewritten to an absolute port path on every install run - fresh copies and existing `~/.cursor/hooks.json` files both converge (path-segment-anchored match so custom entries like `my-stop-context.js` survive byte-for-byte; atomic write; fail-open on malformed JSON or missing python3). Template stays relative by design; install substitutes.
- **Tests:** 10 new/extended test files - per-hook open-silent bounded-exit, slow-chunked no-truncation, Cursor port contract (stdout shape asserted on every case), finalization regression with `deferred_wrap_daemon: true` fixture, install converge/idempotency/anchor regressions.

## Review rigor

Skeptic-signed Brief and architect plan (3 review rounds each), 5 per-unit adversarial code reviews (Tier 3 on the lib and on `stop-context.js`). Review-caught-and-fixed before merge: missing `stdin.unref()` (held-open pipe kept an idle consumer alive - the exact bug class this PR fixes), a traversal guard that silently no-op'd on trailing-slash workspace roots, and an unanchored converge regex (same class as the U6 settings-prune bug).

## Evidence

- All 7 hooks exit 0 in ~790-804ms with stdin open and silent (was: indefinite hang).
- `node hooks/stop-context.js </dev/null` unchanged at ~38ms; full valid payload chunked over 2.1s processed with zero truncation.
- Independent QA pass: 9/9 runtime scenarios, 965 corroborating assertions, 0 failures; end-to-end run of the exact converged command string from an unrelated cwd writes context.md correctly.
- `bash scripts/build-all.sh` + adapter diff: clean (no content/ changes; `.github/hooks` copy regenerated in-commit).

## Operator steps after merge (not covered by CI/QA)

```bash
bash .cursor/install.sh    # converges the live ~/.cursor/hooks.json to the port
bash .claude/install.sh    # re-syncs the Claude hook snapshot (picks up stdin-guard)
bash .gemini/install.sh    # same, if the Gemini adapter is enabled
```
Then close Cursor with an active composer session and confirm the slow-close dialog is gone (Brief rubric R1, judgment line).

## North Star alignment

Advances **low friction** (eliminates an app-close hang that taxed every Cursor session) and **works for everyone** (Cursor gets the same session-continuity substrate as Claude/Codex/Gemini/Copilot; runtime `$REPO_DIR` resolution stays portable). No pillar regressed: Claude Code behavior is bit-identical on the normal path, enforced by the existing 400+ assertion hook suite plus new regression tests.


tyson-solara6 (changes requested):
**Changes requested** - 4 Major, 3 Minor.

The design is right and the engineering is solid: fail-open wiring is correct at all 8 call sites, the `stdin.unref()`-vs-`pause()` distinction is correctly diagnosed, `session-end-wrap.js`'s `.finally(() => process.exit(0))` is subtle and right, and the tests are real (subprocesses, OS pipes, a case-8 regression test that provably hangs on pre-fix code). All 7 adapters use identical default options, no divergence.

Two defects below are reproducible against this branch's HEAD and CI cannot see either, because no test exercises a chatty writer or a multi-MB payload.

---

### Major 1 - the guard is time-bounded only, never size-bounded

`hooks/lib/stdin-guard.js:181-207`

`armInactivityTimer()` is re-armed on every chunk. There is no absolute deadline and no byte cap, so a writer that drips data faster than `inactivityTimeoutMs` keeps the hook alive indefinitely while `accumulated` grows without limit.

Repro: a writer sending 1KB of non-JSON every 200ms (well under the 5000ms inactivity window), holding the pipe open.

```
WRITER: gave up after 59 chunks; child still alive = true
```

The hook never exited. `onData` clears the first-byte timer, appends, fails `tryParse`, then calls `armInactivityTimer()` which resets the only remaining bound. Nothing caps total bytes or total duration. This is the exact hang the module exists to prevent, and the PR title says "bounded".

Suggested fix: arm an absolute deadline once (never re-armed), plus a `MAX_STDIN_BYTES` cap that resolves early once exceeded. Existing case 5 covers stall-after-partial, not continuous-never-parses, so this needs its own test.

### Major 2 - `tryParse` is O(n^2) on the normal path

`hooks/lib/stdin-guard.js:193-205`

`tryParse(accumulated)` re-parses the entire accumulated string on every chunk. Measured with a single well-formed JSON payload where the writer closes immediately (the ordinary harness path):

| payload | this PR | pre-PR `readFileSync` | regression |
|---|---|---|---|
| 1 MB | 55 ms | 21 ms | 2.6x |
| 4 MB | 799 ms | 25 ms | 32x |
| 8 MB | 3236 ms | 30 ms | 108x |

4MB to 8MB doubles the input and quadruples the time. Node reads pipes in 64KB chunks and `JSON.parse` must scan the whole prefix before discovering truncation, so an n-byte payload costs n/65536 parses averaging n/2 bytes each.

Scoping this honestly: at typical hook payloads (under ~100KB, 1-2 chunks) the cost is negligible. It stays Major because this is a shared utility whose manifest advertises "~3-20ms", it is an algorithmic regression against the `readFileSync` it replaces, and it lands on `pre-tool-use-spawn-emit.js` (every Agent spawn) and `post-tool-use-capture-nudge.js` (whose payload embeds `tool_response`).

Suggested fix: gate the parse attempt on a cheap precheck (only attempt when the trimmed tail is `}` or `]`), or track brace/quote depth incrementally for O(n) total.

Majors 1 and 2 share a remediation surface - both live in `onData`/`armInactivityTimer`, and an absolute deadline + byte cap + parse precheck closes both.

### Major 3 - `stdin-guard.js` manifest is false as shipped

`hooks/lib/stdin-guard.js:45` states:

> This unit ships the module and its own tests only; no consumer is modified yet.

This PR wires 9 consumers. It reads like it was written for a per-unit landing that got squashed into one PR. A maintainer reading this file would believe nothing depends on it yet and could change the contract freely. The consumer list at `:33` also omits `.github/hooks/stop-context-copilot.js`.

### Major 4 - `pre-tool-use-spawn-emit.js` Performance manifest is stale

`hooks/pre-tool-use-spawn-emit.js:63-64` - all three claims are now wrong:

- "~1-3 ms typical" - up to 799ms at 4MB per Major 2
- "one JSON parse" - now one per chunk
- "Runs **synchronously** on the PreToolUse critical path" - `run()` is now `async`

The adjacent Failure-modes block in this same manifest was edited (+18/-7), so this is a miss inside a section already being touched.

---

### Minor

- `hooks/AGENTS.md:37` - the `lib/stdin-guard.js` row lists 4 consumers and omits 5, including `.cursor/hooks/stop-context-cursor.js`, this PR's headline new file. The stated counts ("Thirteen scripts", "5 Node lifecycle handlers") were correctly updated, so this is a non-misleading omission.
- `hooks/lib/stdin-guard.js:126-128, 195-205` - early-completion-by-parse can silently truncate a top-level scalar: `123456` arriving as `123`+`456` resolves as `123`, discarding the rest. Not reachable today (every harness payload is a top-level object, whose proper prefixes never parse), but `tryParse` is documented public API with `JSON.parse` as the default. Worth a manifest note that the default assumes object/array payloads.
- `.cursor/install.sh:486` vs `:410` - when `python3` is absent the converge function warns to stderr and leaves the file untouched, but the fresh-copy branch unconditionally prints `+ hooks.json copied to $HOOKS_DST (stop -> $CURSOR_STOP_CMD)`, claiming a convergence that did not happen. The installed `hooks.json` retains the template's relative `node hooks/stop-context.js`, which will not resolve - a silently broken hook alongside a success message. Gate the message on the returned count.

---

### Checked, no finding

- **Silent truncation producing malformed JSON downstream**: does not occur. Every consumer wraps `JSON.parse(raw)` in try/catch and exits 0, so truncation degrades to a silent no-op rather than corruption.
- **Empty / TTY / already-consumed / non-UTF-8 stdin**: all correct. `setEncoding('utf8')` delegates to `StringDecoder` (case 9 covers a 4-byte emoji split mid-codepoint), `setEncoding` is try/caught for a destroyed stream, and TTY falls to the 750ms first-byte timeout.
- **Dangling timers**: none. `finish()` -> `cleanup()` -> `clearTimers()` on every path.
- **Fail-open at all 8 call sites**: verified individually. None throws, exits non-zero, or writes stdout where it should not.
- **Adapter consistency**: all 7 use `readStdinGuarded()` with default options. `.github/` and `.copilot/` are byte-identical by pre-existing mirror, not this PR's duplication.
- **Cursor port fidelity**: matches the family convention. `hooks/stop-context.js:1076` itself writes `context.md` with a plain `writeFileSync` (tmp+rename is used only for JSON state files), and the codex/copilot ports likewise do no session-log append or identity resolution. No atomicity issue here.
- **Test quality**: not vacuous. All 10 cases are invoked from `main()`, and the conditional `if (parsed)` guards sit behind a preceding `assert(!!parsed)` so they cannot mask a failure. The gap is coverage, not rigor.

Thanks for this - the stdin-guard direction is clearly the right call, and the two Majors are localized rather than structural.

tristanlee85:
All 7 findings addressed; pushed to `c87936d6`.

**Majors**
- **Major 1 (guard was time-bounded only)** `27d6801b`: added `absoluteTimeoutMs` (10s default, armed once at read start, never re-armed) and `maxStdinBytes` (10 MiB default, tracked via an incremental byte counter so the check stays O(chunk)). Both fail-open via `finish(accumulated)` and are cleared in `clearTimers()`. New regression cases `testAbsoluteDeadlineBackstop` + `testMaxStdinBytesBackstop`. Pre-fix attestation: case 10 resolves at ~3.3s (EOF, not the deadline) and case 11 at ~2.1s (EOF, not the cap); post-fix both resolve via the new bounds.
- **Major 2 (O(n^2) parse on the normal path)** `27d6801b`: the parse attempt is now gated on a cheap backward tail-scan (`lastNonWhitespaceChar` -> `}`/`]`), so the accumulated buffer is only parsed when it can plausibly be complete. The backward scan handles a trailing newline that a naive `endsWith('}')` would miss.
- **Major 3 (false manifest)** `27d6801b`: removed the "no consumer is modified yet" line and listed all 9 consumers, including the `.github/hooks/stop-context-copilot.js` generated mirror the list had omitted.
- **Major 4 (stale Performance manifest)** `27d6801b`: rewrote the `pre-tool-use-spawn-emit.js` block to match reality (async `run()`, bounded-not-synchronous, parse gated by precheck) and its own Failure-modes paragraph.

**Minors**
- **`hooks/AGENTS.md:37`** `27d6801b`: the `lib/stdin-guard.js` row now names all 9 consumers (the "Thirteen scripts" / "5 Node lifecycle handlers" counts were already correct and left untouched).
- **scalar truncation** `27d6801b`: with the tail precheck, top-level scalar payloads no longer early-complete mid-stream (they resolve at EOF or via the bounded backstops), which removes the `123`+`456` truncation path. Documented as a manifest scope note.
- **`.cursor/install.sh:486`** `c87936d6`: the fresh-copy branch now captures the converge helper's returned count (mirroring the exists-branch) and prints the success line only when convergence actually happened; otherwise it reports non-convergence. Verified with a python3-stripped dry-run that reproduced the old false-success and confirmed the corrected message (file left pointing at the un-converged relative command).

**Also folded in** `2cf85632` (same manifest-accuracy theme): the sibling `.cursor/hooks/stop-context-cursor.js` Performance manifest now lists the new `absoluteTimeoutMs` / `maxStdinBytes` bounds (same class as Major 4), and the `stdin-guard.js` header "seven hooks" provenance is reframed as the historical extraction count so it does not read as stale next to the corrected "9".

~~Left as-is (out of scope for this review): the stdin-guard test suite (these new cases plus the prior case-8 regression) is not wired into any CI workflow, so it was run manually - `node hooks/tests/test-stdin-guard.js` 55/55, plus the 6 sibling integration suites green. Wiring the suite into CI is worth a separate follow-up.~~

**Correction (superseded):** this was inaccurate - the stdin-guard suite IS wired into CI via `.github/workflows/hooks-tests.yml`'s `test-*.js` glob (#461, DS-88). The `hooks-js-tests` check runs `node hooks/tests/test-stdin-guard.js` (55/55) and the sibling suites on every push/PR, so no separate wiring is needed. Thanks @tyson-solara6 for the catch.

tyson-solara6 (changes requested):
Round 2 review. Verified each of the 7 claimed fixes from the prior CHANGES_REQUESTED against the diff directly (not just the summary) via an independent skeptic pass, plus a fresh security audit focused on the new stdin-guard/install.sh/Cursor-port surface.

## The 7 prior findings: all genuinely fixed, no new Critical/Major bugs

- **Major 1** (unbounded time/size growth) — fixed correctly: `absoluteTimeoutMs` armed once (never re-armed) + `maxStdinBytes` byte cap, both fail-open via `finish()`, cleared in `clearTimers()`. New tests exercise real timing/content, not just presence.
- **Major 2** (O(n²) reparse) — fixed correctly via the tail-scan gate (`lastNonWhitespaceChar` check before `JSON.parse`). Checked adversarial cases (`{"a":"}"}`, split-mid-string) — the surrounding try/catch means a tail-scan false positive only costs a wasted parse, never an incorrect early resolution.
- **Major 3/4** (false manifests) and the **AGENTS.md / scalar-truncation / install.sh converge-message** minors — all confirmed fixed against the diff, logic-checked.

One correction for the PR description: the claim that the stdin-guard test suite is "not wired into any CI workflow — run manually only" is **factually wrong** — `.github/workflows/hooks-tests.yml` globs `hooks/tests/test-*.js` with no exclusion for the new stdin-guard test files, so they already run in CI on every push/PR. Good news, but please fix the PR description so nobody wastes time re-wiring something already wired.

## New findings from this round (both in code this PR introduces — requesting these be addressed here, not deferred)

**Medium — TOCTOU + unvalidated arbitrary local file read in the new Cursor port**
`.cursor/hooks/stop-context-cursor.js:231-244` (`readTranscriptExcerpt`) reads `transcript_path` straight from the untrusted payload / `CURSOR_TRANSCRIPT_PATH` env var with no canonicalization or allow-listing before `fs.statSync` → `fs.readFileSync`. This is new surface: the sibling ports (`stop-context.js`, codex/copilot/gemini) never read `transcript_path` content, only `cwd` joined with fixed subpaths. Two consequences: (a) a compromised/misconfigured harness can get up to ~2000 trailing chars of an arbitrary local file written into `<cwd>/.agentic/context.md`, which later sessions read as trusted context; (b) `statSync` then `readFileSync` is a classic TOCTOU — swapping a symlink between the two calls defeats the 256KB gate entirely, reintroducing the same unbounded-read-into-memory pattern this PR otherwise fixes. Suggest: realpath/canonicalize + re-stat after opening an fd (`fstat`), or read via a bounded stream instead of `statSync`+`readFileSync`.

**Medium — silent fail-open reported as success in `.cursor/install.sh`**
`.cursor/install.sh:405-410` (malformed JSON) and `:386-390` (missing python3) both return `changed=0`, but the existing-file branch (`:456-461`) unconditionally prints `"already current - preserving customizations"` for any `changed=0` — including genuine convergence *failure*. An operator has no way to tell "hooks.json is correctly wired to the bounded reader" from "convergence silently failed, the old blocking hook command is still active." This directly undercuts the reliability guarantee this PR exists to ship. Suggest: have the python helper emit a distinct sentinel on the exception/no-python3 paths and branch install.sh on it with a clearly non-reassuring message (the fresh-copy branch already does the right thing here per the round-1 fix — mirror that pattern into the exists-branch).

Both are Medium, not Critical/High — they require attacker control of the hook payload/env or local file swap, a bar that already implies some trust-boundary compromise — but they're new code in this exact diff and straightforward to close, so requesting they land before merge rather than as follow-ups.

## Non-blocking

- No large-payload (e.g. 8MB) benchmark test quantifying the claimed near-O(1) parse improvement — logically sound, just unverified at scale.
- This branch currently shows merge conflicts with `main` (`mergeStateStatus: DIRTY`) — will need a rebase regardless of the above.

Nice work closing out the round-1 findings cleanly — the design direction (bounded reader, fail-open everywhere) is right, this is about closing the two gaps the new attack surface opened.

tristanlee85:
Round-2 items addressed. Rebased onto current `main` (merge `35330f22`); the branch is now MERGEABLE and all checks are green - including `hooks-js-tests`.

**Merge conflict** (`hooks/pre-tool-use-spawn-emit.js`): resolved to the bounded `await readStdinGuarded()`. This preserves #461's fd-0 correction - `readStdinGuarded()` reads `process.stdin` (fd 0), never the `/dev/stdin` path - while keeping the non-blocking behavior this PR ships.

**Medium 1 - TOCTOU + unbounded arbitrary read** (`readTranscriptExcerpt`): fixed in `0a8dceae`. Opens `transcript_path` once (`O_RDONLY|O_NONBLOCK`), runs `fstatSync` + `readSync` on that same fd so check and read share one inode (TOCTOU closed), rejects non-regular files via `isFile()`, and caps the read at `min(size, 256 KB)`. Added symlink-swap and writer-less-FIFO regression tests, both proven to fail on pre-fix code.

**Medium 2 - silent fail-open reported as success** (`.cursor/install.sh`): fixed in `79440569`. The converge helper now emits a distinct `ERR` sentinel on every genuine-failure path (no python3, invalid JSON, missing `hooks`/`hooks.stop`, atomic-write exception); both call sites branch three ways - `ERR` to an actionable failure message, `>0` to converged, else already-current - with the string sentinel tested before any arithmetic (`set -e` safe). Supersedes `c87936d6`'s partial fresh-copy-only gating.

**PR-description correction**: correct - the stdin-guard suite IS wired into CI via `.github/workflows/hooks-tests.yml`'s `test-*.js` glob (#461). The `hooks-js-tests` check on this PR runs it and is green. I've corrected the earlier comment that said otherwise.

**Non-blocking**: no large-payload (8 MB) parse benchmark added - happy to add one in-PR if you want it, otherwise leaving the O(n) precheck as reasoned rather than measured at scale.

An independent skeptic pass on the install.sh fix noted two pre-existing, non-blocking items for optional follow-up: `mkstemp()` is outside the `try`, so a write-*setup* failure (read-only `~/.cursor`) aborts with a traceback instead of `ERR`; and an existing `hooks.stop` list with no AE entry reports "already current" (by design - it preserves operator config rather than force-injecting). Neither is the reported Medium; flagging for transparency.

tyson-solara6 (approved):
**Automated review** — synthesized from a skeptic (correctness/assumptions), a security-auditor (threat-model pass), a qa-engineer (test-coverage verification), and an architect (design review), all working independently against the live diff.

## Verdict: Approve

CI is green across all 18 checks, and none of the four reviewers found a Critical or Major issue.

### Skeptic (correctness)
Verified the stdin-guard timer/listener lifecycle, the exit-contract preservation on all 7 retrofitted hooks (traced actual hunks, not just the PR description), the `.cursor/install.sh` converge regex anchoring (hand-traced against the `my-stop-context.js`/`nonstop-context.js` false-positive cases), and the Cursor port's TOCTOU fix. No Critical/Major findings. Two Minor, non-blocking:
- Near-identical `run().catch(...).exit(0)` wrapper boilerplate repeated across the codex/gemini/copilot ports (`.codex/hooks/stop-context-codex.js`, `.gemini/hooks/stop-context-gemini.js`, `.copilot/hooks/stop-context-copilot.js`, `.github/hooks/stop-context-copilot.js`) — a shared `runHookMain()` helper could collapse this, but each file has a genuinely different payload shape so the duplication is bounded.
- The "early-completion-by-parse" resolution in `hooks/lib/stdin-guard.js` would silently drop a second value if a harness ever wrote two concatenated top-level JSON payloads on one stdin stream. Current hook contracts are single-object-per-invocation so this is theoretical, but worth a one-line explicit non-goal comment in the module manifest.

### Security-auditor
No Critical/High findings. One Medium: `.cursor/install.sh`'s `tempfile.mkstemp` + `rename` converge path writes the new `hooks.json` at mode `0600` regardless of the original file's permissions — a `hooks.json` that was previously group/world-readable (e.g. shared dev box) silently gets narrowed on every converge run, with no notice to the user. Low exploitability (tightens rather than loosens access) but worth a `os.chmod(tmp_path, os.stat(dst).st_mode)` before the rename in a follow-up. The previously-flagged TOCTOU on the transcript read and the injection surface on the python3 converge script were both checked and are already closed/non-issues in this diff.

### QA (test-coverage)
5/6 specific test claims verified directly against the test files (chunked-delivery no-truncation, per-hook open-silent bounded-exit across all 7 hooks + Cursor, the `deferred_wrap_daemon:true` finalization regression, the Cursor stdout-shape contract, and the install converge/idempotency/anchor regressions). The one claim not verifiable from a static diff review is the specific runtime numbers cited in the PR body (~790-804ms, 965 assertions, 9/9 scenarios) — flagging as unverified-by-this-review rather than accepting at face value, not a blocker. No changed file was found with zero corresponding test coverage.

### Architect (design)
Approve on design. The shared `stdin-guard.js` abstraction is clean and uniformly adopted at all 9 call sites. The 4-port duplication (codex/gemini/copilot/cursor) is pre-existing and unchanged by this PR, not worsened. One non-blocking note: the PR bundles a genuine new feature (the Cursor port, ~300 new lines) together with the actual bug fix (the stdin hang) — reasonable given the shared review history cited in the PR body, but worth knowing for future PRs of this shape (could have shipped as two).

None of the above findings are pre-existing/out-of-scope tech debt — they're all within this diff — so no Jira follow-up tickets were filed for this PR (and note: this run's Jira/mcp-atlassian integration was unavailable in this environment regardless, so ticket-filing was skipped run-wide; see run summary).

---
*Automated review pass — Space-Dinosaurs/DinoStack scheduled PR review.*

tyson-solara6:
Follow-up note on the approve above: I directly re-verified (reading the actual file content at this PR's HEAD via `gh api .../contents`, not just the diff hunks) that both Medium findings from the 2026-07-18 round-2 review are genuinely resolved, not just claimed:
- `stop-context-cursor.js`'s `readTranscriptExcerpt` now opens the transcript path exactly once (`fs.openSync`) and does both the size check (`fstatSync`) and the bounded read (`readSync`) against that same fd — the TOCTOU window is closed, matching the fix the round-2 review itself suggested.
- `.cursor/install.sh`'s converge helper now returns a distinct `ERR` sentinel on every failure path (missing python3, invalid JSON, missing `hooks.stop`, failed atomic write), and both call sites (existing-file and fresh-copy branches) explicitly branch on `ERR` to print a clear "convergence FAILED" message instead of the reassuring "already current" one.

No open blocking findings remain from either prior round.

tyson-solara6 (approved):

Verified:
- `readStdinGuarded` genuinely bounds both time (750ms first-byte, 5s re-armed inactivity, 10s absolute) and size (10 MiB cap); handles empty stdin, non-UTF8/mid-codepoint splits (StringDecoder), and oversized payloads. The `unref()` on every cleanup path is the real anti-hang fix and has a proven pre-fix-failing regression test.
- Applied identically across all 8 JS consumers plus the generated .github mirror; catch-handler differences between harness ports are intentional per each harness's stdout contract, not drift.
- Fail-open holds everywhere: the reader never rejects, every hook exits 0 on every path, and the PreToolUse spawn-emit hook can never deny. `session-end-wrap.js`'s move to `.finally(() => process.exit(0))` correctly lets the awaited read finish before exit.
- Cursor port is universal (no hardcoded paths/identities), resolves the shared guard via `__dirname`, and correctly matches the cross-harness convention of writing context.md only (session_total is Claude-only by design).
- hooks/AGENTS.md counts updated (12->13 hooks, 4->5 Node handlers); DCO clean on all 15 commits; new JS tests are CI-wired via the `hooks/tests/test-*.js` glob.

Minor (non-blocking):
- The install-converge shell tests (`.cursor/tests/install-converge.test.sh`, and the `.claude/tests/` sibling) are not run by any CI workflow - the only shell-test glob covers `hooks/tests/test-*.sh`. The converge logic (anchored regex, atomic write, ERR sentinel) is unit-tested but never CI-gated. Pre-existing gap for this test family; worth a follow-up to wire both dirs in.

PR #459, opened 2026-07-17
fix(install): drop ignored path-scoped Write() permission rules and migrate them out

tristanlee85:
## Summary

- Claude Code's file-permission checks only match `Edit(path)` rules; the two path-scoped `Write(~/.claude/**)` and `Write(~/.claude/projects/**)` allow rules the installer wrote were dead weight that now trigger a startup warning telling users to remove them.
- Every `install.sh` run re-added those rules to `~/.claude/settings.json` (the `missing_allow` merge), so deleting them by hand only lasted until the next `agentic-update` / `pull-and-install`. This is the "why does the warning keep coming back" loop.
- Removed both from `recommended_allow` (the bare `Write` tool rule plus the `Edit(path)` rules already provide the intended coverage) and added a `legacy_allow` migration so existing installs get the stale rules stripped on the next `install.sh` run, in both the already-bypass merge branch and the interactive configure branch.
- Fixed a malformed summary line that read "added removed 2 legacy Write rules rules" in the removal-only case.
- Doc-sync: corrected the allow-list description in `docs/safe-configuration.md`, `README.md`, and `.claude/README.md` (they described the removed `Write(path)` rules).

## North Star alignment

- Pillar(s) advanced (see docs/overview/vision.md): Low friction - removes a recurring startup warning users had to hand-fix repeatedly (the installer re-added the ignored rules every run). Produce verifiable outcomes autonomously - the fix ships with a regression test (case (g)) that runs the real installer end-to-end against a sandboxed HOME.
- Trade-off or pillar this could regress, if any: None. The removed rules never granted anything (Claude Code ignores path-scoped `Write(path)`), so effective permissions are unchanged. Known non-blocking limitation: the migration is scoped to the install's config-dir label, so a redirected-config (`AE_CONFIG_DIR`) install won't strip literal `~/.claude` legacy rules - consistent with how all recommended rules already behave.

## Test plan

- [x] `bash -n .claude/install.sh`
- [x] New regression case (g) in `.claude/tests/install-converge.test.sh`: seeds a sandboxed fake HOME with the legacy rules, runs the real `install.sh`, asserts removal + retention of valid rules + the report line. Confirmed it fails against pre-fix code.
- [x] `bash .claude/tests/install-converge.test.sh` - 16/16 pass
- [x] Two rounds of adversarial Skeptic review - APPROVED, no findings

## Known follow-ups (non-blocking)

- `install-converge.test.sh` (cases a-g) is not wired into any CI workflow - a pre-existing gap.

### DIFF
--- a/.claude/README.md
+++ b/.claude/README.md
@@ -67,7 +67,7 @@ The install script offers to configure `bypassPermissions` mode in `~/.claude/se
 **What it configures:**
 
 - `defaultMode: "bypassPermissions"` - agents can use tools without prompting
-- **Allow list** - `Bash(*)`, the bare `Write` and `Edit` tool rules, and path-scoped `Edit(~/.claude/**)` / `Edit(~/.claude/projects/**)` rules. Only `Edit(path)` rules are matched by Claude Code's file-permission checks - path-scoped `Write(path)` rules are ignored and are migrated out of existing settings
+- **Allow list** - `Bash(*)`, the bare `Write` and `Edit` tool rules, and path-scoped `Edit(~/.claude/**)` / `Edit(~/.claude/projects/**)` rules. Only `Edit(path)` rules are matched by Claude Code's file-permission checks - path-scoped `Write(path)` rules are ignored and are migrated out of existing settings that already have the bare `Write` rule (left in place otherwise)
 - **Deny list** - blocks destructive commands as a safety net:
   - `git push --force`, `rm -rf`, `git reset --hard`, `git clean -f`
   - `sudo rm`, `dd if=`, `shutdown`, `reboot`
--- a/.claude/README.md
+++ b/.claude/README.md
@@ -67,7 +67,7 @@ The install script offers to configure `bypassPermissions` mode in `~/.claude/se
 **What it configures:**
 
 - `defaultMode: "bypassPermissions"` - agents can use tools without prompting
-- **Allow list** - `Bash(*)`, `Write`, `Edit`, and write access to `~/.claude/` directories
+- **Allow list** - `Bash(*)`, the bare `Write` and `Edit` tool rules, and path-scoped `Edit(~/.claude/**)` / `Edit(~/.claude/projects/**)` rules. Only `Edit(path)` rules are matched by Claude Code's file-permission checks - path-scoped `Write(path)` rules are ignored and are migrated out of existing settings
 - **Deny list** - blocks destructive commands as a safety net:
   - `git push --force`, `rm -rf`, `git reset --hard`, `git clean -f`
   - `sudo rm`, `dd if=`, `shutdown`, `reboot`
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ bash .claude/install.sh --dry-run             # preview symlink + repo_dir chang
 Agents need uninterrupted access to Bash, Edit, and Write - constant permission prompts break agent flow and cause subagents to stall. The Claude Code installer offers to configure `bypassPermissions` mode in `~/.claude/settings.json`:
 
 - `defaultMode: "bypassPermissions"` - agents use tools without prompting
-- **Allow list** - `Bash(*)`, the bare `Write` and `Edit` tool rules, and path-scoped `Edit(~/.claude/**)` / `Edit(~/.claude/projects/**)` rules. Only `Edit(path)` rules are matched by Claude Code's file-permission checks - path-scoped `Write(path)` rules are ignored and are migrated out of existing settings
+- **Allow list** - `Bash(*)`, the bare `Write` and `Edit` tool rules, and path-scoped `Edit(~/.claude/**)` / `Edit(~/.claude/projects/**)` rules. Only `Edit(path)` rules are matched by Claude Code's file-permission checks - path-scoped `Write(path)` rules are ignored and are migrated out of existing settings that already have the bare `Write` rule (left in place otherwise)
 - **Deny list** (safety net for destructive commands) - `git push --force`, `rm -rf`, `git reset --hard`, `git clean -f`, `sudo rm`, `dd if=`, `shutdown`, `reboot`
 - **Additional directories** - `~/.claude/projects` for cross-session context
 
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ bash .claude/install.sh --dry-run             # preview symlink + repo_dir chang
 Agents need uninterrupted access to Bash, Edit, and Write - constant permission prompts break agent flow and cause subagents to stall. The Claude Code installer offers to configure `bypassPermissions` mode in `~/.claude/settings.json`:
 
 - `defaultMode: "bypassPermissions"` - agents use tools without prompting
-- **Allow list** - `Bash(*)`, `Write`, `Edit`, and write access to `~/.claude/` directories
+- **Allow list** - `Bash(*)`, the bare `Write` and `Edit` tool rules, and path-scoped `Edit(~/.claude/**)` / `Edit(~/.claude/projects/**)` rules. Only `Edit(path)` rules are matched by Claude Code's file-permission checks - path-scoped `Write(path)` rules are ignored and are migrated out of existing settings
 - **Deny list** (safety net for destructive commands) - `git push --force`, `rm -rf`, `git reset --hard`, `git clean -f`, `sudo rm`, `dd if=`, `shutdown`, `reboot`
 - **Additional directories** - `~/.claude/projects` for cross-session context
 
--- a/.claude/install.sh
+++ b/.claude/install.sh
@@ -1244,6 +1244,13 @@ recommended_deny = [
     "Bash(reboot*)"
 ]
 
+def _migrated_allow(existing_allow):
+    """Merge recommended allow rules in and strip legacy path-scoped
+    Write() rules out. Single-sourced so both the already-bypass and the
+    fresh-configure branches apply the identical migration (regression
+    case (g) in install-converge.test.sh covers this helper for both)."""
+    return list((existing_allow | set(recommended_allow)) - set(legacy_allow))
+
 already_bypass = perms.get("defaultMode") == "bypassPermissions"
 
 if already_bypass:
@@ -1256,7 +1263,7 @@ if already_bypass:
     stale_allow = existing_allow & set(legacy_allow)
 
     if missing_allow or missing_deny or missing_dir or stale_allow:
-        perms["allow"] = list((existing_allow | set(recommended_allow)) - set(legacy_allow))
+        perms["allow"] = _migrated_allow(existing_allow)
         perms["deny"] = list(existing_deny | set(recommended_deny))
         if os.path.islink(settings_path):
             sys.stderr.write(f"refusing to write through symlink: {settings_path}\n")
@@ -1283,7 +1290,7 @@ else:
         existing_allow = set(perms.get("allow", []))
         existing_deny = set(perms.get("deny", []))
 
-        perms["allow"] = list((existing_allow | set(recommended_allow)) - set(legacy_allow))
+        perms["allow"] = _migrated_allow(existing_allow)
         perms["deny"] = list(existing_deny | set(recommended_deny))
         perms["defaultMode"] = "bypassPermissions"
         perms.setdefault("additionalDirectories", [])
--- a/.claude/install.sh
+++ b/.claude/install.sh
@@ -1222,11 +1222,16 @@ perms = settings.get("permissions", {})
 recommended_allow = [
     "Bash(*)",
     "Write",
-    f"Write({_cfg_label}/**)",
     "Edit",
     f"Edit({_cfg_label}/**)",
-    f"Write({_cfg_label}/projects/**)",
     f"Edit({_cfg_label}/projects/**)"
+]
+# Legacy rules from older installs: path-scoped Write() rules are ignored by
+# Claude Code's file-permission checks (only Edit(path) rules match) and
+# trigger a startup warning. Strip them wherever found.
+legacy_allow = [
+    f"Write({_cfg_label}/**)",
+    f"Write({_cfg_label}/projects/**)"
 ]
 recommended_deny = [
     "Bash(git push --force*)",
@@ -1248,9 +1253,10 @@ if already_bypass:
     missing_allow = set(recommended_allow) - existing_allow
     missing_deny = set(recommended_deny) - existing_deny
     missing_dir = f"{_cfg_label}/projects" not in perms.get("additionalDirectories", [])
+    stale_allow = existing_allow & set(legacy_allow)
 
-    if missing_allow or missing_deny or missing_dir:
-        perms["allow"] = list(existing_allow | set(recommended_allow))
+    if missing_allow or missing_deny or missing_dir or stale_allow:
+        perms["allow"] = list((existing_allow | set(recommended_allow)) - set(legacy_allow))
         perms["deny"] = list(existing_deny | set(recommended_deny))
         if os.path.islink(settings_path):
             sys.stderr.write(f"refusing to write through symlink: {settings_path}\n")
@@ -1263,6 +1269,8 @@ if already_bypass:
             added.append(f"{len(missing_allow)} allow")
         if missing_deny:
             added.append(f"{len(missing_deny)} deny")
+        if stale_allow:
+            added.append(f"removed {len(stale_allow)} legacy Write rules")
         print(f"  ~ Permissions: bypassPermissions already set, added {' and '.join(added)} rules")
     else:
         print("  = Permissions already configured (bypassPermissions mode)")
@@ -1275,7 +1283,7 @@ else:
         existing_allow = set(perms.get("allow", []))
         existing_deny = set(perms.get("deny", []))
 
-        perms["allow"] = list(existing_allow | set(recommended_allow))
+        perms["allow"] = list((existing_allow | set(recommended_allow)) - set(legacy_allow))
         perms["deny"] = list(existing_deny | set(recommended_deny))
         perms["defaultMode"] = "bypassPermissions"
         perms.setdefault("additionalDirectories", [])
--- a/.claude/install.sh
+++ b/.claude/install.sh
@@ -1264,14 +1264,14 @@ if already_bypass:
         with open(settings_path, "w") as f:
             json.dump(settings, f, indent=2)
             f.write("\n")
-        added = []
+        parts = []
         if missing_allow:
-            added.append(f"{len(missing_allow)} allow")
+            parts.append(f"added {len(missing_allow)} allow rules")
         if missing_deny:
-            added.append(f"{len(missing_deny)} deny")
+            parts.append(f"added {len(missing_deny)} deny rules")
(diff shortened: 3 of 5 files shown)


tyson-solara6 (changes requested):
## Automated review — request changes

Solid, well-scoped fix for a real recurring papercut (the installer re-adding dead `Write(path)` rules every run). Ran an independent skeptic + security-auditor + architect pass plus a North Star alignment check. One converging, concrete gap needs to be closed before merge; a couple of other points are worth a look but don't block.

### Blocking: fresh-configure branch has no test coverage for the same migration logic

Two independent reviewers (adversarial correctness pass and design/architecture pass) flagged the same gap: the new case (g) in `.claude/tests/install-converge.test.sh` only seeds `defaultMode: "bypassPermissions"` up front, so it exercises exclusively the `already_bypass` branch's
```python
perms["allow"] = list((existing_allow | set(recommended_allow)) - set(legacy_allow))
```
The `else` (fresh/interactive-configure) branch a few lines down runs the identical expression but has zero test coverage seeding a non-bypass settings.json that already contains the legacy `Write(<cfg>/**)` / `Write(<cfg>/projects/**)` rules. A future edit to that branch could silently regress the migration with nothing to catch it.

Suggested fix: add a case (h) mirroring (g) but starting from a settings.json without `bypassPermissions` set (or without a `permissions` block at all) that already has the legacy rules present, and assert they're stripped on the fresh-configure path too. Alternatively, extracting the repeated `(existing_allow | set(recommended_allow)) - set(legacy_allow)` expression into a one-line helper shared by both branches would close this gap by construction and also address a minor DRY nit the architect pass raised.

### Worth a look, not blocking

- **Provenance of the "no-op" claim.** The core premise — "Claude Code only matches `Edit(path)` rules; path-scoped `Write(path)` is ignored" — is stated as fact in a code comment and three docs with no citation. I corroborated it independently against a local Claude Code changelog (v2.1.210, which added the exact startup warning this PR is responding to), so the claim checks out — but consider adding that changelog reference as a citation in the code comment for future readers who won't have that context handy.
- **Literal-string removal has no provenance check** (security-auditor, High). The `legacy_allow` set-difference removes any allow entry that string-matches `Write(<cfg>/**)` / `Write(<cfg>/projects/**)`, regardless of whether the installer or the user added it. It's disclosed via the "removed N legacy Write rules" print line (not silent data loss), and this matches the existing full-list-merge pattern already used elsewhere in the script, but there's no dry-run/backup. Given the rules are confirmed non-functional, the practical risk is low — flagging for awareness, not requesting a redesign.
- **Pre-existing, not introduced by this PR:** `settings.json` is still written via truncate-in-place (`open(path, "w")`) rather than tmp-file+rename, and the new `stale_allow` trigger makes this write path fire on effectively every existing install, increasing how often that race window opens if a concurrent Claude Code session touches the same file. Also pre-existing and self-flagged in the PR body: `install-converge.test.sh` isn't wired into any CI workflow. Both are real but out of scope for this PR — not filing Jira tickets for these since the Jira/mcp-atlassian integration isn't available in this environment this run; noting them here so they don't get lost.
- **`check-vision-alignment` CI check is currently failing** on this PR, but pulling the PR body directly shows the `## North Star alignment` section is present and filled in — this looks like a transient CI issue (possibly a stale read at PR-open time) rather than a real gap. Worth a re-run; that check is explicitly advisory/non-blocking per its own error message either way.

### North Star alignment
ALIGNED — removes a recurring, self-reinstating startup warning (operator-attention win) with no functional permission change and a regression test backing it, consistent with the "low friction" and "verifiable outcomes" pillars.

Nice work on the regression test, the doc-sync across 3 files, and catching the "added removed 2 legacy Write rules rules" grammar bug along the way. Once the fresh-configure branch gets test coverage this looks ready to merge.

tristanlee85:
Addressed in `8f43b8ca` (pushed to this branch).

**Blocking - fresh-configure branch test coverage:** Took the single-source route you endorsed rather than a new case (h). The fresh-configure `else` branch only runs its migration on an interactive `y`, and `tty_input()` reads `/dev/tty` rather than stdin, so the test harness (no controlling terminal, same as CI) cannot actually execute that branch's migration without adding a test-only assume-yes affordance to the installer. Instead the merge-in/strip-out expression is now a single `_migrated_allow(existing_allow)` helper called by both branches, so case (g) covers the exact code both run. That closes the gap by construction and removes the duplicate (the architect-pass DRY nit). One honest caveat: it is a DRY guarantee, holding as long as the `else` branch keeps calling the helper; the case (g) comment documents that.

**Provenance of the no-op claim:** Corroborated independently - the startup warning and `Edit(path)`-only matching landed in Claude Code v2.1.210. Recording that here rather than in the code comment: this repo keeps external tool-version strings out of source (they age and mislead later readers), so the PR discussion is where that provenance lives.

**Literal-string removal (security, High):** Acknowledged, no redesign per your call. The removed rules are confirmed non-functional and the removal is disclosed via the "removed N legacy Write rules" print line, so it is not silent data loss.

**Pre-existing (out of scope, agreed):** the truncate-in-place `settings.json` write and `install-converge.test.sh` not being wired into CI are both real and pre-date this PR. Flagging them as follow-ups rather than expanding scope here.

**`check-vision-alignment`:** stale PR-open-time read. The `## North Star alignment` section is present and filled in the PR body now, and this push re-triggers the check.

tyson-solara6 (changes requested):
Automated review (skeptic + security-auditor pass).

**Verdict: Requesting changes** — one security-relevant logic gap in the new migration helper, otherwise solid.

## Why request-changes

The security audit found a real gap in the new `_migrated_allow()` helper (`.claude/install.sh:1247-1253`). The PR's safety argument is "removing the two `Write(path)` rules is harmless because the bare `Write` rule already covers it" — but that's only true if the user's *existing* config already contains the bare `Write` rule before the migration runs. The `already_bypass` branch computes the stripped set as `(existing_allow | set(recommended_allow)) - set(legacy_allow)` — i.e. it unions in `recommended_allow` (which includes bare `Write`) *before* subtracting the legacy rules. For a user who has a bypass-mode config where they'd removed the bare `Write` rule but left one of the path-scoped `Write(path)` rules in place (a plausible manually-narrowed config), this silently **adds back unrestricted `Write` while removing the only scoped rule they had** — the opposite of what the PR claims ("no functional permission change").

Suggested fix: compute what's safe to strip against the user's *actual pre-migration* state, not the post-union state — e.g. `stale_allow = existing_allow & set(legacy_allow) if "Write" in existing_allow else set()` — so stripping only fires when the bare rule is *already* present, not merely present-after-merge.

This is Medium severity (local trust boundary, requires a specific manually-drifted config to trigger), but it directly undercuts this PR's core safety claim on a security-relevant config file, so I'd rather see it addressed before merge than ship on the assumption.

## Otherwise: solid work

- Exact-string set membership for the migration (not regex/substring) — confirmed safe against unrelated user rules.
- The `_migrated_allow()` refactor in patch 3 correctly de-duplicated logic that was previously copy-pasted across both branches — good catch fixing that DRY issue mid-review.
- Structured JSON parsing throughout, no injection risk, symlink-write guard preserved on both write paths.
- Docs (`README.md`, `.claude/README.md`, `docs/safe-configuration.md`) accurately describe the new behavior.

## Minor (non-blocking)

- `install-converge.test.sh` case (g) doesn't positively assert idempotency by re-running install a second time against already-migrated state. Provably idempotent by construction, but an explicit double-run assertion would be stronger regression insurance for exactly the "warning keeps coming back" bug class this PR fixes.

## Pre-existing issues spotted (not this PR's responsibility — flagging so they don't get lost, since Jira auto-filing was unavailable this run)

1. **Non-atomic `settings.json` write** — both write sites (`:1271-1273`, `:1305-1307`) do `open(path, "w")` + `json.dump()` directly against the live file, no temp-file+rename. A crash mid-write (SIGKILL/OOM/disk-full) could truncate/corrupt a user's `~/.claude/settings.json`. Pre-existing on `main`, worth a hardening ticket (`os.replace` after writing to a `.tmp` sibling).
2. **`missing_dir` computed but never acted on in the `already_bypass` branch** (`:1262-1266`) — unlike the fresh-configure branch, the already-bypass branch never appends the missing entry to `additionalDirectories`, so a user missing that entry will see the same "recurring warning" bug this PR fixes, just for a different field, forever. Pre-existing, good companion-fix candidate.
3. **`install-converge.test.sh` isn't wired into any CI workflow** — self-disclosed in the PR's "Known follow-ups," confirmed against all 12 workflow files. The only regression coverage for a security-relevant permission migration currently runs on nobody's CI.
4. **No shellcheck job exists anywhere in this repo's CI**, despite inline `# shellcheck` pragma comments throughout `install.sh` implying one should. Not blocking for this diff (the new code is pure Python in a heredoc), but worth a repo-wide follow-up if bash static analysis is intended to be part of the safety net.

None of these block this PR — flagging them here since there's currently no ticket tracker wired up to this automated review pass to file them against.

tristanlee85:
Addressed in `379e22d5`.

**Blocking - `_migrated_allow()` union-before-subtract gap:** Fixed. The legacy-rule strip is now gated on the bare `Write` rule already being present in the user's pre-migration `existing_allow`, inside the helper itself. Note the strip does not live in the `stale_allow` variable (that only drives the write-trigger and the "removed N" message) - it lives in `_migrated_allow`, so the gate had to go there or the subtraction would still fire. New regression `case (h)` seeds a bypass config with a scoped `Write(~/.claude/**)` rule but no bare `Write`, and asserts the scoped rule is retained with no removal reported; it fails against the pre-fix helper and passes post-fix.

One clarification on the framing, since it shaped the fix and the docs: bare `Write` is unioned in from `recommended_allow` on every path regardless of the gate, and path-scoped `Write(path)` rules are inert (ignored by Claude Code), so effective permissions are unchanged either way. What the gate buys is conservatism - it stops the installer from rewriting a scoped rule out of a config a user may have deliberately narrowed, without their input. Trade-off, now documented in the docstring and `docs/safe-configuration.md`: the startup warning about the inert scoped rule persists for that bare-Write-absent edge case. The prose no longer claims "no functional permission change"; it states the invariance and the trade-off accurately.

**Minor - idempotency:** `case (g)` now runs install a second time against the migrated state and asserts stable output (`already configured`, no re-reported removal). Suite is 23 passed, 0 failed.

**Pre-existing (out of scope, agreed):** the non-atomic `settings.json` write, `missing_dir` never appended in the already-bypass branch, `install-converge.test.sh` not wired into CI, and the absent shellcheck job - all real, all pre-date this PR. Flagging as follow-ups rather than expanding this PR's scope.

README.md
```diff
@@ -157,7 +157,7 @@ bash .claude/install.sh --dry-run             # preview symlink + repo_dir chang
 Agents need uninterrupted access to Bash, Edit, and Write - constant permission prompts break agent flow and cause subagents to stall. The Claude Code installer offers to configure `bypassPermissions` mode in `~/.claude/settings.json`:
 
 - `defaultMode: "bypassPermissions"` - agents use tools without prompting
-- **Allow list** - `Bash(*)`, `Write`, `Edit`, and write access to `~/.claude/` directories
+- **Allow list** - `Bash(*)`, the bare `Write` and `Edit` tool rules, and path-scoped `Edit(~/.claude/**)` / `Edit(~/.claude/projects/**)` rules. Only `Edit(path)` rules are matched by Claude Code's file-permission checks - path-scoped `Write(path)` rules are ignored and are migrated out of existing settings
```
tyson-solara6:
```suggestion
- **Allow list** - `Bash(*)`, the bare `Write` and `Edit` tool rules, and path-scoped `Edit(~/.claude/**)` / `Edit(~/.claude/projects/**)` rules. Only `Edit(path)` rules are matched by Claude Code's file-permission checks - path-scoped `Write(path)` rules are ignored and are migrated out of existing settings that already have the bare `Write` rule (left in place otherwise)
```
The unconditional "are migrated out" phrasing omits the gate condition install.sh actually applies, and is false in the case-(h) scenario (bare Write absent -> scoped rule retained). docs/safe-configuration.md already documents the gate correctly.

.claude/README.md
```diff
@@ -67,7 +67,7 @@ The install script offers to configure `bypassPermissions` mode in `~/.claude/se
 **What it configures:**
 
 - `defaultMode: "bypassPermissions"` - agents can use tools without prompting
-- **Allow list** - `Bash(*)`, `Write`, `Edit`, and write access to `~/.claude/` directories
+- **Allow list** - `Bash(*)`, the bare `Write` and `Edit` tool rules, and path-scoped `Edit(~/.claude/**)` / `Edit(~/.claude/projects/**)` rules. Only `Edit(path)` rules are matched by Claude Code's file-permission checks - path-scoped `Write(path)` rules are ignored and are migrated out of existing settings
```
tyson-solara6:
```suggestion
- **Allow list** - `Bash(*)`, the bare `Write` and `Edit` tool rules, and path-scoped `Edit(~/.claude/**)` / `Edit(~/.claude/projects/**)` rules. Only `Edit(path)` rules are matched by Claude Code's file-permission checks - path-scoped `Write(path)` rules are ignored and are migrated out of existing settings that already have the bare `Write` rule (left in place otherwise)
```
Same doc-sync gap as README.md:160 - the unconditional phrasing omits the gate condition and is false in the case-(h) scenario.

tyson-solara6 (approved):
## Automated review — approve

Round 3. Both blocking findings from the last two rounds are verified fixed:

- **Round 1 (test coverage gap):** the fresh-configure branch's migration logic previously had zero test coverage. Fixed by extracting the shared logic into a single `_migrated_allow()` helper called by both branches, so case (g) now covers the exact code both run (the fresh-configure branch itself can't be driven from CI - it gates on an interactive `/dev/tty` read - so the DRY-by-construction argument is the correct fix here, not a workaround).
- **Round 2 (security-relevant union-order bug):** `_migrated_allow()` used to union `recommended_allow` in *before* subtracting `legacy_allow`, so a user who'd removed bare `Write` but kept a scoped `Write(path)` rule would have it silently stripped while bare `Write` got silently added back. Fixed by gating the strip on `"Write" in existing_allow` checked against the pre-union parameter. New regression case (h) covers exactly this scenario.

Ran an independent skeptic pass and a security-auditor pass against just these two fixes (not a full re-review — rounds 1-2 already covered broad ground). Both verified the fixes empirically, not just by reading:

- Skeptic ran the real test suite (23/23 pass) plus two targeted mutation tests that reintroduce the exact round-1 and round-2 bug patterns — both are caught by the current tests, confirming the regression coverage is load-bearing rather than vacuous. Sign-off: **APPROVE**, 0 Critical, 0 Major, 2 Minor (below).
- Security-auditor independently traced Python set-mutation semantics (`|` vs `|=`) to confirm the gate genuinely reads pre-union state, and confirmed the human-facing "removed N legacy Write rules" report can't lie about what happened in the previously-vulnerable scenario. Round-2 finding: **RESOLVED**, 0 Critical/High/Medium remaining.

### Minor (non-blocking) — suggested changes attached

1. `README.md:160` and `.claude/README.md:70` both state unconditionally that scoped `Write(path)` rules "are migrated out of existing settings," which omits the `"Write" in existing_allow` gate and is literally false in exactly the case-(h) scenario (bare `Write` absent → scoped rule is retained). `docs/safe-configuration.md` already documents this correctly — these two are just out of sync. Suggested wording attached inline.
2. Case (g)'s new idempotency assertions verify the outer "nothing to do" gate on the second run rather than directly re-exercising `_migrated_allow()` twice — not vacuous (mutation-tested: a broken first-run fix still fails the assertion right after run 1), just structurally indirect. No action needed.

Nice work closing both real gaps with regression tests that were verified — not just claimed — to catch reintroduction of the underlying bug classes.

README.md
tristanlee85: Fixed in `53de0ea6` - applied your suggested wording verbatim. The bullet now states scoped `Write(path)` rules are migrated out only "for existing settings that already have the bare `Write` rule (left in place otherwise)," matching the `"Write" in existing_allow` gate in `_migrated_allow()`.

.claude/README.md
tristanlee85: Fixed in `53de0ea6` - same gate clause applied here so `.claude/README.md` stays in sync with `README.md:160` and the already-correct `docs/safe-configuration.md`.

tyson-solara6 (approved):
Reviewed the full 5-commit diff plus a security audit focused on the permission-migration logic (this touches `.claude/install.sh`'s allow/deny-list configuration, so treating it as security-sensitive).

**Verdict: Approve.**

What I checked:
- `_migrated_allow()`'s gate (`"Write" in existing_allow`) and the `stale_allow` reporting predicate use the identical pre-mutation `existing_allow` set - confirmed they can't diverge (the printed summary always matches what's actually written).
- Bare `"Write"` is unconditionally unioned in on every path, so removing the two path-scoped `Write(<cfg>/**)` rules can never narrow effective permissions regardless of whether Claude Code's matcher truly ignores `Write(path)` as claimed - the scoped rules are a provable subset.
- Test coverage is solid: case (g) covers legacy-strip + idempotency (second run doesn't re-report removal), case (h) covers the retention path when bare `Write` is absent pre-migration. Both branches of the gate are exercised.
- Docs (`docs/safe-configuration.md`, `README.md`, `.claude/README.md`) are consistent with each other and with the actual gated behavior.
- Security audit: no Critical/High/Medium findings. deny-list and the existing symlink-write guard are untouched by this diff.

One non-blocking, cosmetic observation (not asking for a fix, just flagging): in the `already_bypass` branch, when `settings_path` is a symlink the write is correctly skipped, but the summary `print()` still unconditionally reports what *would have* been added/removed - this PR is the first change to route the `stale_allow`-only case through that same pre-existing print, so a legacy-rule-only migration on a symlinked settings file would now print "removed N legacy Write rules" even though nothing was written. Pre-existing pattern, not introduced by this PR's logic, and not a permission-boundary issue (the actual write still correctly no-ops) - just worth a follow-up if it's ever seen in practice.

Nice work tightening the gate logic and adding the idempotency + retention test cases in response to earlier review rounds.

tyson-solara6 (approved):

Verified:
- The migration gate keys off the bare `Write` rule being present pre-migration, consistently in both `_migrated_allow()` and `stale_allow`, so any config written by the old installer gets stripped and the recurring startup warning loop is genuinely broken.
- Idempotent: second run is a clean "already configured" no-op (covered by tests). Scoped `Write(path)` removed from `recommended_allow`, so no re-add regression.
- Read-modify-write set-union preserved (no key clobber); no bare `read -p` introduced.
- Doc-sync complete across README, .claude/README, and docs/safe-configuration.md; no stale scoped-Write references remain elsewhere.
- DCO clean on all 5 commits.

Minor (non-blocking):
1. `.claude/tests/install-converge.test.sh` (including new cases g/h) is not wired into any CI workflow, so the regression protection never runs in CI. Pre-existing gap for the whole install-converge test family, but worth wiring up.
2. Message assembly around `.claude/install.sh:~1303`: singular counts render as "1 allow rules", and the combined case repeats "added". Cosmetic.
3. The fresh-configure branch strips legacy rules via `_migrated_allow()` but its output still says only "Configured ... with recommended allow/deny rules" - it migrates silently, unlike the already-configured branch which reports the removal. Reporting asymmetry only.

PR #468, opened 2026-07-22
docs(protocol): optimize /wrap Step 0 ordering and cache scaffold check

tyson-solara6:
## What

Three speed/token optimizations to the `/wrap` command spec (`content/commands/wrap.md`), plus regenerated adapter artifacts.

1. **Fast zero-substance short-circuit (Step 0-pre).** Detects the unambiguous zero-substance case *before* Step 0's expensive full-file reads (AGENTS.md, learnings.md, compression-state.json) and the `gh pr list` network call, and skips them. Gated hard: any file mutation, uncommitted tracked change, specialist finding, or latent memory/decision forces fall-through. The `git status`/`git stash` work-loss safety net always runs.
2. **Scaffold-check sentinel.** An mtime-keyed signature (`.agentic/wrap/.scaffold-verified`, machine-local) lets the ~15-probe pre-flight scaffold-accuracy check short-circuit on an unchanged clean scaffold. Written only on a zero-work pass, so it never suppresses a drift bullet.
3. **Batched Step 0 checks.** Existence probes + `gh pr list` collapse into one Bash invocation (via a `PROBE_PATHS` array), reusing Step 0-pre's git output instead of re-running it, with a fallback to individual calls on batch failure.

## Review

Worker + fresh independent Skeptic (two rounds). Round 1: 2 Major (memory-loss risk in the short-circuit; a shell-syntax defect in the batch heredoc) + 3 Minor - all fixed. Round 2: sign-off granted, 0 Critical / 0 Major, 1 non-blocking Minor.

## North Star alignment

- **Low friction:** the core win - a steady-state `/wrap` on a zero-substance session drops ~15 redundant scaffold probes, multiple full-file reads, and a `gh pr list` network round-trip.
- **Guard operator attention:** faster wrap turnaround, no new prompts or stops introduced.
- **Verifiable outcomes:** the change also fixes two correctness defects surfaced in review - a shell-syntax error in the batch snippet and a lost-memory-entry hazard in the fast path.
- **Trade-off:** the fast short-circuit adds a judgment gate; it is deliberately biased to fall through to the full path on any doubt, so the failure mode is "did the full, correct thing" rather than "skipped needed work." No pillar regressed.

## Scope

`content/commands/wrap.md` is the only source change; the 8 adapter copies are regenerated build artifacts. `docs/index.html` carries a 2-line auto date-stamp from the pre-commit hook. No `content/sections/` change (no methodology-baseline regen).

### DIFF
--- a/.claude/commands/wrap.md
+++ b/.claude/commands/wrap.md
@@ -22,6 +22,14 @@ Manual `/wrap` is synchronous: there is no in-session auto-enrichment protocol. 
 
 **Pre-flight scaffold-accuracy check** (runs BEFORE Step 0). `/init-project` is the canonical scaffolding spec; /wrap uses it as the reference for "what this project should look like." Check for drift and auto-migrate the critical items inline:
 
+**Sentinel short-circuit (evaluate before item 1).** On a steady-state project this pre-flight repeats the same ~15 filesystem probes on every `/wrap` invocation for no reason. Cache the result:
+
+- **Watched-paths signature.** The signature covers exactly the paths this pre-flight inspects: the CLAUDE.md/AGENTS.md set (root + every track directory), the `.claude/context.md` / `.claude/memory.md` / `.claude/memory/` session-state paths and their `.agentic/` counterparts, the legacy config paths `.claude/{qa,deploy,findings,tracking,learnings}.md` and their `.agentic/` counterparts, the stub targets (`.agentic/tracking.md`, `.agentic/deploy.md`, `.agentic/learnings.md`), the docs dirs (`docs/overview/`, `docs/technical/`, `docs/planning/`, `docs/research/`), `.claude/settings.json`, `.claude/settings.local.json`, and `.gitignore`. For each watched path, record `path:exists:mtime` (mtime as epoch seconds, or the literal string `absent` when the path does not exist); sort the resulting lines lexicographically by path, join with newlines, and take the sha256 of the result — this is the **signature**.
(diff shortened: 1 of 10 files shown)


tyson-solara6:
Automated review pass (own PR — no formal verdict, informational only).

Reviewed the substantive diff at `content/commands/wrap.md` (the canonical source; the 8 adapter mirrors + `.hermes`/`.openclaw` skill files are byte-identical copies, confirmed by the passing `check-adapter-sync`/`check-codex-skill-sync` checks).

Both optimizations hold up on a careful read:
- **Sentinel short-circuit**: the "clean pass only" write-gate is correctly conservative (any migration, stub creation, auto-fix, or unfixable drift suppresses the sentinel write), and the release-signal-gated deploy.md stub check is explicitly carved out of the sentinel's scope with a clear rationale — good catch on the case where a first-release session wouldn't otherwise get the stub created.
- **Step 0-pre fast path**: correctly reuses `git status`/`git stash` output rather than re-running it, and the escape-hatch ("any doubt → fall through to full Step 0") is stated for every ambiguous criterion, including the "discussion-only session with an architectural decision but zero file edits" case that a naive mechanical check would miss.
- **Batched Step 0 probes**: the `PROBE_PATHS` heredoc has a documented fallback to individual calls on batch failure, and the doc itself calls out why a `#`-comment inside the array literal is safe (vs. the backslash-continued list syntax hazard) — shows the round-2 skeptic pass caught real shell-syntax nuance.

No new issues found beyond what the PR description already reports fixed (2 Major + 3 Minor in round 1, 0 Critical/Major in round 2). All CI checks are green, mergeable, no conflicts.

PR #470, opened 2026-07-22
fix(hooks): guard the hooks subsystem against PEP 604 import crash on Python 3.9

tristanlee85:
## Summary

`hooks/enforce-background-spawn.py` crashed at **import time on Python 3.9**, so the `Task`/`Agent` `PreToolUse` hook dumped a traceback on every spawn (non-blocking, but the background-spawn enforcement it provides was silently dead everywhere the hook is wired).

Root cause: line 226 used PEP 604 union syntax (`str | None`) in a return annotation. CPython evaluates annotations eagerly at def-time unless the module carries `from __future__ import annotations` (PEP 563); on 3.9 the `|`-on-types operator does not exist, so the module raised `TypeError` the instant it loaded, before `main()` and before any fail-open guard. Introduced by `f6206cf8` (DS-70) when the team-routing helper `_resolve_role_harness` was added. The sibling `enforce-tier.py` already carries the future-import and documents exactly why.

## Changes

- **Fix (4 files, one line each):** add `from __future__ import annotations` to `enforce-background-spawn.py` and to 3 sibling test helpers that carried the identical unguarded PEP 604 pattern in their own `run_hook()` signatures and crashed the same way when run directly on 3.9 (`test-enforce-no-abdication.py`, `test-enforce-orchestrator-singularity.py`, `test-enforce-tier.py`). No logic changed. Mirrors the existing `enforce-tier.py` pattern; compatible with 3.8/3.9/3.10+ alike.
- **Regression guard (new):** `hooks/tests/test-hooks-pep604-guard.py` statically parses every `hooks/*.py` and `hooks/tests/*.py` with `ast` and flags any eagerly-evaluated PEP 604 union in annotation position when a module lacks the future-import. The check is **version-independent** on purpose: a plain "run the file and check exit code" test passes silently on 3.10+ (CI runs 3.11) and would leave 3.9 users unprotected. A secondary runtime smoke check is included (self-excluded from its own subprocess loop to avoid recursion).

## Verification

- Reproduced the exact `TypeError` pre-fix and confirmed clean load post-fix on Python 3.9.6.
- Full `hooks/tests/test-*.py` suite green on 3.9.6 (6 files, 95 assertions); the new guard passes 23/23.
- Guard fails on the pre-fix tree (independently reproduced in a throwaway worktree at the base commit) and passes on the fixed tree - a genuine, specific regression signal.
- No CI wiring change needed: `.github/workflows/bin-tests.yml`'s `hooks-python-tests` job already globs `hooks/tests/test-*.py` (under Python 3.11), so the guard runs automatically.

## Known limitation (follow-up, not blocking)

The guard covers **annotation-position** unions only. A value-position type alias (`X = int | None`, `X: TypeAlias = int | None`) also crashes at import on 3.9 but slips past, because distinguishing a type-alias union from a legitimate bitwise-or in value position is not reliably decidable statically without false positives. There are **zero current triggers** in the tree; this is a distinct bug class from the one fixed. Tracked as a follow-up.

## Reviewer note

After merge, this machine's (and any operator's) live hook snapshot must be refreshed (`bash .claude/install.sh`) for running sessions to pick up the fix - the wired copy is the deployed snapshot, not the checkout.

Test plan
- [x] Crash reproduced pre-fix (`TypeError: unsupported operand type(s) for |`) on 3.9.6
- [x] Clean load post-fix (exit 0, no traceback)
- [x] New guard fails on reverted tree, passes on fixed tree
- [x] Full hook test suite passes on 3.9.6
- [x] Realistic Agent spawn payload passes through (exit 0, no deny)

## North Star alignment

- Pillar(s) advanced (see docs/overview/vision.md): **Pillar 4 (works for everyone)** - the hooks now load on Python 3.8/3.9/3.10+ instead of only 3.10+; a hook that crashes on 3.9 is exactly the "works only for some setups" regression the vision names, and this restores portability. **Pillar 2 (verifiable outcomes)** - the globally-wired enforcement hook actually runs again (it was dead-on-import on 3.9), and a version-independent CI guard now makes the bug class checkable. **Pillar 1 (guard attention)** - removes the traceback spam emitted on every spawn.
- Trade-off or pillar this could regress, if any: None material. Adds ~247 lines of self-contained test/guard code that runs in existing CI; no runtime behavior change beyond preventing the crash (Skeptic-confirmed the annotation deferral is behaviorally inert here). The guard's known annotation-position-only scope is documented above as a follow-up, not a regression.

## Review rigor

- Brief / Plan path: N/A - single Elevated unit (debugger-diagnosed hook compat fix); no Brief required per AE risk classification.
- Skeptic rounds (tier): 1 round, Tier 2 (fresh independent Skeptic). Verdict: approved.
- Findings summary: 0 Critical, 0 Major, 1 Minor (guard covers annotation-position unions only; value-position type-alias unions uncaught - zero current triggers, deferred as follow-up). Skeptic independently reproduced the pre-fix crash in a throwaway worktree and confirmed the fix is behaviorally inert beyond stopping the crash.

### DIFF
--- a/hooks/enforce-background-spawn.py
+++ b/hooks/enforce-background-spawn.py
@@ -122,6 +122,8 @@ Performance: < 5 ms per call (in-memory JSON parse + optional YAML parse of two
              small config files + optional stat/proc check, no network I/O).
 """
 
+from __future__ import annotations
+
 import json
 import os
 import re
--- a/hooks/tests/test-enforce-no-abdication.py
+++ b/hooks/tests/test-enforce-no-abdication.py
@@ -28,6 +28,8 @@ Test coverage:
   - Smoke: clean payload -> empty stdout
 """
 
+from __future__ import annotations
+
 import json
 import os
 import subprocess
--- a/hooks/tests/test-enforce-orchestrator-singularity.py
+++ b/hooks/tests/test-enforce-orchestrator-singularity.py
@@ -6,6 +6,8 @@ Each case pipes a JSON payload (or malformed string) into the hook via stdin
 and asserts ALLOW (exit 0, no deny output) or DENY (exit 0, deny in stdout).
 """
 
+from __future__ import annotations
+
 import json
 import os
 import subprocess
--- a/hooks/tests/test-enforce-tier.py
+++ b/hooks/tests/test-enforce-tier.py
@@ -6,6 +6,8 @@ Each case pipes a JSON payload (or malformed string) into the hook via stdin
 and asserts ALLOW (exit 0, no deny output) or DENY (exit 0, deny in stdout).
 """
 
+from __future__ import annotations
+
 import json
 import os
 import subprocess
--- a/hooks/tests/test-hooks-pep604-guard.py
+++ b/hooks/tests/test-hooks-pep604-guard.py
@@ -1,0 +1,239 @@
+# Run with: python3 hooks/tests/test-hooks-pep604-guard.py
+"""
+Regression guard for the whole hooks subsystem's PEP 604 / Python 3.9
+import-time crash class.
+
+CPython evaluates function and variable annotations EAGERLY at def-time
+unless the module carries `from __future__ import annotations` (PEP 563).
+PEP 604 union syntax (`X | None`) on a builtin type in annotation position
+raises `TypeError` the instant such a module is imported under Python
+3.8/3.9 - before main(), before any try/except fail-open guard. This bites
+ONLY on 3.8/3.9; 3.10+ interprets `X | None` as a runtime `types.UnionType`
+and never crashes, so a purely dynamic "run the file and check exit code"
+test would pass silently on any 3.10+ CI runner (this project's
+hooks-python-tests CI job runs Python 3.11) and fail to protect 3.9 users.
+
+Scope: this guard covers BOTH top-level hooks/*.py (the hooks themselves,
+e.g. hooks/enforce-background-spawn.py) AND hooks/tests/*.py (the test
+helper files, e.g. hooks/tests/test-enforce-tier.py - these carry their own
+`run_hook(payload, extra_env: dict | None = None)`-style helper signatures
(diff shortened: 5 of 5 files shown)


tyson-solara6 (approved):
**Automated review (skeptic + security-auditor + North Star check, no code changes requested)**

Verified independently:
- **Root cause & fix are sound.** `from __future__ import annotations` defers annotation evaluation to string form (PEP 563) instead of eager def-time evaluation, which is exactly what's needed to stop the PEP 604 `TypeError` on Python 3.9. Security-auditor did a full-file read of `enforce-background-spawn.py` (501 lines) and confirmed nothing in the deny/allow logic depends on runtime annotation introspection (`typing.get_type_hints`, `inspect.signature().annotations`, etc.) — this is a behaviorally inert change beyond fixing the crash.
- **New regression guard (`test-hooks-pep604-guard.py`) is sound and version-independent** — static AST scan over all `hooks/*.py` + `hooks/tests/*.py`, correctly walks nested scopes/generics. It's part of the `hooks-python-tests` CI job, which passed — since the guard does a full-repo scan, that green check is itself proof no other file in `hooks/` still carries the same unguarded pattern.
- **Security:** clean. No new dependencies (stdlib-only: `ast`/`os`/`subprocess`/`sys`), no injection risk in the `subprocess.run()` smoke check (fixed argv list, not shell), and every hook fails open on empty-stdin JSON parse before reaching any effectful code path — so the new smoke-test execution can't trigger side effects.
- **North Star:** ALIGNED — restores a background-spawn enforcement gate that was silently dead on Python 3.9, directly serving the "verifiable outcomes" and "works for everyone" pillars, with zero added operator attention cost.

The self-disclosed known limitation (value-position type aliases like `X = int | None` aren't caught by the guard) is reasonable to defer — zero current triggers in the tree, and the PR is upfront about the gap.

No Critical/Major findings. Approving.

PR #483, opened 2026-07-25
docs(protocol): add ticket-rework alert reference doc (U1)

tyson-solara6:
## Summary

U1 of the ticket-rework Plan (6 Elevated units, no dependencies for this unit). Adds `content/references/ticket-rework.md`, the canonical read-on-trigger reference for the rework-alert feature designed in `docs/planning/ticket-rework/architect-plan.md` (Skeptic-approved, 3 rounds, Open Questions empty).

Covers, per the plan's U1 steps 1-4:
- What the alert is and deliberately is not (no continuation-vs-rework discriminator, with the measured evidence: 0 `Done -> non-Done` Jira transitions across 93 tickets, 0 GitHub Issues across 5 consumer repos, 0 Linear state transitions across 80 issues).
- The ledger schema, including the nullability column and the null-render rule (`n/a`, with `qa_status` preferring its skip rationale).
- `pr_number` as the sole identity key, both branch-name failure modes it avoids, and the distinction that `$BRANCH_NAME` is a lookup key at write time, never an identity.
- The Phase 9 write point and why Phase 12 and Phase 11c were both rejected.
- The append-plus-dedupe-on-read concurrency rationale on the correct grounds - `O_APPEND` offset atomicity for regular files, the NFS caveat, and an explicit "do not cite `PIPE_BUF`" note (that bound governs pipes/FIFOs, not regular files - this was a real review-round correction and is recorded so it isn't reintroduced).
- The dual-branch anchoring pattern, covering both verified instances found during review (Phase 9's `gh pr create` Case A / Case B, and `/ds-ticket-triage` Phase 0's no-args / explicit-input branches) with the general rule: anchor downstream of every branch, never inside one.
- A disclaimer that this notice is command-scoped (fires inside `/ds-implement-ticket` Phase 1) and is NOT a session-start stacked notice, so it should never be folded into `content/rules/conventions.md`'s "4th stacked first-user-turn notice" count.
- Trigger rule, notice template, escalation table (risk floor + Tier-3 at 2+ prior attempts), and known limitations.

Nothing in this file enters an always-loaded path - `scripts/build-methodology.sh` assembles only from `content/sections/`, and this file lives under `content/references/`.

## North Star alignment

Advances **produce verifiable outcomes autonomously**: the reference doc is the artifact that lets a future engineer, Skeptic, or maintainer verify the rework-alert design decisions (ledger schema, identity key, write anchor, concurrency grounds) without re-deriving them from the architect plan or tribal memory. It also advances **low friction**: recording the dual-branch anchoring failure mode explicitly (found twice during review) reduces the chance a future editor repeats it. No pillar is regressed - this is a pure documentation addition with no runtime behavior change, no new always-loaded content, and no altered command behavior.

## Review rigor

- Read the Skeptic-approved architect plan (3 review rounds, 0 open findings) and the Brief in full before writing.
- Verified current repo state directly rather than trusting the plan's line-number citations, which the plan itself flags as stale after PR #481 (`ds-implement-ticket.md` is 2815 lines, not 2807). Grepped for the Case A/Case B labels and the Phase 0 no-args/explicit-input breadcrumbs in both live command files to confirm the structural claims before writing prose around them - no line numbers are cited in the new doc itself.
- Confirmed the 30-reference-doc count the plan asserts (`ls content/references/*.md | wc -l` = 30) before adding the 31st.
- Confirmed `scripts/build-methodology.sh` only reads `content/sections/`, so this file cannot leak into the always-loaded methodology body.
- Followed the module-manifest format used by neighboring reference docs (`task-state-file.md`, `qa-gate.md`) rather than inventing a new header shape.
- Did not run `bash scripts/build-all.sh` (U6's job) and touched no file outside `content/references/ticket-rework.md`.

## Test plan
- [x] `content/references/ticket-rework.md` created, all nine plan-required content sections present.
- [x] Confirmed `scripts/build-methodology.sh` sources only `content/sections/` - this file is untouched by that script.
- [x] `git status --short` on this branch shows only the one new file.
- [x] Committed with `-s` (DCO); `git config user.email` verified as `tyson@solara6.com` before committing.


tyson-solara6:
Skeptic wording fixes pushed (commit 7f245b4f).

**Intentional divergence from the architect plan's literal notice text (Minor 4):** the plan's write-up shows the notice's second clause as "this ticket is being re-worked." That clause asserts the exact rework-vs-continuation discrimination the feature explicitly cannot make (see the doc's own "What this is, and deliberately is not" section). Changed to "prior work on this ticket may need verification" - a conclusion the ledger data actually supports regardless of which case it is. U2 will copy this template verbatim, so the fix here is upstream of where it would otherwise ship. Flagging per the conductor's note that this divergence was reviewed and approved.

Also addressed: the Major (manifest fields + H1 line reframing U2-U5 as planned/not-yet-live), Minor 1 (Phase 11c enumeration trimmed to the one real lost-record path), Minor 2 (Phase 9 justification tightened to the two exits that actually discriminate: Phase 10 CI timeout, Phase 10a fix-loop cap), Minor 3 (O_APPEND claim now states the single-write()-call precondition), and Minor 5 (README.md:240 stale "20 files" count fixed to 31).

Adapter rebuild re-run; diff is limited to the 4 hardlinked reference copies (byte-identical to source) plus the .hermes/SKILL.md embed, 1:1 with the source edit. `scripts/.methodology-baseline.sha256` untouched.

PR #491, opened 2026-07-26
fix(protocol): batch-state concurrency gates read a field no writer sets (DS-96)

tyson-solara6:
## Problem

`.agentic/batch-state.json`'s concurrency gates read `last_updated`. Every writer sets `updated_at`. The field is absent on disk, so the freshness comparison is always false - and because both gates are phrased *"abort/refuse IF within 10 minutes"*, they **fail OPEN**.

Concretely: Contract A step 2 never aborts a foreign session's batch-state write, and Contract C never refuses a concurrent batch, despite the spec stating concurrent batches are unsupported. **The one cross-session guard DinoStack already advertises has never once executed.**

Verified before the fix: the batch-state schema declares `"updated_at"`, `hooks/lib/state-mark.js` sets `tsField: 'updated_at'` for `batch-state.json`, and zero writers set `last_updated` on that file.

## Change

Ruling: **batch-state's canonical timestamp is `updated_at`; the readers were the outlier.** Repoint the readers, do not move the writer - the schema and the merged code already agreed.

The subtlety that made this more than a rename: **several of these sentences govern BOTH state files in one clause.** Unit 1 hit that exact trap at Contract A step 4. So every site was classified individually as batch-only, loop-only, or dual-file, and dual-file sentences now state the per-file mapping explicitly rather than being renamed.

| File | Field |
|---|---|
| `.agentic/loop-state.json` | `last_updated` (per-turn, by `refreshLiveness`) |
| `.agentic/batch-state.json` | `updated_at` |

Also: reader tolerance is now documented (absent timestamp is treated as **stale**, matching today's effective behavior, rather than as fresh), and the stale `Known separate defect (DS-96, out of scope here)` placeholder Unit 1 left is deleted.

## North Star alignment

**Advances "produce verifiable outcomes autonomously":** a documented guard that never fires is worse than no guard, because it manufactures confidence. This makes Contract A and C actually executable.

**Advances "works for everyone":** the failure was silent and universal - every operator running a concurrent batch was unprotected while the spec said otherwise.

**Trade-off, named:** these gates were inert, so turning them on changes behavior for the first time. A concurrent batch that previously proceeded silently will now hit Contract C's refusal. That is the documented intent, but it is a genuine behavioral change rather than a no-op.

## Verification

- New regression test `bin/tests/test_batch_state_timestamp_field.sh` (auto-wired by the `bin/tests/test_*.sh` glob). **Verified failing on pre-fix content: 3 passed, 17 failed.** After the fix: 20 passed, 0 failed. Per this repo's `AGENTS.md`, `build-all.sh` was run *before* the mutation check so the byte-identity assertion could not falsely confirm it.
- A per-site iterating gate over every batch-state timestamp reader: 0 misses.
- Full `bin/tests/test_*.sh` suite passes; `build-all.sh` adapter diff clean.

## Second defect, found by review and fixed here

Turning on a guard that had never fired broke the primary batch-resume path. **Contract A step 2 had no `status` precondition** - unlike Contract C and the N=1 warning, which both gate on `status=active`.

Because `hooks/lib/state-mark.js` gives batch-state `touchTimestampOnTerminal: true`, the terminal mark stamps `updated_at=now` at session exit while leaving the dead session's `session_id`. So: run a 2-ticket batch, exit, relaunch 30 seconds later, Phase 0a-pre correctly classifies the file RESUMABLE and you pick resume - and the first Contract A write **aborts**, printing *"kill the other session, or remove the file."* The other session is dead, and removing the file destroys the batch cursor you just chose to resume. Two spec sections issuing contradictory directives for identical on-disk state. Same failure on the paused path.

The repo already documents this exact hazard for the sibling file and protects `loop-state.json` from it via `touchTimestampOnTerminal: false`. batch-state deliberately does the opposite, which was harmless only while the readers were broken.

**Fix:** Contract A step 2 gains the `status=active` precondition, applied to **both** files, with the reasoning stated in-spec - a file whose status is not `active` has no live owner by definition, so an ownership abort there is meaningless and actively harmful on resume. The writer flag was deliberately **not** touched: fixing the reader rather than the writer is the same principle as the original DS-96 ruling, and that flag is named in Unit 1's manifest as the single place the distinction is encoded.

Scoped precisely to the differing-`session_id` branch; the null/absent legacy force-takeover branch is untouched.

## Verification

- `bin/tests/test_batch_state_timestamp_field.sh`: **30/30 pass**. Verified failing on the pre-fix tree at `db6d3b1f` - **20 passed, 10 failed**, with all 10 new assertions failing and the 20 pre-existing ones still green. Per this repo's `AGENTS.md`, `build-all.sh` ran *before* the mutation check so the byte-identity assertion could not falsely confirm it.
- Per-site iterating staleness gate: 0 misses.
- `build-all.sh` -> 11 adapters, `git diff --exit-code` clean. `build-slides.sh` idempotent on a second run.
- Independent Tier-3 review re-derived **all 13 per-site classifications with zero misclassifications**, confirmed `loop-state.json`'s `last_updated` survived intact across every reader, reproduced the pre-fix failure evidence itself, and confirmed the slide regen is surgical (one file, one `<li>`, no unrelated deck churn).

`qa_skip: docs-only` - prose specs, a regenerated static deck, mechanically regenerated adapter mirrors, and a grep-only bash test. No runtime path beyond that test, which is in the quality gates and was proven to catch the regression.

## Note on the brief's own mechanical gate

The obligation grep I supplied (`grep -i 'batch'` over timestamp lines) **under-caught**: it is line-scoped, so batch-only sentences not containing the literal word "batch" on the same line were invisible - including two Phase 0a-pre decision-table rows. The worker caught this with a full unfiltered `last_updated` sweep, classifying every hit by reading context. Recording it because "lower bound, not authority" was the right framing and the real lower bound needed the unfiltered sweep.

## Process note

The first two commits were pushed before adversarial review - a conductor ordering error. The Critical above is the direct first-order consequence of the change and a pre-push review would have caught it; answering it required only cross-reading `touchTimestampOnTerminal: true` against the resume path. Recorded so the cost of that ordering is visible rather than absorbed.

## Related issues

Fixes DS-96.


tyson-solara6:
Automated review pass (own PR — no formal verdict, informational only, self-review restrictions apply).

**Correctness check:** the diagnosis is sound — `hooks/lib/state-mark.js` sets `tsField: 'updated_at'` for `batch-state.json`, the schema block declares `updated_at`, and every staleness reader in `content/commands/ds-implement-ticket.md` previously keyed on `last_updated` (which is only ever written for `loop-state.json`). Repointing the readers rather than the writer is the right call — it matches the schema and the merged writer, and the "fail open" framing checks out: every gate is phrased "abort/refuse IF within 10 min," so an always-absent field always evaluates false, i.e. never aborts/refuses.

**Consistency check:** verified all 9 mirrored copies (`.claude/`, `.codex/`, `.cursor/`, `.gemini/`, `.github/`, `.hermes/`, `.openclaw/`, `.opencode/`, `content/`) apply the identical field-mapping edit — no drift between mirrors. The two dual-file sentences (Contract A step 2, the batch-state-coexistence bullet) correctly spell out the per-file mapping instead of naming one field for both. `loop-state.json`'s own `last_updated` is untouched, as intended.

**Test coverage:** `bin/tests/test_batch_state_timestamp_field.sh` is a solid prose-invariant regression guard — carries a proper module manifest, ~20 targeted assertions covering every call site touched, and correctly pins that the stale "DS-96 out of scope here" placeholder note is gone.

**Doc-sync:** the follow-up slide-regen commit closes the `check-slides-sync` gap correctly; all CI checks are green.

No Critical/Major/Minor findings. Mergeable as far as this pass can tell — `mergeStateStatus` shows BLOCKED, which per repo convention is the required-review gate (self-review can't satisfy it), not a code issue.

PR #497, opened 2026-07-26
feat(protocol): triage in-flight detection via open-PR evidence (Unit 3)

tyson-solara6:
Makes `/ds-ticket-triage` aware of work already in flight in other sessions, so a conductor doing triage stops handing out kickoff prompts for tickets someone else is mid-PR on.

## What changed

An open PR naming a ticket's key is now a second **source** for the existing `in_progress` flag - not a new category. Triage already had complete reviewed machinery for "badge it, keep it out of kickoff prompts"; its only weakness was firing solely when a human moved the tracker column.

- New `## In-flight code detection` section, sited after both Phase 0 breadcrumbs and before Phase 1 so it runs on either entry path and is not gated behind `TRACKER != none`.
- **One** `gh pr list` call per run, O(1) regardless of ticket count. Zero shell `git` calls.
- Degrades to a total no-op plus one notice line when `gh` is absent, unauthenticated, or has no remote.
- Branch-name enumeration is explicitly **not** evidence (measured: 82 remote-tracking refs vs 76 live heads vs 7 open PRs). Only open PRs count; `headRefName` of an open PR remains a match candidate.
- New `entry.IN_PROGRESS_TRACKER` provenance flag, so the Notes cell composes clauses rather than selecting a fixed string - required because the modal case is "tracker column moved **and** a PR is open", where a fixed-string design prints a false claim.
- Rules 1-4, `num_chains` accounting, and the consume-and-remainder invariant are unchanged and now gate-enforced against alteration.

## Verification

`bin/tests/test_ticket_triage_inflight.sh` - 72 assertions, green under bash **and** zsh with identical transcripts. Pre-existing `test_ticket_rework_triage_badge.sh` 30/0 (exactly one pattern updated, and **tightened** rather than loosened). `build-all.sh` idempotent, zero drift across all 16 CI-derived adapter paths.

The suite pins **mechanism, not presence**. Verified by mutation, each producing a non-zero failed count and exit 1: deleting every writer of `IN_PROGRESS_TRACKER`; inverting the qualifier's trigger condition; corrupting each of the six reachable `(tracker x PR x rework)` worked examples; plus the three prose-invariant gates and the fence-scoped `--repo` check. An earlier revision of this suite was green under all of those - the count grew 35 -> 72 specifically to close that.

Rebased onto `8fdcd119` before push, and confirmed #495's `ds-wrap` content survives intact in the shared generated `.hermes/SKILL.md` - a stale-base regeneration there would have silently reverted it, and `check-adapter-sync` cannot detect content dropped *before* regeneration.

## North Star alignment

**Works for everyone (universality).** No operator identity, workspace, or tracker is baked in. The step makes no `--repo` assumption, resolves the repo from the ambient checkout, and runs identically at `TRACKER=none` - which is the state most consumer repos are in, and the reason it is sited outside Phase 1's tracker gate. A machine with no `gh` auth gets today's artifact plus one notice line, never an error.

**Guard the operator's attention.** The whole point: an in-flight ticket is badged and withheld from kickoff prompts, and the Notes cell names the causing PR so the operator can act without re-investigating. The evidence cap (3 plus `(+N more)`) keeps that from becoming noise.

**Enforcement floors are not lowered.** No risk gate, Skeptic obligation, or CI check is relaxed. The change adds a prose-invariant suite and three build-failing gates; the one pre-existing assertion it touches is tightened.

## Known limitations

- `/ds-implement-ticket A, B, C` gets no in-flight detection - a genuine hole, follow-up unit.
- A pushed branch with no PR yet, and purely local work, are invisible by design.
- A stale open PR, title-match false positives ("revert DS-34 change"), and `DS-4.1` matching key `DS-4` are accepted and documented in-spec.
- 4 non-blocking Minors from review (presence-only pinning of three prose bodies whose behavior stays determined by other pinned text, plus one wording symmetry) are going to a follow-up ticket.

## Review

Three commits, each independently reviewable: implementation, then two fix passes. Tier-3 adversarial review across three rounds (2 Major -> 1 Major -> clean) plus an independent QA pass; both original Majors were state-coverage defects that five rounds of plan review missed, which is recorded in the plan's rev-7 note.


tyson-solara6:
Automated review pass (self-authored PR — informational only, no formal verdict per self-review restrictions).

**Adversarial review (skeptic + security-auditor + product-discovery alignment check, run in parallel):**

- **Skeptic (correctness):** Clean. 0 Critical, 0 Major, 3 Minor. Verified against the actual diff text (not just the PR body's claims): the one-`gh pr list`-call/run design, the full degradation table (gh absent/unauthenticated/no-remote/call-fails all collapse to a silent no-op), the `git branch -r`-is-not-evidence rationale, and the `IN_PROGRESS_TRACKER` Notes-cell composition rule all check out. Independently verified the one edge case that plausibly could have been a real bug — whether `DS-4` boundary-matches inside `DS-40`/`DS-41` — and confirmed the boundary check correctly rejects it. All 8 adapter mirrors are byte-identical to the canonical file (0 divergence). The `test_ticket_rework_triage_badge.sh` change is confirmed to be exactly one hunk, strictly tightened rather than loosened. Minors (non-blocking): (1) the "why remote-ref enumeration isn't evidence" rationale prose isn't itself pinned by a test, only the mechanical exclusion is; (2) most new AC-* assertions are prose presence/absence checks rather than state-coverage checks, appropriate for a spec file but brittle to meaning-preserving paraphrase; (3) the per-consumer-impact-table Global-context input wasn't explicitly marked n/a in the review brief, though no actual information gap resulted since the one downstream consumer (`/ds-implement-ticket` Phase 0a getting no in-flight detection) is already documented in-diff.

- **Security-auditor:** 1 High, 2 Medium, 0 Critical. **High** (recommend addressing before merge): the new `## In-flight code detection` section instructs the conductor to read attacker-controlled free text (`title`, `headRefName` — any GitHub user who can open a PR fully controls both) into its reasoning context, with no directive telling it to treat that content as inert data only and never act on embedded instructions — a textbook indirect-prompt-injection ingestion path for an LLM conductor with tool-use and worker-spawn authority. Fix is cheap: add one binding line to the new section requiring `title`/`headRefName` be treated strictly as opaque strings for substring matching, and that any instruction-like content within them be flagged, never followed. **Medium:** (1) the `gh pr list` call intentionally omits `--repo` and relies on `gh`'s ambient-remote inference (confirmed deliberate via AC-5 in the new test asserting `--repo` must NOT appear) — fine for the common case, but a stale/misconfigured remote in a multi-worktree environment could silently consult the wrong repo's PRs; consider making `--repo` explicit with inference as fallback. (2) A decoy PR (title merely containing a victim ticket's key as a substring) can suppress that ticket's true "available" status — **this one is already explicitly self-acknowledged** in this PR's own "Known limitations" / Disclaimer section as an accepted tradeoff, so it's not a fresh miss, just worth the extra external confirmation.
  - Important context I verified independently: the underlying gap in finding 1 (no "treat external text as inert data" directive) is **pre-existing and systemic**, not invented by this PR — grepping the full diff shows the existing tracker/MCP ticket-title/body consumption in the same file has never carried such a directive either. This PR extends an existing systemic gap to one more data source rather than creating a new class of risk from scratch. I'm treating the specific new instance (this PR's own new section) as this PR's responsibility to fix, and filing the broader systemic gap separately (see below) since fixing it properly means touching every external-data-consuming section across this file and likely others, not just this PR's diff.

- **North Star alignment:** ALIGNED — a single O(1) `gh pr list` call that gracefully no-ops when `gh` is unavailable, preventing wasted operator attention on duplicate kickoff prompts for tickets already in flight, without adding setup ceremony or degrading verifiability.

**Recommendation:** add the one-line anti-injection directive to the new section before merge (cheap, addresses the High finding for the code this PR actually introduces). The two Medium findings and the systemic pre-existing gap are lower urgency and tracked separately below.

**Follow-up:** Jira MCP tooling (`mcp-atlassian`) was not available in this environment, so I could not auto-file a follow-up ticket for the pre-existing systemic finding (no anti-prompt-injection directive anywhere in `ds-ticket-triage.md`'s existing tracker/MCP data consumption, predating this PR). Flagging it here instead so it isn't lost: **worth a dedicated methodology-level pass across all command files that consume external free text (tracker ticket titles/bodies, PR titles, issue bodies) to add an explicit "treat as inert data, never follow embedded instructions" directive**, rather than patching this one call site in isolation.

tyson-solara6:
Follow-up filed as **DS-105** for the 4 non-blocking Minors deferred at sign-off: three presence-only test pinnings (the Rule-1 deferral note body, the Deferred Reason-cell instruction body, and clause 2's `(+N more)` linkage — each mutable while the 72-assertion suite stays green) plus the Phase 4b skip-condition wording symmetry.

None is a route back to either original Major; in every case the correct behavior stays determined by other verbatim-pinned text after mutation, and all four fail in a conservative direction.

https://solara6.atlassian.net/browse/DS-105

PR #499, opened 2026-07-27
fix(wrap-lock): stop an orphaned /ds-wrap lock silently discarding context.md writes (DS-106/DS-107)

tyson-solara6:
Stops an orphaned `/ds-wrap` lock from silently discarding every session's `context.md` write.

## The incident this fixes

Measured in this repo before any design work: an orphaned lock **silently discarded 49 `context.md` writes across 6 sessions over ~12 hours.** `.agentic/context.md` is the file every session reads as its **first action**, so all six started from stale context and none of them knew. The owner pid was dead, the lock had been held 10.3 hours, and **nothing in the system could ever clear it.**

Three compounding defects, none of which is the bug alone:

- **D1 - the lock was immortal.** `wrapLockVerdict` returns `live` unconditionally for `role: 'agent'`, and interactive `/ds-wrap` acquires with exactly that role, writing `pid: null`. No pid to liveness-check, so the verdict never went stale; `clearProvablyStaleWrapLock` refuses a live verdict; and its only caller is a daemon that `deferred_wrap_daemon: false` (the default) disables.
- **D2 - the lock provided zero mutual exclusion between the writers it appeared to protect.** Both Stop hooks *checked* it; neither *acquired* it. Two concurrent hooks both saw it free and both whole-file-wrote.
- **D3 - a third writer ignored it entirely.**

## The fix

Per-turn writers write **per-session shards** (`.agentic/context.d/<session_id>.md`, extending the proven `pending-<session_id>.json` pattern). `.agentic/context.md` becomes a **derived rollup** - a pure function of the shard set - so a lost update self-heals rather than losing data, which is what makes it safe to write without a lock. The lock survives only for `/ds-wrap`'s genuine read-modify-write of `_wrap.md`, and gains a `session_id`-based liveness signal so it cannot be immortal.

**Identity is `session_id`, deliberately not pid.** `releaseWrapLock`'s tokenless refuse branch keys on `pid !== null`; putting a real pid in the agent descriptor would arm it, and `/ds-wrap` releases without a token - so it would leak a lock on **every run**, converting an occasional orphan into a guaranteed one.

**Bounded degradation:** a stuck lock now surfaces a `WRAP-LOCK-STUCK` banner in `context.md` within one turn, carried by a write path that is structurally lock-independent. There is no longer a state where writes are discarded *and* nothing says so.

Also: the migration seeds `_wrap.md` from the pre-existing curated region with lines 1-2 byte-exact (anything else and the next `/ds-wrap` discards it via its second-line check, destroying the 10-slot rolling window); `bin/agentic-status` gained a `wrap lock:` line, since no operator-invocable command could previously reveal the wedge that had to be diagnosed by hand.

## Verification

**The headline, reproduced then cleared:** the exact 5h-orphan shape, default config, zero daemon involvement -> `cleared-abandoned-lock` + `acquired`, exit 0. And a live lock with a fresh heartbeat is **not** stolen (`busy`, exit 5, descriptor md5 identical). Both halves matter - a fix that clears orphans by stealing everything is worse than the bug.

Independent QA passed all 11 scenarios, including a hand-constructed race whose retry counter reached `attempts: 2`, proving compare-and-retry fired; migration against 6 adversarial file shapes including the **real live `context.md`** (lines 1-2 byte-exact, every curated section preserved, `CAPTURE-GAP` correctly stripped); and mutation-testing that confirmed forcing `heartbeatFresh()` false makes the helper wrongly steal a live lock - so the guard is real, not incidental.

Adversarial review across two rounds. Round 1 found a **Critical**: three writes discarded their return values, so one fs failure on the `_wrap.md` seed with the rollup write succeeding destroyed the curated region **permanently** - the marker guard suppresses any retry. Fail-open inverted into anti-heal, on the one file class this ticket exists to protect. Fixed as three checked returns and verified by injecting real EACCES on each publishing rename; the reviewer re-injected independently on all four paths and confirmed `written: false, attempts: 0`, `context.md` byte-intact, full self-heal next turn.

Round 1 also found the `WRAP-LOCK-STUCK` guarantee had **zero automated coverage** - its own falsifying mutation passed the entire suite. Now 9 assertions on the real default path with both-direction controls; both mutations killed.

15/15 AC mutations killed. Gate: 27 assertions, exit 0 under bash 3.2.57 and zsh 5.9, **20 FAIL against the pre-implementation tree** - every assertion is bidirectional. hooks 39/40 suites, pytest 491, `build-all.sh` clean, methodology baseline untouched.

## North Star alignment

**Works for everyone (universality).** `session_id` validation is deliberately permissive rather than fail-loud, because `makeLockDescriptor` is called inside `acquireWrapLock`'s try whose catch removes the lock dir - a throwing validation would make `/ds-wrap` refuse to run on every adapter where the env var is unset. Harnesses without a session id fall back to an age-based rule: degraded, never stealing a live lock. The OpenCode plugin is ported in lockstep, since `check-adapter-sync` structurally cannot see it.

**Guard the operator's attention.** A stuck lock is now loud and bounded to one turn instead of silent and indefinite, and `agentic-status` reports lock state - including wedges from a plain file or symlink at the lock path, which previously read as `free`.

**Enforcement floors are not lowered.** The never-steal invariant is narrowed, not removed: verified across Table A rows A1-A6, D1-D4, L1-L3, with the only live-lock steal being the plan-enumerated >30-minute single-turn case, a documented deferred default.

## Known limitations

- Five passive adapters (`.cursor`, `.copilot`, `.github`, plus `.codex`/`.gemini` which write harness-local paths and never contend) retain one turn of ordering damage until ported; content is recovered as `_foreign.md`.
- `.agentic/context.d/` grows one shard per session with no pruning yet.
- The OpenCode parity gate is **structural, not behavioural** - it cannot execute closure-scoped functions without an OpenCode runtime, and says so in its own header rather than implying equivalence.
- Two Minors accepted: a non-atomic `Bun.write` on the TS seed path, and one comment that overstates ("saved EVERY byte" - `stripDerivedBlocks` deliberately drops derived notices, which is what AC22 requires).


tyson-solara6:
**Automated review pass (self-authored PR — informational only, no formal verdict per review policy).**

This is a large, well-documented change (per the PR body: two rounds of adversarial Skeptic review, mutation testing, 15/15 AC mutations killed, independent QA). One item worth a look before merge:

## CodeQL: high-severity insecure-randomness alert (introduced by this PR)

`hooks/lib/wrap-marker.js` — new `writeShard()` function:

```ts
const tmp = `${target}.tmp.${process.pid}.${Math.random().toString(36).slice(2, 10)}`;
```

CodeQL flags `Math.random()` here as `js/insecure-randomness` (CWE-338, high severity — see [alert #114](https://github.com/Space-Dinosaurs/DinoStack/security/code-scanning/114)). In this specific use it's not a secrets/token scenario, but it *is* used to build a temp-file path that's later `rename()`d into place without an exclusive-create check — a predictable temp name (PID + low-entropy `Math.random`) opens a theoretical TOCTOU/symlink-race window if another local process can pre-create a file at the guessed path before the rename.

Worth noting: this same file's manifest already documents a "prefer `crypto.randomUUID`, fall back to `Math.random` only on old runtimes" pattern for other identifiers (see the comment ~line 5406 in this diff) — this new call doesn't follow that existing convention. Suggest swapping to `crypto.randomUUID()` (or `crypto.randomBytes`) for the temp-file suffix to close the CodeQL finding and stay consistent with the rest of the module.

Everything else (the lock-liveness fix, per-session shard writes, migration byte-exactness checks) looks consistent with the PR description and I did not spot any other introduced security-sensitive patterns (`eval`/`exec`/shell-injection surface) in the diff — the two `child_process` usages in the diff are both in test files invoking the local test harness with fixed paths, not user input.

Not blocking (informational only, per self-review policy) — flagging so it can be fixed before merge.

PR #516, opened 2026-07-28
feat(hooks): fire-log telemetry for six enforce-*.py hooks

tyson-solara6:
## Summary

Six of the seven `enforce-*.py` hooks left no trace of whether they ever fire, making an inert rule indistinguishable from a load-bearing one. Adds a shared `hooks/lib/enforcement_log.py` that appends one JSONL line to `.agentic/.enforcement-fires.jsonl` on every non-passthrough action (deny, or allow-with-advisory). Silent allows never call it.

Motivated by `docs/planning/opus5-pruning-assessment-2026-07-28.md`: the abdication guard reports 0 fires, and that currently cannot be distinguished from "never instrumented".

Already producing real data: `.agentic/.enforcement-fires.jsonl` contains genuine fires from `enforce-planning-artifact-spawn` (allow_advisory) and `enforce-shippable-edit` (deny).

## Review

Two adversarial rounds. Round 1 withheld sign-off on a Critical: the log call preceded the decision `print()`, so a raising `log_fire` propagated into each hook's outer fail-open handler and **silently converted a deny into an allow** - demonstrated by execution on four hooks, all returning rc=0 with empty stdout. Fixed by printing the decision first, then lazily loading and calling the logger inside its own guard. Six regression tests (one per hook) install a raising stub and assert the decision still reaches stdout; all six fail behaviorally against the pre-fix tree.

Round 2: 0 Critical, 0 Major, sign-off granted.

## North Star alignment

- **Verifiable outcomes autonomously.** Hook fire rates become measurable, so a rule's necessity can be evidenced rather than assumed. Directly serves the "curate, do not accrete" imperative: an enforcement floor that never fires is now a visible deletion candidate instead of a guess.
- **Guard operator attention.** Fires only on non-passthrough actions, so the file stays small and silent allows cost nothing. The loader is lazy, so kill-switched invocations skip the import entirely.
- **Universality.** No operator identity, path, or tooling is baked in; the log lands under the existing `.agentic/` convention and is gitignored. Fail-open at every layer, verified against missing, syntactically invalid, raise-at-import, and raise-at-call lib states.

Trade-off: a ~20-line loader bootstrap is duplicated six times. It is the bootstrap that loads the shared lib, so it cannot itself be shared; accepted by review with that reasoning.

## Verification

- 24 `hooks/tests/test-*.py` pass on 3.13; on the 3.9 floor the same 2 failures as `origin/main` (pre-existing, reproduced on both trees - round 1's count of 1 was an undercount).
- `scripts/build-all.sh` idempotent; every removed line in the cumulative diff audited - no unrelated content dropped, `.hermes/SKILL.md` purely additive.
- `python3 scripts/codex-skills.py check --repo .` reports OK (4 skills).
- Decision stdout byte-identical to `origin/main` across a deny/allow payload set for all six hooks.
- `enforce-no-abdication.py` and `.agentic/.abdication-guard-fire-count` untouched.

## Follow-ups (non-blocking, from round-2 Minors)

- `hooks/lib/enforcement_log.py:140` inline comment still cites PIPE_BUF; the docstring was corrected but the comment was missed.
- `os.write` return value unchecked (no short-write loop); hardening only.
- Deterministic `spawn_complete` telemetry (G1) remains deferred - `PostToolUse` fires at `async_launched`, so wall-time and token data do not exist in the payload. `SubagentStop` plus `bin/agentic-parse-subagent-usage` is the likely mechanism; needs a live payload capture. Currently recorded only here and in the commit message; should become a ticket.


tyson-solara6:
Automated review pass: this is our own PR, so no formal verdict (self-review not applicable). Note: GitHub reports `mergeStateStatus: DIRTY` / `mergeable: CONFLICTING` against `main` - needs a rebase/merge-conflict resolution before it can be merged. No other action taken this run.

tyson-solara6:
Automated review pass (our own PR, so no formal verdict): checks are green and the design/adversarial-review writeup in the description is solid. One operational note - GitHub reports this branch as `CONFLICTING`/`DIRTY` against `main` right now, so it'll need a rebase/merge before it can go in. North Star check: ALIGNED (makes previously-silent hook enforcement measurable without adding operator attention cost).

PR #519, opened 2026-07-29
fix(hooks): mechanically enforce the ballot ban on prose Operator decisions blocks

tyson-solara6:
## Summary

METHODOLOGY.md forbids the conductor from presenting a co-equal multiple-choice ballot when a best option is derivable, and states the ban "applies identically to prose asks." It was enforced only on the `AskUserQuestion` tool path.

In the session that produced this, the conductor violated the ban **five consecutive turns** and nothing caught it:

- `enforce-askuserquestion-default.py` is registered `matcher: "AskUserQuestion"` and only inspects that tool call. The conductor never called the tool - it wrote prose.
- `enforce-no-abdication.py` scanned only for `_PERMISSION_PHRASES` ("let me know if you'd like"). The ballots used none of those, and `_NEGATIVE_GATE_PATTERNS` actively *suppresses* the guard on irreversibility vocabulary - which those ballots were saturated with ("reverses decision #203", "one commit to revert").
- PR #513 had just introduced a blessed prose surface, the `## Operator decisions` heading, with no enforcement.

Net: the ban was enforced on the tool path and unenforced on the prose path. This closes that.

## What lands

**A detector** in the existing Stop hook: an `## Operator decisions` block with 2+ items where 2+ carry no recommendation marker blocks the turn and injects a corrective directive. Fence-masked before heading search, column-0 item anchoring, no paragraph fallback.

**A corpus rule change**, which is the larger half of the value: `content/sections/02-delegation.md` now mandates the `(Recommended)`/`Recommendation:` marker on in-block items - the same token the tool path already requires. This converts an unfalsifiable style expectation into something an enforcement can check, and unifies both paths under one convention.

## Honest coverage

The detector recognizes three item forms (`- `, `1. `, `**1.`) under one literal heading. Adversarial review tested 17 realistic ballot shapes: **2 fire, 15 escape** - bold-without-number, tables, sub-headings, indented and blockquoted bullets, `(1)` numbering, unicode bullets, reworded headings. The always-loaded text says so explicitly - *"a floor, not a guarantee of coverage"* - mirroring the self-limiting clause `enforce-tier.py` already carries. Residual evasions are enumerated in-code at each pattern's definition. A guard that silently misses most cases while the docs claim enforcement would be worse than no guard, because the absence of a block reads as compliance.

## Review

Three adversarial rounds, 0 Critical throughout.

- **Round 1 (5 Major):** a spec-compliant block fired the guard; indented sub-bullets counted as items; the heading matched inside unmasked fences; the paragraph fallback was untested and over-fired on "None this turn." narrative; **the flagship regression test was vacuous** - mutating away the exact fix it named left it passing.
- **Round 2 (2 Major):** the spec fix re-opened round 1's finding in reverse - lines 76/78 still gave a marker-free canonical form labeled "Format is MANDATORY" while line 80 mandated the marker, so a block in the file's own prescribed format was blocked. And the enforcement claim was unqualified while catching 2 of 17 shapes.
- **Round 3:** sign-off granted, 2 Minor residuals.

Verified by execution at each round: mutation-testing every fix (both new tests fail behaviorally when their pinned behavior is reverted), no catastrophic backtracking (worst 6.7 ms across 10 adversarial inputs; 54 ms on a synthetic 2.8 MB/200k-bullet input, linear), stdout byte-identical to `origin/main` across 8 legacy dispositions, kill switches short-circuit before any new work, no double-fire with the tool-path hook, `CONSECUTIVE_BLOCK_CAP = 2` bounding any false positive.

## North Star alignment

- **Guard operator attention.** A co-equal ballot offloads the conductor's own job onto the operator. This is the failure the rule exists to prevent, now caught rather than described.
- **Verifiable outcomes autonomously.** The marker mandate makes compliance machine-checkable instead of a matter of taste. The detector's coverage limits are stated in the same always-loaded text that states the rule, so the guard cannot be mistaken for full enforcement.
- **Universality.** Hooks are Claude Code only; the other 10 harnesses rely on the prose, and the amended text says so ("Other adapters rely on this prose rule"). The rule change benefits every harness; the detector is a Claude Code floor on top.

## Verification

- Full hook suite passes on the default interpreter and the Python 3.9 floor (2 pre-existing 3.9 failures unchanged, reproduced on `origin/main`).
- `scripts/build-all.sh` idempotent across 11 adapters; `.methodology-baseline.sha256` regenerated; `codex-skills.py check` OK (4 skills); four `.claude/skills/agentic-engineering/*` symlinks relative.
- `.hermes/SKILL.md` diff is exactly 6 removed lines, each mapping 1:1 to an intentionally-edited paragraph - no unrelated content lost.

## Follow-ups (non-blocking, round-3 Minors)

- `docs/configuration-reference.md:80` and `docs/safe-configuration.md:95-99` still describe only the interrogative detector. Incomplete rather than false, and outside the mandated doc-sync set.
- `content/sections/02-delegation.md:80` states the marker requirement unconditionally, then scopes it to 2+ items. Zero runtime surface - the detector cannot fire on one item under either reading - but the wording invites a negative inference.


tyson-solara6:
Automated review pass (our own PR, so no formal verdict - self-review not applicable).

Quick skim confirms this is a mature, well-tested diff: the substantive logic (`hooks/enforce-no-abdication.py` `_is_prose_ballot` classifier + `hooks/tests/test-enforce-no-abdication.py`) has regression tests pinned to each prior Skeptic-round finding, and every known evasion gap is documented inline as an accepted limitation rather than an oversight. No new Critical/Major concerns beyond the two doc-sync gaps already disclosed in the PR body's "Follow-ups" section.

One actionable item: GitHub currently reports this PR as `mergeStateStatus: DIRTY` / `mergeable: CONFLICTING` against `main`. It needs a rebase/merge-conflict resolution before it can land, independent of anything else.

PR #522, opened 2026-07-30
docs(agents): pin adapter-drift checks to the CI-enforced pathspec

tyson-solara6:
Recovers a signed-off commit that was stranded unpushed on a local `main` checkout.

## What happened

Commit `27e4d5bc` was authored directly on a local `main` branch by a concurrent session and never pushed. It was discovered when `git pull --ff-only` refused to fast-forward after #520 merged: local `main` held one commit that `origin/main` did not.

Rather than reset the checkout and destroy it, the commit was preserved on this branch. `git reset --hard` is permission-blocked in this repo, which is what saved it - the destructive path was unavailable and the recovery path was taken instead.

## The change

One line in `AGENTS.md`: any local adapter-drift check must copy its pathspec verbatim from the `Fail on adapter drift` step in `.github/workflows/adapter-sync.yml` - the 11 adapter dirs plus `.github/copilot-instructions.md` and `.github/prompts/`.

That is a real trap worth documenting. The Copilot build writes outside `.copilot/`, so a hand-written local drift check that enumerates only the adapter directories silently misses two paths that CI does check - a green local check followed by a red `check-adapter-sync`. This session hit adjacent versions of that problem more than once.

## Provenance

Authored and signed off by `Tyson Hummel <tyson@solara6.com>`; the commit is unmodified from its original form. Nothing was rewritten, amended, or squashed in recovery - only pushed to a branch and opened as a PR.

## North Star alignment

- **Works for everyone (universality).** Points contributors at the CI workflow as the single source of truth for the adapter pathspec instead of a prose list that drifts. Any operator writing a local pre-flight check gets the same answer CI enforces.
- **Produce verifiable outcomes autonomously.** Removes a false-confidence failure mode: a local check that passes while CI fails is worse than no local check.
- No pillar regressed; no write to `docs/overview/vision.md` or `requirements.md`.


tyson-solara6:
Superseded by #523 (`6a6c28df`), which landed the identical `AGENTS.md` change on `main`.

This branch existed as insurance: the commit was authored on a local `main` checkout and unpushed, and was discovered when `git pull --ff-only` refused to fast-forward. It was preserved on a branch rather than risking a `reset --hard`. The originating session has since shipped it directly, so there is nothing left to merge here - verified `git diff origin/main..` is empty after updating the base.

Closing rather than merging an empty diff. Deleting the branch.

PR #565, opened 2026-08-04
docs(memory): archive 23 non-load-bearing entries, restore MEMORY.md to its stated budget

tyson-solara6:
Restores root `MEMORY.md` to the budget stated in its own header.

## What this does

`MEMORY.md` was 123 lines / 3,120 words against its own stated cap of ~120 lines / ~2,000 words. This moves 23 non-load-bearing entries out, landing at **77 lines / 1,968 words**.

The file's own rule governs the move: entries go **verbatim** to `MEMORY-archive.md` rather than being deleted, and where the archive already holds a copy the archived (usually fuller) version wins and the `MEMORY.md` copy is simply dropped.

- **14 removals were de-duplications** against a longer copy already in `MEMORY-archive.md` — no append, no shorter variant introduced.
- **9 were fresh verbatim appends** under a new `## Archived 2026-08-03` heading.

## Verification

Checked mechanically rather than asserted, because "moved, not deleted" is exactly the claim worth distrusting:

- Entry-level accounting: 50 entries → 27. All 23 removals confirmed recoverable from `MEMORY-archive.md`.
- **All 27 surviving entries are byte-identical to `HEAD`** — nothing silently reworded, merged, or "improved".
- All 10 sections preserved, none emptied. Management header untouched. Zero entries added (pure move).

Deliberately preserved as still load-bearing:

- Both 2026-08-02 entries from the session that just closed (worktree-isolation-vs-`Read` harness mechanics; verify a cross-repo handoff doc's premise before scoping from it).
- The resident-budget entry that says *run the script, do not trust a cited baseline/THRESHOLD pair* — it landed hours earlier specifically to stop figures going stale.
- All three entries carrying the 15-required-checks facts (ruleset `14778332`, `strict_required_status_checks_policy: true`, and the "full 15-check re-run" line), corrected this session from a stale 14.

Close calls kept rather than archived: the DS-74 `TRACKER=none` entry and the extraction-staleness entry (both are explicitly referred back to by newer entries), the worktree-branch-leak entry (it corrupted `main` once), the Elevated-signal duplication entry, and the "MEMORY.md is public" entry.

## North Star alignment

**Guard operator attention.** `MEMORY.md` is always-loaded via `@MEMORY.md`, so every word costs context in every session on this repo. A file 58% over its own word cap spends that budget on entries whose lessons are now encoded in CI gates or superseded by newer entries. Archiving keeps the durable record intact while restoring the always-loaded tier to what a session tomorrow actually needs.

**Works for everyone.** Nothing is destroyed — `MEMORY-archive.md` retains every entry verbatim and stays in the repo, so a teammate looking for historical rationale still finds it.


tyson-solara6:
Closing as superseded by #555 (DS-129), which untracked `MEMORY.md` and deleted `MEMORY-archive.md` outright.

This PR archived 23 entries from `MEMORY.md` into `MEMORY-archive.md`. #555 consolidated both files (99 entries -> 58) and retired the archive pattern entirely - `MEMORY.md` is now gitignored and local-only, and `MEMORY-archive.md` no longer exists. Neither file this PR touches is tracked any more, which is why the rebase produced delete/modify conflicts.

No content lost: the surviving entries live in the now-local `MEMORY.md`, and the two `AGENTS.md` conventions from the same session landed separately in #552.

PR #572, opened 2026-08-05
DS-135: remove O(n^2) body-join from adapter builds, add progress output and a static tripwire

tyson-solara6:
## Summary

`.gemini/build.sh` took **13.9 minutes** on this machine. 96.8% of that was one file, and the cause was a quadratic string-accumulation loop.

The hot loop rebuilt the entire accumulated body string on every iteration:

```bash
body_content="$body_content
$bline"
```

For `content/commands/ds-implement-ticket.md` (3341 lines, 335 KB) that dominated everything else: 805.84s of an 832.99s profiled run, against ~27s for all 24 other commands, 18 agents, GEMINI.md, references, and scaffolding combined. `ds-init-project.md` is 3.2x smaller by bytes but ran **87.8x** faster, so the real cost is worse than plain quadratic on macOS `/bin/bash` 3.2.57.

Replaced with a single O(n) join. `.codex/build.sh` carried the byte-identical loop and is fixed the same way.

**This fires on every local commit, not just CI.** `hooks/pre-commit` runs `.gemini/build.sh` unconditionally whenever `content/` is staged. Three separate sessions misread the silence as a hang: two killed it (one at ~3 min, one SIGKILL at ~10 min), and a third yielded its turn mid-build and had to be explicitly resumed. The script was correct the whole time.

## What changed

- **`.gemini/build.sh`, `.codex/build.sh`** — quadratic accumulate-in-loop replaced with a guarded `IFS=$'\n'` / `"${body_lines[*]}"` join.
- **Progress output in both** — the silence is what caused the misdiagnosis, so each build now prints a line per file. Verified nothing parses either script's stdout.
- **`bin/tests/test_no_quadratic_body_join.sh`** (new) — static tripwire asserting the pattern does not return.
- **`.github/workflows/adapter-sync.yml`** — existence guard before the drift check.

### The guard is load-bearing, not cosmetic

```bash
if [[ ${#body_lines[@]} -gt 0 ]]; then ... else body_content=""; fi
```

Both scripts run `set -euo pipefail`, and on bash 3.2 an empty array expanded as `"${a[*]}"` under `set -u` is **fatal**:

```
$ /bin/bash -c 'set -euo pipefail; a=(); IFS=$'\n'; echo "[${a[*]}]"'
bash: a[*]: unbound variable
```

`.codex/build.sh` is where this is reachable — line 370 already used the empty-safe idiom, which is the existing code telling us that array can be empty. Dropping the guard would break only macOS contributors' local commits and never CI.

`$(printf '%s\n' ...)` was rejected: command substitution strips trailing newlines and breaks byte-identity for bodies ending in blank lines.

### Why a static tripwire rather than a timing test

The first draft proposed asserting wall-clock time. Adversarial review killed it: `bin/tests/test_*.sh` is globbed into a job with `timeout-minutes: 5` and **3 to 63 seconds** of measured headroom, while `.codex/build.sh` alone takes ~31s. It would have killed the job with no diagnostic and taken every other shell test with it. Its 120s threshold was also unfalsifiable — the job cap always trips first — and it would have added a third live-checkout-mutating test in the same PR series as DS-136, which exists to remove exactly that.

The grep costs milliseconds, cannot time out, mutates nothing, and is demonstrably falsifiable: it matches both scripts pre-fix and zero-matches post-fix.

### The adapter-sync guard

`git diff --exit-code` exits 0 silently on a pathspec that does not exist, so an omitted or misspelled path green-lights itself. All 16 entries exist today; the guard fails loudly if one stops existing.

Note the 16-vs-11 "disagreement" DS-138 reported is **not** a defect: the 16 is the 11 adapter directories plus the 5 `.github/*` outputs `.copilot/build.sh` writes outside its own directory. Two lists counting different things, both deliberate.

## Test plan

- [x] **Byte-identity across every generated file** — sha256 manifest of `.gemini/commands` + `.codex/agents` before and after; `diff` empty, `git status --porcelain` empty. This is the gate that matters: `adapter-sync.yml` runs `git diff --exit-code` across every adapter directory.
- [x] Tripwire demonstrated failing on the reintroduced pattern, passing on the fix
- [x] Existence guard verified in both the all-present and induced-missing cases
- [x] Full `bin/tests/test_*.sh` sweep — one failure, `test_codex_hooks_resolve.sh`, which is untouched here, byte-identical to `main`, and fails only because macOS bash 3.2 lacks `mapfile`. **DS-136 fixes exactly that.**
- [x] No em dashes; DCO signed with matching raw author email

## Known and accepted

The `.gemini/build.sh` timing above was measured under CPU contention from a concurrent agent running the same script, so the 13.9-minute figure is worse than a quiet machine would show. The independently profiled baseline was 832.99s. No post-fix timing is quoted as a speedup multiple because the measurement was contended — byte-identity, not wall-clock, is what this PR asserts.

The tripwire is exact-string-matched, so a variable rename would evade it. Accepted for a zero-cost, zero-flake guard; it is now itself one more small hand-maintained artifact, which is the standing shape of this system (see the DS-138 closure note).

## North Star alignment

- **Pillar(s) advanced (see docs/overview/vision.md):** **1 (guard operator attention)** — a mandatory pre-commit step that goes silent for 8+ minutes has cost three sessions directly, including two that killed a correct process and one that abandoned a required CI assertion. **2 (produce verifiable outcomes autonomously)** — the drift-check guard converts a check that could silently pass into one that fails loudly.

- **Trade-off or pillar this could regress, if any:** the byte-identity requirement means any error in the join corrupts all 25 generated TOML files at once; mitigated by the sha256 proof over every generated file rather than a sample. The progress output adds stdout noise to a previously quiet script — verified no consumer parses it. The existence guard adds a third hand-maintained copy of the pathspec, which narrows the failure it guards without eliminating the class; a shared job-level `env:` value was recommended as a follow-up rather than bundled here.

---

## Correction after review (round 1)

**The O(n^2) join was not the dominant cost.** The claim above that 96.8% of runtime was the join loop came from a profile that timed *per file* and then attributed that time to the join by reasoning rather than sub-measurement. The join and the escaping are adjacent lines inside the same timed region.

The real cost was the escaping on the next line, `${body_content//\\/\\\\}`. Measured in isolation on `/bin/bash` 3.2.57 against the 335 KB file: **still running after 200 seconds**. The join fix was correct and is retained, but it left the dominant cost untouched, so the developer-visible symptom persisted on macOS.

Fixed in `d8db2a8b` by moving both escape passes out of bash parameter substitution into a single `python3` pass.

**Measured, real numbers:**

- Before: killed at **12:01.52**, still stuck on `ds-implement-ticket.toml` - never observed completing.
- After: **8.5s total**.

Byte-identity re-proven across the full 43-file manifest after the change.

Also fixed: the tripwire passed vacuously from any non-repo-root cwd and when a build script was missing - the exact "silently-skipped assertion is indistinguishable from a passing one" failure this repo's conventions warn about. It now resolves repo root and hard-fails on an unreadable target.


tyson-solara6:
Automated review note (self-authored PR, no formal verdict per self-review rules):

One test-coverage gap worth flagging before merge: the PR's own "Correction after review (round 1)" note says the real hang-causing bottleneck was the bash-native `${body_content//\\/\\\\}`-style escape passes (not the join loop itself), fixed by routing both escapes through `python3`. But the new regression test (`bin/tests/test_no_quadratic_body_join.sh`) only greps for the old join-loop pattern (`body_content="$body_content...`) - it doesn't guard against someone reverting the escape passes back to the slow bash-native form, which is the change that actually fixed the reported multi-minute hangs. Since this was already misdiagnosed once, a second static tripwire grepping for the old `${body_content//\\/\\\\}` / `${body_escaped//` pattern would close that gap. Not blocking - suggest adding it as a quick follow-up before or shortly after merge.

No correctness bugs found in the escape ordering or the empty-array `IFS` guard; CI is green.

tyson-solara6:

- The O(n) join replacement is byte-identical to the old quadratic loop, including the empty-array and embedded-empty-line cases — verified.
- Flag for visibility: the accompanying escape-step rewrite (piping through `python3` via `$(...)`) introduces a trailing-newline-stripping risk the old pure-bash approach never had. It's mitigated correctly with a `\x01` sentinel, but this wasn't mentioned in the PR summary — worth calling out explicitly as a second behavioral change bundled under this title, even though the fix itself looks sound.
- The new "static tripwire" (`bin/tests/test_no_quadratic_body_join.sh`) is a real but narrow regression guard — it greps for the literal `body_content="$body_content` string, so it'll catch a verbatim reintroduction but not a semantically-equivalent quadratic pattern under a different variable name or formatting.

tyson-solara6:
Automated review pass (self-authored PR, no formal review verdict since GitHub blocks self-review). Adversarial read found the O(n^2)->O(n) join and the python3 escape-pass rewrite to be behavior-preserving (traced both by hand, including the `set -euo pipefail` interaction with the escape subshell). Two things worth noting:

**Pre-existing, out-of-scope bug spotted in the touched code path** (would normally file a Jira ticket for this, but the mcp-atlassian tools aren't available in this environment - flagging here instead): the `while IFS= read -r line; do ... done < "$src"` loops in `.gemini/build.sh` (~line 150) and `.codex/build.sh` (~line 280) silently drop the source file's last line whenever it lacks a trailing newline, because `read -r` returns non-zero on that final partial read and the loop exits before the body runs for it. This predates this PR but sits directly in the code this PR modifies. Fix: `while IFS= read -r line || [[ -n "$line" ]]; do ... done`.

**Nit (introduced by this PR):** the new ~15-line python3 escape block is now byte-identical and duplicated between `.gemini/build.sh:239-249` and `.codex/build.sh:398-408` - the same shape of duplication this PR is fighting. Might be worth a small shared `scripts/toml-escape.py` (matching the existing `scripts/codex-skills.py` pattern) as a fast follow.

tyson-solara6:
Automated review pass (self-authored PR, no formal review verdict since GitHub blocks self-review). Round-2 check on the 2 new commits since the last check-in.

**Confirmed correct:** the O(n^2)->O(n) join fix (940fa3d) is behavior-preserving - `IFS=$'\n'; body_content="${body_lines[*]}"` reproduces the original per-line join exactly, and the `${#body_lines[@]} -gt 0` guard correctly avoids an "unbound variable" crash on macOS's bash 3.2 for empty arrays under `set -u`. The new tripwire test (`bin/tests/test_no_quadratic_body_join.sh`) does catch reintroduction of the exact removed pattern and no longer passes vacuously if the target file is unreadable. The sentinel-byte TOML-escape fix (d8db2a8) correctly strips only the injected sentinel, not real trailing content, and preserves backslash-before-`"""` escaping order.

**Still open (out-of-scope, correctly left alone):** the pre-existing trailing-newline-drop bug in `.gemini/build.sh` (~line 150) / `.codex/build.sh` (~line 280) - `while IFS= read -r line; do ... done < "$src"` silently drops a final line lacking a trailing newline - remains untouched, as expected since it's outside this PR's scope. (Would normally file a Jira ticket for this since it's out-of-scope and pre-existing, but the mcp-atlassian tools aren't available in this environment - flagging here instead, same as noted in the prior round.)

**Still open, nit:** the new ~14-line python3 escape block is still byte-identical between `.codex/build.sh:51-61` and `.gemini/build.sh:124-134` - no shared `scripts/toml-escape.py` was extracted, so the duplication this PR is fighting persists (and grew slightly with more duplicated comment lines).

**New finding, worth-fixing:** commit d8db2a8 reworks the escape path but the logic still only escapes backslash and `"""` - it doesn't escape TOML-required control characters (0x00-0x1F except tab/newline, and 0x7F) per the TOML basic-string spec. Low practical risk for markdown source content, but a stray control byte would produce invalid TOML with no diagnostic.

North Star alignment: ALIGNED - fixes a build script that was silently hanging for minutes, restoring a fast, verifiable commit-time step.

PR #579, opened 2026-08-06
fix(hooks): correct post-merge stale text and close the ballot-classifier corpus gap

tyson-solara6:
Corrects two files that describe PR #519's ballot classifier as an unmerged candidate on a branch. It merged as `a950bdd4` three days ago; the text now describes live shipped code in third person. Also closes the corpus gap that should have gated that merge.

## What was wrong

The identical stale paragraph sat in `hooks/enforce-no-abdication.py` and `hooks/tests/test-corpus-abdication.py`, wrong in three ways:

1. **Referred to `feat/prose-ballot-guard (PR #519)` as an alternative design that "exists at" a branch** - that design *is* the file. Now `shipped in a950bdd4 (PR #519)`; the branch name is gone, since branches get pruned and SHAs do not.
2. **Claimed it "does not satisfy the two requirements above" - it satisfies one.** `_BALLOT_REASON` is a dedicated constant whose directive is "revise the Operator decisions block", not the unconditional "proceed now" of `_ABDICATION_REASON`. The requirement was met; the text said otherwise.
3. **Framed two known limitations as unresolved pre-merge tension**, which reads as "should this ship at all" about code that shipped. They are now stated as deliberate, documented limitations of the shipped design: the negative-gate exemption (self-documented in `_is_prose_ballot`'s own docstring, a reasoned choice) and the absent `>`-blockquote heading handling.

## Corpus gap closed

The classifier went live with **zero** of the corpus's 34 rows containing an `## Operator decisions` heading or a recommendation marker, so no row exercised its code path. The governing plan required two such rows before shipping. Adds Group 8, each row **measured against the real hook before pinning**, not predicted:

| row | shape | measured |
|---|---|---|
| g8-00 | genuine 2-item ballot, no markers | BLOCK |
| g8-01 | 2-item block, both marked | ALLOW |
| g8-02 | single marked item | ALLOW |

All three matched prediction; none had to be pinned against expectation. Corpus is now 37 rows.

Note the classifier was never untested - `hooks/tests/test-enforce-no-abdication.py` already carries a thorough ballot suite. The gap was specific to the artifact the plan named as the floor.

## Evidence

Gate proven able to fail before shipping: flipping `g8-00`'s expectation produced `FAIL g8-00 ... expect_block=False actual=True` and exit 1; restored, exit 0, 37/37.

`test-corpus-abdication.py`, `test-enforce-no-abdication.py`, `test-enforce-stall-detection.py` all exit 0. `check-resident-budget.sh` OK at 121043 B / 121066 B, unchanged (no `content/**` touched). Diff is exactly 2 files, both under `hooks/`. DCO verified against raw `%ae`.

## Sweep result (no changes made)

Because `content/sections/02-delegation.md:82` now mandates a `(Recommended)`/`Recommendation:` marker on every item, an old-style marker-less block would block at runtime. Swept `content/**`, `docs/**`, `README.md`, `CONTRIBUTING.md`, `.claude/**` for examples teaching that shape: **none found.** The only example block is `content/references/conductor-turn-format.md:57-58`, which is single-item and so never trips the check (it requires 2+ unmarked items). Every other hit is prose describing the rule. No follow-up needed.

## North Star alignment

- **Verifiable outcomes over asserted ones** - a live guard gains a regression floor in the artifact designated as its floor, with each row's expectation measured rather than assumed, and the suite demonstrated able to fail.
- **Guard operator attention** - removes text that would send the next reader chasing a merged PR as if it were an open candidate, and stops a false claim that a met requirement is unmet.
- **Works for everyone** - stdlib only, no identity, path, tracker, or platform assumption; hooks are outside the resident set so no session pays for this.
- **Low friction / proportionality** - named explicitly, because a prior review found this workstream regressed it. Both fixes land in one PR touching only the two files already at issue; the doc sweep is report-only precisely to avoid pulling `content/**` and its budget cost into scope. Headroom is currently 23 B, which is its own reason not to widen this.
- **No enforcement floor removed** - zero behavioral change to the hook; the docstring edit cannot execute and the corpus only adds rows.

### DIFF
--- a/hooks/enforce-no-abdication.py
+++ b/hooks/enforce-no-abdication.py
@@ -210,22 +210,22 @@ Regression corpus: hooks/tests/test-corpus-abdication.py is the permanent
                   built, never reached a measurable shape. Hard-gate/surface-
                   and-proceed suppression and code-fence/blockquote exclusion
                   must be present from first authoring, not bolted on after
-                  review. None of the corpus's 34 rows contain an
-                  "## Operator decisions" heading or a "(recommended)"/
-                  "Recommendation:" marker, so none exercise a ballot
-                  classifier's path, leaving Groups 3/5/7 as the floor for
-                  the OTHER rejected attempts only; a revival must add a
-                  compliant ALLOW row and a genuine co-equal-ballot BLOCK
-                  row, and must route through a dedicated reason constant
-                  that does not carry an unconditional "proceed now"
-                  directive. A per-item, fence-masking design with a
-                  dedicated non-"proceed now" reason exists at
-                  feat/prose-ballot-guard (PR #519) - per its PR body, 0
-                  Critical findings across three Skeptic rounds, and it
-                  passes this corpus - but it deliberately exempts its
-                  ballot check from the negative gate and does not handle
-                  ">"-blockquote headings, so it does not itself satisfy the
-                  two requirements above; that tension is unresolved.
+                  review. A revival needed to add a compliant ALLOW row
+                  and a genuine co-equal-ballot BLOCK row to this corpus,
+                  and to route through a dedicated reason constant that
+                  does not carry an unconditional "proceed now" directive.
+                  A per-item, fence-masking design with a dedicated
+                  non-"proceed now" reason (_is_prose_ballot /
+                  _BALLOT_REASON) shipped in a950bdd4 (PR #519) - per its
+                  PR body, 0 Critical findings across three Skeptic
+                  rounds, and it passes this corpus. Groups 3/5/7 remain
+                  the floor for the OTHER rejected attempts; Group 8 below
+                  now carries the compliant ALLOW row and the genuine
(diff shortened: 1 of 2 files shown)


tyson-solara6:

- The corpus additions (g8-00/01/02) meaningfully close the coverage gap — they exercise the actual "## Operator decisions" + marker-detection path that had 0/34 rows hitting it before, with distinct block/allow/edge-case outcomes across two marker syntaxes. Not cosmetic.
- The doc corrections (a950bdd4 / PR #519 references) are internally consistent between the two files; no factual errors found.
- The three CI failures on this PR (check-codex-skill-sync, DCO check, check-no-planning-docs) all trace to "Service Unavailable" / GitHub Actions infra timeouts during workflow setup, before any PR-specific step ran — these look environmental, not caused by this diff. Worth a re-run rather than a code fix.

PR #581, opened 2026-08-06
test(harness): knowledge-commit fixtures and harness support (unit 1)

tyson-solara6:
Test-harness groundwork for the knowledge-commit feature. **No `content/` changes, no methodology prose** - this unit builds the fixtures and harness support that a follow-up Phase 11e unit will need, and answers four questions empirically that three rounds of plan review could only speculate about.

## Why this landed first

The Phase 11e plan was blocked three times. Round 3's findings were **entirely in the test specifications, none in the shell logic** - reviewers were arguing about a harness nobody had run. Building it first replaced that speculation with executed evidence, and immediately surfaced three defects prose review had no way to find.

## What it found

- **`render()` raises on any block with zero `PLACEHOLDER_WHITELIST` keys.** The whitelist is global, not per-marker, and fail-closed pre-render. "Model your tests on `test_phase8_telemetry_shell.py`" walks straight into this at all 18 of its `render()` call sites. Documented the bypass rather than making the whitelist per-marker, which would have churned Phase 8 for no gain. A third consumer (`test_qa_knowledge_capture_shell.py`) had already hit the same wall independently.
- **A no-op push short-circuits client-side** (`Everything up-to-date`, exit 0) and never contacts the `pre-receive` hook. The planned push-failure test would have passed vacuously.
- **`git diff-index --quiet HEAD` is not a reliable unchanged-check.** A byte-identical rewrite has an mtime past the index, and outside git's racily-clean window git trusts stat data over content: `identical` ~70% of runs, `modified +0 -0` ~30%. The fixture now forces the stat-dirty state deterministically and the block must `git update-index -q --refresh` first. Phase 11e's per-file gating uses exactly this comparison, so the harness will catch it there.

## What it adds

- Bare-origin support, knowledge-file seeding (4 per-file modes: differs / identical / absent-from-tip / fewer-lines), `DINOSTACK_KNOWLEDGE_GITIGNORE`, and 4 `build_knowledge_*` builders.
- `with_shell_assignments()` - injects a value as a **non-exported** shell assignment, reproducing production where `$BRANCH_NAME` is substituted rather than exported. Without it, an `awk ENVIRON[...]` defect passes its test while failing in production.
- A tee-and-delegate `git` stub that records argv then execs real git, with per-subcommand forced failure.
- 30 tests, `\x1e`-framed argv records with per-record argc validation, double-install guard.

Existing builders untouched; `PLACEHOLDER_WHITELIST` and `render()` behavior unchanged.

## Verification

- Full suite **917 passed**. Floors verified live: smoke 30, phase8 18, qa 52.
- **15 mutations, all confirmed RED.** One (M15) came back green on first attempt - the mutation was weak, not the assertion; re-targeted and confirmed red.
- 10 consecutive smoke-suite runs, no flake.
- Skeptic review: 0 Critical, 1 Major, 3 Minor - all resolved. The Major was argv-log framing: unescaped newlines split one invocation into three records, and Phase 11e's own multi-line DCO commit message is exactly that shape.

## North Star alignment

- **Verifiable outcomes autonomously** - advances it directly. Replaces three rounds of prose speculation about harness behavior with executed evidence, and makes a methodology shell block mechanically verifiable rather than review-only.
- **Works for everyone** - neutral. Test-only code, no operator-facing surface, no identity/tracker/harness assumptions. Fixtures are hermetic (`GIT_CONFIG_NOSYSTEM`, fixture-local `GIT_CONFIG_GLOBAL`, `LC_ALL=C`, no network, everything under `tmp_path`).
- No pillar regressed.

---

## CI timeout hardening (added after a runner-starvation incident on this PR)

`python-bin-tests` was cancelled on this branch at exactly 10 minutes. It was **not** a hang in this diff: the job recorded `steps: []`, so no step ever executed. The account was in runner starvation (23 runs queued, 1 executing), and `10:00` is `timeout-minutes: 5` plus GitHub's 5-minute cancellation grace, consumed entirely by runner acquisition. The suite runs green in ~56s under CI's exact invocation.

Two mechanisms are worth knowing, and the workflow now carries a comment explaining them so nobody "fixes" the next queue cancellation by raising a job timeout alone:

- A job's `started_at` is stamped at **runner assignment**, not first-step execution, so job-level `timeout-minutes` can be fully consumed by queue wait. Only **step-level** timeouts measure real work.
- `.steps | length == 0` on a cancelled job is categorical proof no step ran (cancelled jobs normally retain step records), which cleanly separates infra failure from a code hang.

This was already biting the repo independently of this PR: `bin-sh-tests` ran 4m56s against its own 5-minute job timeout here, and was **already cancelled once on `main`** (run `31115638748`).

Changes: job timeouts 5 -> 20 as backstops, step-level timeouts on every work step, `pytest-timeout` (`--timeout=60 --timeout-method=thread`) so any future stall becomes one red test with a stack trace instead of a killed job with no log, `PYTHONUNBUFFERED: 1`, `timeout=30` + `stdin=subprocess.DEVNULL` on all 9 `subprocess.run` sites (AST-audited, not grepped), and no-prompt git env vars in `_base_env`.

**Verified by making a timeout fire, in three arms with a control** - a green suite proves nothing about a timeout that never triggers. With a `sleep 999` git on the fixture PATH: **(A)** all bounds present -> terminated 60.6s with `TimeoutExpired ... timed out after 30 seconds`; **(B)** harness timeouts stripped, pytest-timeout only -> terminated 60.2s via the plugin; **(C)** control, both stripped -> **hung past 90s**. Arm C is what proves A and B terminated because of the bounds rather than because the sleeping git never took effect.

### DIFF
--- a/.github/workflows/bin-tests.yml
+++ b/.github/workflows/bin-tests.yml
@@ -6,23 +6,45 @@ on:
       - main
 permissions:
   contents: read
+# TIMEOUT POLICY (all three jobs below).
+#
+# A job's `started_at` is stamped at RUNNER ASSIGNMENT, not at first-step
+# execution, so a job-level `timeout-minutes` can be consumed entirely by queue
+# wait. That is not hypothetical here: python-bin-tests was cancelled on this
+# repo having recorded `steps: []` - no step ever ran - and bin-sh-tests was
+# cancelled the same way on main (run 31115638748, sha f81b63c0). The suites
+# themselves finish in well under a minute.
+#
+# So: job-level `timeout-minutes` is a BACKSTOP ONLY (20), and every real work
+# step carries its own `timeout-minutes`, which starts when the step starts and
+# is therefore immune to acquisition delay. Do not "fix" a queue-starvation
+# cancellation by raising the job timeout alone - that trades a fast silent
+# failure for a slow one and leaves the steps unbounded.
 jobs:
   python-bin-tests:
     runs-on: ubuntu-latest
-    timeout-minutes: 5
+    timeout-minutes: 20
+    env:
+      # Keep partial output if the job is force-killed anyway.
+      PYTHONUNBUFFERED: '1'
     steps:
       - uses: actions/checkout@v7
       - uses: actions/setup-python@v7
         with:
           python-version: '3.11'
-      - run: pip install pytest pyyaml
+      # pytest-timeout converts an unbounded subprocess stall anywhere in
+      # bin/tests/ into ONE failing test with a stack trace at the exact call,
+      # instead of a killed job with no usable log.
+      - run: pip install pytest pytest-timeout pyyaml
       - name: install zsh (required for the bash/zsh parity matrix in test_phase8_telemetry_shell.py)
         run: sudo apt-get update && sudo apt-get install -y zsh
-      - run: python3 -m pytest bin/tests/ -q
+      - run: python3 -m pytest bin/tests/ -q --timeout=60 --timeout-method=thread
+        timeout-minutes: 5
       - name: floor collected test_phase8_telemetry_shell.py count at 18
         # 9 test functions x 2 shells (bash, zsh) = 18 collected IDs. Adding or
         # removing a test function in bin/tests/test_phase8_telemetry_shell.py
         # obliges updating the floor below in the same PR.
+        timeout-minutes: 2
         run: |
           python3 -m pytest bin/tests/test_phase8_telemetry_shell.py --collect-only -q | grep -c '::test_' | awk '{ if ($1 < 18) { print "expected >= 18 collected tests (9 functions x 2 shells), got " $1; exit 1 } }'
       - name: floor collected test_qa_knowledge_capture_shell.py count at 52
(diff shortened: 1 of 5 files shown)


tyson-solara6:

- The `PLACEHOLDER_WHITELIST` bypass (global, not per-marker) is a real gap but reasonably bounded: `render()` still fails closed with a clear error pointing at the workaround pattern, rather than silently mis-rendering. Worth tracking as tech debt for whoever implements the actual knowledge-commit block, not a blocker here.
- Both vacuous-pass traps this unit found (no-op push short-circuiting client-side; `git diff-index --quiet HEAD` unreliable for byte-identical rewrites) are correctly worked around in the actual test fixtures (throwaway-ref push target; forced stat-dirty mtimes + `update-index --refresh` + a dedicated regression test).
- Minor: the hardcoded throwaway ref name (`ae-fixture-probe`) has no collision guard across parallel test runs — low risk given per-test tmp_path/bare-repo isolation, but worth a comment if parallel CI runs ever become a source of flakiness.

PR #582, opened 2026-08-06
DS-128: atomic write discipline for /ds-wrap + ownership corrections

tyson-solara6:
## Summary

Closes DS-128. Two gaps, both prerequisites for any future concurrent-`/ds-wrap` work (the concurrency design itself is explicitly out of scope).

**Atomic write discipline.** `/ds-wrap` promised tmp+rename for its small sentinel artifacts but wrote every large human-facing document with a plain overwrite, so a crash or a concurrent reader mid-write could observe truncated or half-merged content. Adds one normative mechanism to `wrap-context-format.md` with two publish variants, cited at **17 write sites** - 13 unconditional, 4 conditional where the site's own prose gates on the target file's absence.

**Ownership correction.** `ds-wrap.md` claimed in two places that `/ds-wrap` writes root `MEMORY.md`, `decisions.md`, and `.agentic/learnings.md` inline. It does not: Part B writes `.agentic/memory.md` (gitignored scratch); `decisions.md` belongs to `wrap-ticket`; `.agentic/learnings.md` to `learnings-agent`/`learning-extractor`. The sole exception is a bootstrap stub-seed that never overwrites. The same false claim was live in `conventions.md`, three slide-deck locations, and `docs/index.html`.

The ticket's own Gap 2 premise was itself false, which is why the deliverable became "correct the claim and record the real writer map" rather than "assign ownership."

## What is in each commit

| Commit | Unit |
|---|---|
| `48944430` | Ownership corrections at `ds-wrap.md:9`/`:702`, write-ordering resolution, `conventions.md:50` (-13 B), deck sweep |
| `8cbe8cbb` | Deck + doc-site accuracy, slide HTML regen, two stale enumerations replaced |
| `8941284c` | NORMATIVE section + 17 marked citation sites, golden fixture regen, Codex inventory refresh |
| `7782b06e` | `hooks/tests/test-wrap-write-discipline.js` |
| `46789890` | Four Skeptic-flagged prose fixes |

## Notable decisions

**The write-site list is mechanically derived, not hand-inspected.** Hand-inspection found 4, then 11. A grep-then-classify pass over every write verb in both files found **17**, and independently surfaced a second false ownership claim nobody had noticed. Every grep hit is accounted for - in-scope with a variant, or scoped-out with a stated reason.

**Write-ordering is resolved by documenting the absence of a guarantee.** `learning-extractor` is fire-and-forget and `wrap-ticket` reads "whatever entries exist ... may be partial or empty." That is a documented, accepted race, not an ordering guarantee - and it sits outside `/ds-wrap` entirely.

**The regression test asserts identity, not proximity.** Three earlier designs tried to prove "this citation belongs to this site" via text windows; each failed differently on real file shapes (overlapping radii; a partition shorter than its own clause; trailing `Result:` anchors putting citations in the previous site's window). Each fix relocated the defect rather than closing it. The test now asserts an exact-substring occurrence count on a per-site marker - no windows, offsets, or anchors. Verified by reverting each of the 19 assertions in turn and confirming exactly one reddens; and the test fails 22 of 27 assertions against pre-fix `main`.

## Test plan

- [x] `hooks-js-tests`: 43 files, 0 failures - incl. `test-wrap-md-partA-parity.js` 19/19 and `test-wrap-context-format-golden.js` 20/20 unmodified
- [x] `check-resident-budget`: OK, 121030 B / 121066 B, **36 B headroom** (`conventions.md` edit is byte-negative by design)
- [x] `check-adapter-sync`: `build-all.sh` + `git diff --exit-code` over the verbatim workflow pathspec, clean
- [x] `codex-skill-sync`: `OK (4 skills)` after inventory refresh (not cleared by `build-all.sh`)
- [x] `check-slides-sync`: slide HTML regenerated, not hand-edited
- [x] All 17 markers verified at exactly one *occurrence* each (not line count - two share a line)
- [x] Test proven able to fail: 22/27 red on pre-fix `main`; all 17 markers mutation-tested individually

## Known limitation

A marker typed next to text that does not implement the mechanism is not catchable by a textual test. Disclosed in the plan rather than papered over; no purely textual assertion can close it.

## North Star alignment

**Verifiable outcomes.** Replaces prose that silently drifted with a CI-visible pin: 17 write sites and 2 ownership claims now fail a required check if reverted. The test's ability to fail was demonstrated by mutation, not asserted.

**Works for everyone.** No operator identity, workspace, tracker, or local path is baked in. The mechanism is prose executed by whichever harness runs `/ds-wrap`; the marker is a plain HTML comment that survives adapter regeneration to all 11 directories, verified.

**Truthful intent layer.** Removes a false claim that had already propagated to a second file, three decks, and the docs site - the class of defect that generates phantom tickets. The corrected claims are narrower and carry their own exception explicitly.


tyson-solara6:
Automated review note (self-authored PR, no formal verdict since self-review isn't applicable):

A skeptic pass surfaced one Major issue worth fixing before merge:

**Lock-coverage safety claim doesn't hold for all atomic-write sites.** `content/references/wrap-context-format.md`'s NORMATIVE section justifies the non-atomic "conditional publish" TOCTOU pattern (`if [ -e <target> ]; then rm ...; else mv ...; fi`) by claiming "`/ds-wrap`'s writes are already serialized by `.agentic/wrap/lock`; concurrency beyond that lock is out of scope." But per `content/commands/ds-wrap.md`, the "Pre-flight scaffold-accuracy check" (line 17, runs *before* Step 0) performs several of the 17 write sites - `claude-md-root`, `memory-md-seed`, `claude-md-track`, `stub-creation`, `settings-json`, `gitignore` (lines 29-51) - while "Pre-flight lock acquisition" (line 65) happens *after* that check. So two concurrent `/ds-wrap` invocations can genuinely race on those sites, unprotected by the lock the doc says covers them.

Suggested fix: either move the pre-flight scaffold-accuracy check inside the lock window, or re-scope the NORMATIVE section's claim to exclude the pre-lock sites and explicitly document the residual race there.

Everything else the skeptic checked (the ownership corrections, the wrap-ticket writer carve-out cross-reference, the regression test assertions) looked sound.

tyson-solara6:
Superseded by the rebased branch: `main` moved 10 commits underneath this PR and it went CONFLICTING. Force-push is denied here, so the work was rebased onto a new branch rather than rewriting this one. All reviewed content is byte-identical; no re-review needed beyond CI.

PR #598, opened 2026-08-08
feat(implement-ticket): Phase 11e knowledge commit onto the PR branch

tyson-solara6:
Ticket-session knowledge now rides the ticket's own PR instead of stranding in the operator's local checkout. **Unit 2** of the knowledge-commit feature; Unit 1 (the test harness, #581) is already on `main`.

## The problem

Knowledge written during a ticket session (`MEMORY.md`, `decisions.md`, `.agentic/learnings.md`) landed only in the conductor's checkout and never reached the PR. The obvious fix - "write it in the worktree" - is impossible: the isolation worktree is destroyed at Phase 8, **before** the PR opens at Phase 9, while `wrap-ticket` writes at Phase 11b.

So the write stays where it is and the **commit** moves. New **Phase 11e** builds a commit with a temporary index plus `commit-tree` and pushes the SHA directly to `refs/heads/$BRANCH_NAME`. No checkout, no worktree, HEAD never touched, and the real `.git/index` never read for staging or written.

Per-unit attribution is out of reach and is not attempted - fan-out unit branches are merged and deleted before Phase 6 fires. Per-**ticket** is what this delivers.

## Behavior

- **Trigger:** every ticket where Phase 9 opened a PR. Skipped when no PR exists, and when `knowledge_commit_on_pr` is `false` (new toggle, default `true`).
- **Does not inherit Phase 11b's Trivial skip** - `learning-extractor` and `learnings-agent` may still have written `.agentic/learnings.md`, and gating on Trivial would silently drop exactly that.
- **Reads from disk, never an agent return value.** The deleted Phase 11c read `wrap-ticket`'s return JSON, so writes by other agents were invisible to it; it shipped zero commits for its entire lifetime.
- **No path-prefix floor.** `.agentic/learnings.md` is a first-class candidate. A categorical `.agentic/*` refusal was another of 11c's three defects.
- **Revert guard, fail-closed.** A stale local copy can revert knowledge another session merged. Deletions warn loudly and proceed when a human will read the diff; under `auto_merge_on_ci_green: true` nobody reads it, so the commit is skipped. A `diff-index` failure counts as risk **present**.
- **Part G gains a dedup gate** so `/ds-wrap` does not re-commit what the ticket PR already carries. Fail-open toward committing: a duplicate is a reviewable diff, a silent drop is not.
- In DinoStack all three files are gitignored, so this is a deliberate, **audible** no-op here. Consumer projects get the commit.

## Defects found by running it, not reading it

- **zsh applies the `:r` history modifier to an unbraced `$VAR:` even inside double quotes.** `"$KC_SHA:refs/heads/..."` became `<sha-minus-extension>efs/heads/...` and the push failed with a misleading `src refspec does not match any`. 9 of 47 tests red, zsh only. Both colon-adjacent expansions are now braced. Second occurrence of this class in this repo.
- **A test that counted diagnostic lines missed a real suppression.** Emptying the matched rule still printed `(rule: )` - present but useless. The assertion now requires the rule text to be real.
- **A documented mutation was a no-op.** `check-ignore -v` writes to stdout, so `2>/dev/null` changes nothing; the real suppression is `>/dev/null`. Corrected, and the other docstrings audited for the same stdout/stderr confusion.
- **Two guards are genuinely defense-in-depth.** Deleting either the staging guard or the tree-identity short-circuit alone still produces no empty commit; only deleting both reds the test. A single-mutation check would have reported a false green.

## Verification

- **47 new tests** over `{bash, zsh}`, executing the real extracted blocks against real git repos with bare origins. **Every assertion mutated and confirmed red.**
- `test_wrap_knowledge_commit.sh` extended 31 -> **41 assertions** covering the new dedup gate: its five fail-open conditions, its load-bearing "evaluated last" position (mutation: hoist it above the byte-identity bullet -> red), the `knowledge-commit-state.json` writer/reader contract across two files, and the `11c`-absence invariant extended to the new phase number.
- Full suite **1026 passed**; four floors (18 / 52 / 30 / 47); hooks JS 44/0; shell suites 0 failing; adapter drift clean over the CI pathspec copied verbatim; symlinks relative after an in-worktree rebuild; resident budget OK with `content/rules/**` and `content/sections/**` byte-unchanged.
- Skeptic review: 0 Critical, 1 Major, 6 Minor - all resolved. The Major was that the dedup gate had no coverage anywhere, which is what the 10 new assertions fix.

## North Star alignment

- **Verifiable outcomes autonomously** - advances it. Two methodology shell blocks that were previously prose-only are now mechanically executed under two shells, and every assertion is mutation-proven rather than assumed.
- **Operator attention** - advances it. Knowledge reaches the reviewer inside the diff they already read, instead of stranding locally until someone remembers to commit it by hand.
- **Works for everyone** - neutral. No operator identity, tracker, or harness assumption; behavior is per-project via `knowledge_commit_on_pr`, and the phase soft-fails absolutely - it never blocks a PR, a merge, or a session.
- **Known trade-off, named in the diff rather than papered over:** the commit re-triggers CI, and Phase 10a cannot re-fire, so a knowledge commit that reds a check leaves the PR red for the operator. Bounded by markdown-only scope, an explicit runtime `NOTE:`, and the kill-switch toggle. No pillar regressed.


tyson-solara6:
Superseded by #600. `check-adapter-sync` failed here after `main` advanced to c9f44f77 (DS-143, #599) - that gate tests the merge result, not the branch, so a base move invalidates it. Rebasing rewrote the pushed commits and force-push is blanket-denied in this repo, so the work moved to `feature/phase-11e-knowledge-commit-v2` rather than being bundled or force-pushed. Same four commits, same review history (four rounds, sign-off granted), rebased onto c9f44f77 with adapters rebuilt.

PR #604, opened 2026-08-08
DS-143 follow-up: close four gaps in the trigger-loaded methodology gates

tyson-solara6:
## Summary

Four gaps in DS-143 (#599), found by adversarial review of a parallel implementation of the same design. All four are additive fixes - nothing from #599 is reverted or restructured.

**1. The migration detector scanned the whole `CLAUDE.md`.** `.claude/install.sh` tested `old_import_marker in existing` where `existing` is the entire file, so the one-time migration fired on the import string appearing anywhere. Reachable path: a user who re-adds `@skills/agentic-engineering/METHODOLOGY.md` to their own section to restore the old always-loaded behaviour had `skill_auto_load` forced true on every subsequent `/ds-update`, permanently overriding a deliberate choice. The in-code comment claimed the check was self-disarming; it was not. Now scoped to managed-block content via `finditer` across all blocks, congruent with the rewrite's `pattern.sub` (which has no `count=`).

**2. The embed-budget gate could not detect losing a `content/rules/*` file.** Verified live before the fix: excluding `code-standards.md` from `.claude/build.sh`'s embed loop rebuilt to exactly 117,144 B and the gate exited 0 - 9 KB of always-embedded rules content silently absent, comfortably inside the FLOOR/CEILING band. A single `conventions.md` tail phrase only witnesses truncation at the end, and `code-standards.md` sorts before it. Now asserts every embedded source file's own top-level heading, for both `content/sections/*.md` and `content/rules/*.md`.

**3. Deleted-source tautology.** Expectations derived from the same glob the build reads move together with the content, so deleting a source file outright removed both the content and the expectation. Closed with pinned `EXPECTED_SECTION_COUNT` / `EXPECTED_RULES_COUNT` constants checked before the heading loop, with direction-aware diagnostics: `found > expected` says bump the named constant, `found < expected` says restore the missing file and do not lower it.

**4. Test coverage for the above**, plus a fixture-generator fix so the pre-existing bound scenarios survive the new completeness check.

## Verification

Every fix was mutation-tested - the guard was removed, the corresponding assertion confirmed red, then restored:

| Mutation | Result |
|---|---|
| migration scope reverted to whole-file | case (p) alone reddens |
| migration scope reverted to first-match-only | case (o, lean_then_old) alone reddens |
| `code-standards.md` excluded from the embed loop | gate exits 1, names the file (pre-fix: exit 0 at 117,144 B) |
| pinned count replaced by a tree-derived count | missing-section scenario reddens |
| duplicate-heading guard disabled | 2 assertions redden |
| FLOOR `-lt`->`-le` / CEILING `-gt`->`-ge` | the corresponding exact-bound assertion reddens |

Suites: `install-converge` 53/53, `test_check_skill_embed_budget` 19/19, `test_check_resident_budget` 13/13, codex hook 9/9, gemini hook 2/2 - all under bash and zsh. Both gate scripts exit 0.

## New contributor obligation - please read

Fix 3 means **adding any `content/sections/[0-9][0-9]-*.md` or `content/rules/*.md` file now fails CI** until `EXPECTED_SECTION_COUNT` or `EXPECTED_RULES_COUNT` is bumped in the same commit. This is the same shape as the existing `THRESHOLD` obligation for `check-resident-budget.sh`. Documented in `CONTRIBUTING.md` and `AGENTS.md` rather than left as a script comment.

## North Star alignment

- **Verifiable outcomes autonomously.** The embed gate demonstrably missed a failure class it existed to catch (117,144 B, exit 0, an entire rules file gone). It now detects loss as well as growth, and every guard is pinned by a mutation-tested assertion rather than by a passing run.
- **Works for everyone.** Fix 1 stops the installer overriding a user's deliberate `skill_auto_load: false`. The affected users are precisely those who hand-edit their `CLAUDE.md` to restore behaviour they preferred - the change respects that choice instead of silently reversing it on every update.
- **Guard operator attention.** Trade-off, surfaced rather than buried: fix 3 adds a constant-bump step when source files are added. That friction is the price of closing the tautology, and it is now documented in both contributor-facing files.

Refs #599. Ticket DS-143.


tyson-solara6:
Automated review note (author == reviewer, so no formal approve/request-changes verdict per self-review policy — informational only).

Ran a skeptic pass, a QA test-coverage pass, and a North Star alignment check against this diff.

**Skeptic (correctness):** No Critical or Major issues. All four claimed fixes hold up under adversarial review — the migration-detector scoping genuinely checks every managed block (not just the first), the per-file heading assertion is a real "token survives the pipeline" check (not tautological), and `EXPECTED_SECTION_COUNT`/`EXPECTED_RULES_COUNT` are genuinely pinned literals with no test/impl drift risk. One **Minor, unverified** risk: the new embed-completeness check assumes every real `content/sections/*.md` and `content/rules/*.md` file's true top-level heading is `## ` and that no two real files share the same first `## ` heading text. Worth confirming `bash scripts/check-skill-embed-budget.sh` passes clean against the current repo (or that CI already did) before merge.

**QA (test coverage):** Substantive coverage of all 4 claimed fixes with real mutation-style reproductions, not restatements — including both count-mismatch directions for the deleted-source tautology. Two moderate gaps:
- The `content/rules/*.md` side of the embed-completeness check (`EXPECTED_RULES_COUNT`) shares `_check_embedded_set` with the sections side but has zero dedicated test cases (different exclude-filter branch) — an untested code path, not just an untested restatement.
- The "source file has no top-level heading at all" failure branch in the gate script has no test; only "heading present in source but dropped from SKILL.md" is covered.

**North Star:** ALIGNED — closes silent-failure gaps (false migration triggers, an undetectable dropped rules file, a tautological count check) without adding operator attention burden.

Neither gap is blocking; flagging for optional follow-up before/after merge since this is our own PR.

PR #608, opened 2026-08-09
fix(ds-151): mechanical volume gate for conductor turn shape

tyson-solara6:
## Summary

`hooks/enforce-turn-shape.py` checked identity-line shape, warrant classification, and forced-yield shape, but its length check (added in round 1) was a per-warrant exclusion model: each region (status prose, fenced content, `## Operator decisions`) was independently subtracted out of the count, subject to a per-warrant budget table, with no aggregate bound on the subtraction. Two convergence failures followed from that shape - CF-1 (pasting many separate fences multiplied the free-line pool without bound) and CF-2 (the count stopped at the `## Operator decisions` heading, so anything unrecognized below it fell into an unaccounted gap). This PR replaces the whole exclusion model with a single whole-message non-negative **charge model**: every non-blank line contributes 0 or 1 to a running total, or participates in one of three named, individually-capped free pools; nothing is ever subtracted, so there is nothing left to multiply or leave unaccounted.

## What changed (as shipped)

- `_segment`/`_regions`/`_decision_items` - new shared helpers, the single source of truth for fence and region structure, consumed by both the volume check and the item-sprawl check (the old design had two independent parsers drift apart, which is what produced CF-2).
- `_turn_charge` (replaces `_count_core_body_lines`): computes `charge = status_charge + fence_charge + decisions_charge`.
  - `status_charge` - every unfenced, non-blank status-region line charges 1, unless it is a well-formed `Waiting:` line (matches the `Waiting:` shape AND is <=120 chars, `WAITING_LINE_MAX_CHARS`) **and** `stoppage` is the SOLE warrant present on the turn, in which case it charges 0.
  - `fence_charge = max(0, fenced_nonblank_lines_in_status_region - FENCE_FREE_LINES)` - a single AGGREGATE pool across every fence in the status region (20 free lines total, not 20 per fence), which is what makes CF-1 structurally impossible: there is one scalar pool, not one per fence.
  - `decisions_charge` - every non-blank line under the `## Operator decisions` heading not folded into a recognized item charges 1 (non-recognition now fails toward charging, closing CF-2), plus `sum(max(0, item_line_count - ITEM_FREE_LINES) for each item)`. Fenced content inside an item folds into that item's own line count rather than escaping into the status-region fence pool - charged once, never twice.
- Flat `BASE_BODY_BUDGET = 10` replaces the deleted `BODY_BUDGET_DECISION/STOPPAGE/COMPLETION/ANSWER` table and the "most generous of multiple warrants' budgets" rule entirely. The deleted `answer` budget in particular was worse than merely self-serviceable: its only detector (`_QUOTED_FRAGMENT_RE`) fires on any incidental 8+ char quoted fragment, so most substantive turns acquired the most generous budget by accident.
- `_classify_warrants` is now restricted to the identity line plus UNFENCED body lines - a warrant token that only appears inside a pasted fence is an example being discussed, not a live warrant.
- `_decision_item_sprawl_flag` rewritten to consume the shared helpers; item COUNT stays completely unbounded (per `content/sections/02-delegation.md`'s ban on a numeric item cap), only per-item LINE COUNT is bounded at `ITEM_FREE_LINES = 3`, now correctly folding in fenced continuation lines.
- `content/references/conductor-turn-format.md` §Hook contract rewritten to describe the charge model, the free-pool table, and the corrected residual-false-positive list (3, not 4 - the per-warrant `answer`-budget false positive no longer exists because there is no per-warrant budget).
- `hooks/tests/test-turn-charge-model.py` (new) - an independently-written reference implementation of the charge definition, enumerated exhaustively over an 8-symbol line alphabet for lengths 1-6 (299,592 messages, 877,376 monotonicity checks), asserting 8 invariants (I1-I8: monotonicity, pure-prose exactness, no-crash/no-negative, the unbounded-`Waiting:` and unbounded-item-count constraints, the bounded-free expression, fence-count monotonicity, and `charge <= nonblank_body_lines`).

## Corrected claims from the original PR description

The original description of this PR (round 1) is stale on three points and is corrected here:
- ~~`BODY_BUDGET_ANSWER = 15`~~ - deleted. There is no per-warrant budget table any more; every warrant is judged against the same flat `BASE_BODY_BUDGET = 10`.
- ~~The Operator decisions block is "never length-bounded (no item cap, no aggregate line cap)"~~ - item COUNT is still unbounded (unchanged, per `02-delegation.md`), but each individual item's LINE COUNT **is** bounded at `ITEM_FREE_LINES = 3`, and non-item lines under the heading now charge full weight toward the whole-message budget. The decisions region is no longer exempt from measurement outright.
- ~~"11 new regression cases (s1-s11)" / "54 assertions"~~ - the shipped fixture suite is 91 assertions in `hooks/tests/test-enforce-turn-shape.py` (existing fixtures re-sized/re-labeled per the flat-budget flip, plus `v1`-`v9` new regression fixtures for both convergence-failure classes) and 18 assertions in the new `hooks/tests/test-turn-charge-model.py`.

## Old-code fixture evidence (test-strategy obligation)

Every `v1`-`v9` regression fixture is a **verified escape**: it must fail (produce a different verdict) against the pre-DS-151 hook (commit `40ba9ce4`), or the fixture is vacuous. Running the `v1`-`v7`/`v7b` fixtures directly against that commit's `hooks/enforce-turn-shape.py`:

```
v1:  QUIET   (new hook: ADVISORY, charge is 69 - CF-1 gate)
v2:  QUIET   (new hook: ADVISORY, charge is 25 - minimal multiplication case)
v3:  QUIET   (new hook: ADVISORY, turn volume exceeded - CF-2, no item marker)
v4:  QUIET   (new hook: ADVISORY, turn volume exceeded - CF-2, no markers at all)
v5:  QUIET   (new hook: ADVISORY - CF-2, column-2-indented item)
v5b: QUIET   (new hook: QUIET - contrast fixture, not an escape; correct on both)
v6:  QUIET   (new hook: ADVISORY, status-only - fenced heading, zero warrants, A8)
v7:  QUIET   (new hook: ADVISORY, turn volume exceeded - fat Waiting: lines, A1)
v7b: QUIET   (new hook: QUIET - contrast fixture, not an escape; correct on both)
```

`v1`-`v4`, `v5`, and `v7` all flip from QUIET (old, buggy) to ADVISORY (new, fixed) - the required proof that the fixture actually exercises the fixed defect rather than passing on both sides. `v5b` and `v7b` are deliberate contrast fixtures (not escapes) and are QUIET on both, as intended - they prove the fix does not over-flag a legitimate 40-item or 50-`Waiting:`-line turn.

## North Star alignment

Pillar 1 (`docs/overview/vision.md:19`: "Guard operator attention. Surface decisions and work-stoppages, not status.") is enforced for turn length by prose (`content/references/conductor-turn-format.md`) with a mechanical backstop in `enforce-turn-shape.py`. Round 1 of that backstop introduced its own defect class - an exclusion model where subtracted-out regions had no aggregate bound, closing incidents but reopening the same shape twice (CF-1, CF-2). This PR replaces the exclusion model with a charge model precisely so that defect class has no representation: there is nothing to subtract, so there is nothing to multiply or lose track of. The hook stays advisory-only throughout (never blocks the stop, fail-open on any error) - the change is scoped entirely to how sprawl is measured, not to the hook's non-blocking contract.

## Test plan

- [x] `python3 hooks/tests/test-enforce-turn-shape.py` - all 91 assertions pass.
- [x] `python3 hooks/tests/test-turn-charge-model.py` - all 18 assertions pass (299,592 enumerated messages, 0 violations).
- [x] Full `hooks/tests/test-*.py` loop (mirrors `bin-tests.yml`'s `hooks-python-tests` job) - all files pass.
- [x] `v1`-`v9` verified against the pre-DS-151 hook (commit `40ba9ce4`) - old-code failure list above.
- [x] `bash scripts/build-all.sh` then `git diff --exit-code` over the adapter-drift pathspec copied verbatim from `.github/workflows/adapter-sync.yml` - only the regenerated `conductor-turn-format.md` copies changed, no stray drift; `.claude/skills/agentic-engineering/{agents,commands,references,rules}` symlinks stayed relative.
- [x] `bash scripts/build-methodology.sh && bash scripts/check-resident-budget.sh` - OK, no change (this unit does not touch the resident set).
- [x] Em-dash check on the branch diff: clean.


tyson-solara6:
Automated review pass (self-authored PR, no formal review verdict since GitHub blocks self-review). Adversarial read found a real gap in the exact mechanism this PR adds:

**Worth fixing before merge:** `hooks/enforce-turn-shape.py:548-551` (`_count_core_body_lines`) - the `## Operator decisions` heading match runs unconditionally, even while `in_code` is `True`. Two consequences:
1. If a fenced code block (e.g. a pasted diff/doc excerpt) happens to contain a line that is exactly `## Operator decisions` - plausible here since this same PR touches 5 copies of `conductor-turn-format.md` containing that literal string - counting stops there entirely, silently dropping everything before *and* after that point in the real message. This reopens the fence-content bypass that findings 2b/s12/s13 were written to close.
2. If an unclosed fence precedes a genuine decisions heading, the buffered fence lines are discarded rather than counted at full weight, contradicting the PR's own documented/tested "unclosed fence => no exclusion, every buffered line counts" invariant (covered for the EOF case by s13, but not for this heading-interrupts-fence path).

Neither case is exercised by the 11 new tests (s1-s13, t1-t4), so it ships without a regression guard. Suggested fix: only check the heading regex when `not in_code`, and flush `fence_buffer` at full weight before discarding on an early heading break.

**Nit:** the PR description's "Budget measurements" section says `BODY_BUDGET_ANSWER = 15`, but the code and tests correctly use `10` - stale text in the description only, not the diff.

PR #609, opened 2026-08-09
feat(DS): add prose-move-impact enumerator for ds-implement-ticket.md split

tyson-solara6:
## Summary

Adds `scripts/prose-move-impact.py`, a mechanical enumerator that answers "what breaks if I move these line ranges out of content/commands/ds-implement-ticket.md" - built to replace four failed prose-only architect review rounds on that exact split, each of which fixed the named defects and left or reintroduced structurally identical ones.

## North Star alignment

This diff touches `bin/tests/**`, which IS a canonical methodology-shaping trigger path (`content/agents/skeptic.md` step 3.6's applicability list includes `bin/**`) - the vision-alignment check applies, not "N/A".

Advances **"produce verifiable outcomes autonomously"**: this tool replaces four failed prose-only architect review rounds on the same split with a mechanical enumerator that either resolves every consumer's assertion or fails loud (`UNRESOLVED`) - a human/architect gets a deterministic verdict instead of another round of manual re-reading. The tradeoff being fixed in this round: r2 review found the test suite's own known-answer fixtures pinned that verdict to hardcoded absolute line numbers, which is a low-friction cost on every future PR that edits the target file above the highest pinned line (silently reddening the required `python-bin-tests` check for unrelated work) - now derived from heading text instead, so the verifiability gain no longer taxes unrelated contributors.

## Review rigor

- Brief / Plan path: N/A - spawned directly as a Worker unit with a detailed task brief, no separate Architect plan artifact
- Skeptic rounds (tier): pending (this PR is submitted for Skeptic review per the spawning conductor's workflow)
- Findings summary: N/A pending Skeptic review

## What it does

- Discovers every consumer of the target file under `bin/tests/`, `hooks/tests/`, `content/`, `scripts/`, `.github/` via `git grep`.
- Scopes mechanical assertion extraction to actual executable gates (`bin/tests/**`, `hooks/tests/**`, `scripts/check-*.sh`) - doc/prose consumers (`.github/prompts`, `content/commands/*.md`, etc.) are reported separately via a lightweight heading-citation check, never via arbitrary literal matching. An earlier iteration matched on any quoted string shared with the (3600-line) target and produced hundreds of false "BREAKING" rows on plain words like "default"/"status"/"worktree" before this scoping was added.
- Extracts: literal presence/absence/count assertions, `_extract_block()`-style heading-anchored block readers (fence-aware - tracks ``` open/close state so a heading inside a fenced example is never mistaken for a real section), Python `re.compile(...)` patterns (via `ast`, constant-folded through string concatenation), and shell `git grep -c[iE]` count-floor assertions including heredoc per-file floor tables (`git grep -c` counts MATCHING LINES, not raw occurrences - verified against a real target line that contains a pinned phrase twice).
- Reports each checked consumer's inferred SCANNED SET (single-file hardcode vs. `content/**`-wide) and flags when a proposed destination leaves that scanned set even when the consumer's own numeric floors still clear - the false signal that defeated three of the four failed review rounds.
- Fails loud: any consumer or assertion it cannot resolve lands in an explicit `UNRESOLVED` section and forces a non-zero exit, including a non-vacuity guard that refuses to report OK if discovery finds zero consumers.

## Testing

`bin/tests/test_prose_move_impact.py` (auto-discovered by `bin-tests.yml`'s `python3 -m pytest bin/tests/ -q`, 24 tests as of r2) pins the four real, previously-verified misses from the failed review rounds as known-answer fixtures:

- `test_tracker_dev_complete_spec.py`'s `_extract_block`/`HEADING` reader breaks when `480-548` (`## Tracker Writeback Helper`) moves.
- `test_tasks_jsonl_fold.sh`'s `FOLDSPEC` per-file floor (`>= 7`) drops to 4 when `1369-1458` + `1651-1763` move (verified: before=14, moved=10, after=4) - **plus** its separate `G6` floor (`git grep -cE 'is in_progress under another session' -- "$DIT"`, floor `-ge 1`), whose only occurrence (line 1448) also falls inside `1369-1458` (before=1, after=0), fixed in r1 (see below).
- `test_batch_state_timestamp_field.sh` really carries 8 `_absent(` call sites, not 2 (pin on the fixture premise), **plus** a second fixture that actually exercises the tool's shell-continuation joiner (the original one was vacuous - see r1).
- `test_loop_state_site_coverage.sh` hardcodes a single-file `FILE=` scan with no `content/**` fallback - flagged as leaving the scanned set.

Plus regression tests for all four r1 Skeptic findings (see below), a mutation test confirming `discover_consumers()` returning zero consumers must NOT produce an OK report, and a **differential test against ground truth**: `test_differential_against_ground_truth_post_move_gates` builds a scratch git repo with the 9 ranges actually deleted from the target and actually written to the 6 destination files, then runs the two real shell gates (`test_batch_state_timestamp_field.sh`, `test_tasks_jsonl_fold.sh`, both honor `GATE_REPO`/derive their own root) against it and compares real pass/fail to the tool's prediction. It fails loud (disqualifying) if the tool ever reports a consumer clean that actually fails post-move; over-report is recorded, not asserted on. Mutation-verified: reintroducing the r1 pre-fix behavior (disabled ERE fallback + fail-loud suppressed) makes this test go red. Scoped to these two shell gates rather than the full 8 shell + 7 pytest set the reviewer used by hand - see the test's own docstring for why and how to extend it.

## Result against the ticket's 9 proposed ranges

Run against all 9 ranges from the ticket brief (all 9 range boundaries re-verified against the live `origin/main` target - zero drift, every start is a real heading line): as of r2, **128 BREAKING assertions, 3 scanned-set findings (2 LEAVE the scanned set), 5 UNRESOLVED items, verdict FAIL** (exit 1, confirmed). The BREAKING count rose from 126 to 128 in r2 because the pattern-slot meaningfulness-filter fix (Major 1 below) stopped silently dropping `'mark-blocked-and-continue'`/`'<ISO8601>'` as assertions. Notably surfaces a whole additional consumer file, `test_tracker_writeback_ranking_spec.py`, with ~30 of its own load-bearing assertions on the `## Tracker Writeback Helper` block - a file that appeared in none of the four failed review rounds' matrices.

Of the 5 UNRESOLVED items: 3 are "no assertion extractable" (`bin/tests/fold_model.py`, `bin/tests/test_check_command_file_budget.sh`, `scripts/check-command-file-budget.sh`) and 2 are genuine regex-polarity ambiguity on patterns whose matches DO fall inside a move range (`test_tracker_dev_complete_spec.py`'s `TSV_LITERAL_RE`, `test_tracker_writeback_ranking_spec.py`'s `_CANCEL_RE`) - verify these by hand before the split. `scripts/codex-skills.py` is discovered as a doc consumer (its generated `.codex/skills/implement-ticket/RESOURCE-MAP.json` will need the new `content/references/*` paths during the actual refactor - flagged for docs-currency review, not a mechanical gate).

## r1 Skeptic fixes

- **Critical 1** (silent literal-vs-ERE drop): `_present`/`grep -qiE` patterns containing regex metacharacters were matched via naive substring `find()` and silently dropped on zero literal occurrences. Fixed via `_resolve_literal_in_target`, an ERE fallback (case-insensitive, mirroring `grep -qiE`) that only ever escalates the actual `<pattern>` slot of a shell `_present`/`_absent` call scoped to a variable that resolves to the target file (never a label argument, another spec file, or unrelated Python-side literals - each was a real false-positive source hit while narrowing this). Also fixed `_STR_RE`'s single-quote branch, which excluded backslashes and so never even extracted ERE patterns with backslash-escaped metachars (`\(`, `\+`) as candidate literals in the first place - a distinct, earlier failure. Zero-match found: `if not occ: continue` -> now `Unresolved`, never silently dropped.
- **Critical 2** (`$`-anchored floor regex): `_VAR_ASSIGN_RE` required `$(...)` to end the line, missing `test_tasks_jsonl_fold.sh`'s `G6` floor (`; [ -n "$G6" ] || G6=0` trails the assignment). Un-anchored + made non-greedy. A second bug surfaced once the assignment matched: the resolved file-variable token (`"$DIT"`) wasn't stripped of its `$` before the `varmap` lookup, so the floor still silently dropped - fixed.
- **Major 3** (vacuous fixture): the shell-continuation-joiner known-answer test's predicate (`startswith("_absent ")`) was satisfied by the UNjoined first physical line too, never actually exercising the joiner. Added a second fixture requiring the joined logical line to also END with the pattern's closing quote - mutation-verified red under `while False:`.
- **Major 4** (vacuous OK on zero checked consumers): `analyze()`'s non-vacuity guard only fired on zero DISCOVERED consumers, not zero CHECKED ones - a `bin/tests/` rename or `SEARCH_DIRS` drift would silently report OK. Added a second guard.

All four have regression tests, each confirmed failing pre-fix by literal mutation (see individual test docstrings for the pre-fix attestation).

## r2 Skeptic fixes

- **Major 1** (pattern-slot literal silently dropped by the meaningfulness filter): `_is_meaningful_literal`'s short-token noise filter ran BEFORE `is_pattern_slot` was computed, so a pattern-slot literal (the actual `<pattern>` payload of a `_present`/`_absent` call) with no `[`:/=$]` punctuation was dropped with no Assertion and no UNRESOLVED row - the same silent-drop class as Critical 1, one filter earlier in the same function. Three live instances in `test_batch_state_timestamp_field.sh`: `'mark-blocked-and-continue'`, `'fail-open'`, `'<ISO8601>'`. Fixed by computing `is_pattern_slot` first and exempting it from the meaningfulness filter (a pattern slot is by construction a deliberate assertion). Regression-tested and mutation-confirmed red under the pre-fix ordering.
- **Major 2** (zero-match regex reclassification could mask a real break): `extract_python_regex_patterns` discarded the compiled pattern's own `re.IGNORECASE`/`re.DOTALL`/`re.MULTILINE` flags, and matching was per-line while a real consumer searches whole text - both could make a pattern that genuinely matches the live target (case-insensitively, or across a newline) report ZERO pre-move matches and get reclassified as a resolved, non-breaking assertion instead of UNRESOLVED. Fixed: `extract_python_regex_patterns` now captures the actual flags from the `re.compile(...)` call site (positional or `flags=` keyword, including `|`-combined flags), and matching runs against the whole target text via `finditer` + offset-to-line mapping rather than per-line. Two regression tests (IGNORECASE case, DOTALL cross-line case), both mutation-confirmed red under the pre-fix behavior.
- **Major 3** (hardcoded absolute line numbers in a required-CI test): `test_known_move_ranges_align_exactly_with_heading_boundaries`, `NINE_RANGES`, and every individual known-answer fixture's move range hardcoded absolute line numbers, so any unrelated edit above the highest one anywhere in the 3600-line target would silently redden this suite (which the required `python-bin-tests` check collects) with a diagnostic about a refactor that hasn't happened. Fixed: every range is now derived at test-collection time from the tool's own `fence_aware_headings()` by heading TEXT (`_range_for_heading`), never a line number. Mutation-proof: inserting an unrelated sentence at line 20 of the live target and re-running the full 24-test suite stays green (previously would have reddened the hardcoded-number test).
- **Minor** (`Report.ok` ignored `scanned_sets`): a run whose only finding was `leaves_scanned_set=True` (nothing else broke, nothing unresolved) exited 0 and printed `Verdict: OK` - the exact finding class the scanned-set column exists to surface, unenforced. Fixed: `Report.ok` now also fails when any scanned-set finding has `leaves_scanned_set=True`; the module manifest's exit-code contract is corrected to match. Regression-tested (isolated Report construction) and mutation-confirmed red under the pre-fix property.
- Docstring corrections: the differential test's "only check ... that can catch the NEXT under-report class" overclaim (both in the tool's own manifest and the test module's docstring) is corrected to name its actual scope - 2 of the 7 tool-clean consumers the reviewer manually differential-tested, with the reviewer's pre-move-red-skip generalization approach recorded as the path to extend it.

All four r2 fixes have regression tests; each is separately mutation-confirmed against the specific pre-fix code path (not just re-run against the unfixed function in the abstract) per the finding above.

## Checklist

- [x] DCO sign-off on every commit (`git commit -s`)
- [x] Affected adapters declared (none)
- [ ] Before/after transcript included for agent-behavior changes (N/A - not an agent-behavior change)
- [ ] `install.sh` re-run locally and behavior verified (N/A - no install.sh path touched)
- [x] Tests/lint passing
- [x] Docs updated if behavior changed (N/A - tool has no existing doc surface to update; module manifest is in the file header)

## Affected adapters

- [x] None (methodology / docs only)

### DIFF
--- a/bin/tests/test_prose_move_impact.py
+++ b/bin/tests/test_prose_move_impact.py
@@ -99,16 +99,36 @@ TARGET = "content/commands/ds-implement-ticket.md"
 # ---------------------------------------------------------------------------
 
 
-def _heading_index() -> dict[str, "pmi.Heading"]:
+def _heading_index() -> tuple[dict[str, "pmi.Heading"], dict[str, int]]:
     target_path = REPO_ROOT / TARGET
     lines = target_path.read_text(encoding="utf-8").split("\n")
-    return {h.text: h for h in pmi.fence_aware_headings(lines)}
+    headings = pmi.fence_aware_headings(lines)
+    # This dict comprehension is last-wins on a duplicate heading text,
+    # while the tool's own extract_heading_block_assertions (line ~672) is
+    # first-wins via `next(...)` - the two disagree on which occurrence a
+    # duplicate heading resolves to. Rather than silently picking either
+    # one, track per-text counts so _range_for_heading can fail loud on any
+    # duplicate instead of guessing which occurrence the fixture meant.
+    counts: dict[str, int] = {}
+    for h in headings:
+        counts[h.text] = counts.get(h.text, 0) + 1
+    return {h.text: h for h in headings}, counts
 
 
-_HEADINGS = _heading_index()
+_HEADINGS, _HEADING_COUNTS = _heading_index()
 
 
 def _range_for_heading(heading_text: str, dest: str) -> "pmi.MoveRange":
+    count = _HEADING_COUNTS.get(heading_text, 0)
+    if count > 1:
+        raise AssertionError(
+            f"heading text {heading_text!r} appears {count} times in the live target - "
+            "ambiguous: this index is last-wins while the tool's own "
+            "extract_heading_block_assertions is first-wins, so picking either occurrence "
+            "here would not provably match what the tool itself resolves. Disambiguate the "
+            "fixture with a more specific heading text, or resolve the duplicate in the "
+            "live target, before using this heading in a proposed range."
+        )
     h = _HEADINGS.get(heading_text)
     if h is None:
         raise AssertionError(
@@ -551,6 +571,61 @@ def test_regression_regex_assertion_matches_across_a_newline():
         "a DOTALL pattern spanning the alpha/beta newline must match against "
         "the whole target text - a per-line scan sees zero matches here"
     )
+
+
+# ---------------------------------------------------------------------------
+# Regression: Major (r3 Skeptic finding) - the zero-match "cannot break"
+# reclassification was unsound whenever this tool's matching granularity
+# (whole-text) differs from the consumer's real granularity. Round 1's fix
+# (compile flags) and round 2's fix (whole-text over per-line) each fixed a
+# DIFFERENT cause of the same false-clean-pass shape without fixing the
+# shape itself - a THIRD cause (an anchored, non-MULTILINE pattern that a
+# consumer applies per-token, e.g. `.match(name)` against an extracted
+# identifier) still produced whole-text zero while the real per-consumer
+# match count was nonzero. This mirrors the live
+# bin/tests/lib/md_shell_extract.py::_IDENTIFIER_RE case: 0 matches
+# whole-text, 29 matches per-line, 7 inside a proposed move range. The fix
+# deletes the reclassification entirely - a zero-match regex assertion is
+# now always UNRESOLVED, never inferred non-breaking.
+# ---------------------------------------------------------------------------
+
+
+def test_regression_zero_whole_text_matches_never_resolved_non_breaking():
+    """A round-3 regression: an anchored (^...$), non-MULTILINE pattern that
+    matches every individual line but zero times against the whole target
+    text (anchors bind to the whole string's start/end, not per-line,
+    without re.MULTILINE). Pre-fix, this whole-text zero was silently
+    reclassified `resolved=True, breaks=False` ('zero pre-move matches
+    means this move cannot change the assertion's outcome either way'),
+    discarding the fact that the consumer applies the pattern with
+    different granularity (e.g. per-token via `.match(name)`) than this
+    tool's own whole-text scan. Confirmed failing pre-fix: the old code
+    appended a resolved, non-breaking Assertion and emitted nothing to
+    `unresolved` for this exact case."""
+    consumer_text = (
+        "import re\n"
+        "_TOKEN_RE = re.compile(r'^[A-Za-z_][A-Za-z0-9_]*$')\n"
+        "def check(name):\n"
+        "    return bool(_TOKEN_RE.match(name))\n"
+    )
+    # Every individual line is a bare identifier, but the whole text never
+    # matches ^...$ without re.MULTILINE: '^' only binds to offset 0 and
+    # '$' only to the string's end, and there is a newline directly after
+    # 'alpha'.
+    target_text = "alpha\nbeta\ngamma\n"
+    line_starts = pmi.build_line_starts(target_text)
+    ranges = [pmi.MoveRange(start=2, end=2, dest="content/references/scratch.md")]
+    assertions, unresolved = pmi.extract_regex_assertions(
+        "scratch_consumer.py", consumer_text, target_text, line_starts, ranges
+    )
+    assert not assertions, (
+        "a zero-match regex assertion must never be resolved as "
+        "non-breaking - it must always surface as UNRESOLVED so a human "
+        "verifies the consumer's real matching granularity by hand"
+    )
+    hits = [u for u in unresolved if "_TOKEN_RE" in u.detail]
+    assert hits, "expected an UNRESOLVED row for _TOKEN_RE, not a silent resolved pass"
+    assert "ZERO" in hits[0].detail
 
 
 # ---------------------------------------------------------------------------
--- a/bin/tests/test_prose_move_impact.py
+++ b/bin/tests/test_prose_move_impact.py
@@ -1,11 +1,18 @@
 #!/usr/bin/env python3
 """
-Purpose: Known-answer and mutation tests for scripts/prose-move-impact.py -
-         the mechanical "what breaks if I move these line ranges" enumerator
-         built to replace four failed prose-only review rounds on splitting
+Purpose: Known-answer, regression, mutation, and DIFFERENTIAL tests for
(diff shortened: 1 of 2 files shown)


tyson-solara6:
Automated review pass (self-authored PR, no formal review verdict since GitHub blocks self-review). Adversarial read found two real, narrow correctness gaps introduced by this PR (both new files, no pre-existing/out-of-scope debt found nearby):

1. `scripts/prose-move-impact.py:381` (`extract_literal_assertions`) - skips a literal only if it equals `DEFAULT_TARGET` or its basename, but the function only ever sees the hardcoded module constant, not the actual `target_rel` passed via the tool's own `--target <path>` option. Running the tool against a non-default target means self-references to *that* target's path won't be exempted and could be mis-scanned as content literals.

2. `scripts/prose-move-impact.py:622-639` (`_count_pattern_in_target`) - when a `git grep -c[iE]` ERE pattern fails to compile as a Python regex, the code silently falls back to a plain substring match instead of recording an `UNRESOLVED` finding. This contradicts the tool's own stated "fail loud, never silently skip" invariant (restated in this function's own docstring) and could produce a wrong before/after count - and thus a wrong break/no-break verdict - without ever surfacing in `UNRESOLVED`.

Neither is exercised by the current known-answer/mutation test suite (all fixtures use the default target and Python-regex-compatible patterns). Low severity given current usage (single default target, one caller), but worth tightening since the tool is meant to be reused for future doc splits.

PR #613, opened 2026-08-09
fix(tests): harden the mutate-tracked-source scaffold in install-converge.test.sh (DS-148)

tyson-solara6:
## Summary

DS-143 added test cases (m) and (n) to `.claude/tests/install-converge.test.sh` that mutate real, tracked, shippable source files (`content/templates/claude-managed-content.md` and `content/rules/conventions.md`) to exercise the embed pipeline end to end. A Tier-3 Skeptic signed off on that branch but raised two Minor findings against the scaffold. This PR addresses both.

- **Finding 1 (duplicated scaffold):** cases (m) and (n) each carried their own backup/mktemp/trap/restore block, differing only in target path and whether an adapter rebuild is needed on restore. Extracted a shared `_with_mutated_source <target> <case_fn> [restore_hook]` helper and routed both cases through it. Case (a)'s independent manifest-strip reimplementation was left untouched, as instructed.
- **Finding 2 (concurrent runs corrupt tracked sources):** two runs of this suite against the same checkout could corrupt a tracked source file permanently (run B backs up run A's already-mutated file, then writes that "backup" back as if it were pristine). Closed the class with a cross-process `mkdir`-based lock (`_mut_lock_acquire`/`_mut_lock_release`), keyed to `REPO_DIR` so separate worktrees never contend with each other.

Case (n)'s restore-path adapter rebuild keeps its existing soft-fail-with-warning behavior (never fatal, since it runs inside a trap).

### Why `mkdir` over `flock`, and over a temp clone

- **`flock`** is absent by default on macOS. Any fallback for its absence needs its own portability story or silently no-ops, which is worse than no guard at all. `mkdir` is atomic on any POSIX filesystem and needs no fallback branch - it works identically on macOS and Linux.
- **Temp clone** removes the hazard structurally rather than serializing around it, but doesn't fit case (n)'s design: `$FAKE_HOME`'s skill files are symlinks into the *real* `REPO_DIR` checkout, and the case's assertions depend on `scripts/build-all.sh` regenerating the real checkout's generated artifacts (so the symlinked `SKILL.md` picks up the canary). A clone would build into itself, not into the checkout the test's symlinks point at - the assertions would need a parallel rewrite to reroute every path, which is a materially bigger and riskier change than a lock.

### A genuine bug caught while proving the lock

While verifying the concurrency fix under real concurrent runs (see Test plan), the first draft still corrupted a tracked file: it took the file backup **before** acquiring the lock, so a second process could snapshot the first process's in-flight mutation as its own "pristine" backup and later restore that corruption once its own turn under the lock arrived - the exact race the lock exists to close, just moved one line earlier. Fixed by moving the backup to after lock acquisition, then reproduced two clean concurrent runs to confirm.

## North Star alignment

- **Works for everyone:** the lock uses only `mkdir`, `kill -0`, and `sleep` - no `flock`, no GNU-only flags. Verified directly on macOS (this session) under both `bash -n` and `zsh -n`, and under real concurrent execution. Nothing in the fix is CI-only or Linux-only.
- **Verifiable outcomes:** reproduced the corruption with the pre-fix scaffold under two concurrent runs (leftover canary permanently landed in `content/rules/conventions.md`, plus contaminated generated adapter files), then reproduced two clean concurrent runs with the fix (byte-identical tracked files, both runs green). Also ran a mutation test per routed case (m and n) to confirm the refactor is behavior-preserving: broke one guarded behavior per case, confirmed red, restored, confirmed green, with the same 48-assertion count and shape.
- **No enforcement floor lowered:** the soft-fail warning on case (n)'s adapter rebuild is preserved unchanged (still non-fatal, still runs inside the restore trap). No assertion was removed or weakened; the suite reports the same 48 passed / 0 failed before and after.

## Test plan

- [x] `bash .claude/tests/install-converge.test.sh` - 48 passed, 0 failed (baseline unchanged)
- [x] `bash -n` and `zsh -n` on the test file - both clean
- [x] Reproduced the corruption with the pre-fix scaffold: two concurrent runs left a permanent `case-n-canary` marker in `content/rules/conventions.md` and contaminated generated adapter files, while both runs still reported green
- [x] Confirmed the fixed scaffold under two concurrent runs: both report 48 passed / 0 failed, and both tracked source files are byte-identical to their pre-experiment state afterward
- [x] Mutation-tested the refactor: broke one guarded behavior in each of case (m) and case (n), confirmed red (2 failures / 1 failure respectively), restored, confirmed green again at 48/0
- [x] `git status --porcelain` clean after every run; repo-wide grep for the canary marker matches only the test file itself
- [x] Verified the four `.claude/skills/agentic-engineering/{agents,commands,references,rules}` symlinks were not absolutized by any build run triggered during testing

Refs DS-148.

### DIFF
--- a/.claude/tests/install-converge.test.sh
+++ b/.claude/tests/install-converge.test.sh
@@ -16,9 +16,11 @@
 #                All side effects use TEMP HOME dirs; the real ~/.claude and
 #                ~/.agentic are NEVER touched.
 #
-# Performance: ~50 s wall time (19 install.sh invocations after the DS-143
-#              Skeptic loop 3 fix pass, up from 17; each runs two full adapter
-#              builds against the real checkout).
+# Performance: ~90-120 s wall time, varying by machine load (19 install.sh
+#              invocations, unchanged by DS-148; each runs two full adapter
+#              builds against the real checkout). Measured directly across
+#              several runs - do not re-estimate this figure without timing a
+#              real run; it has drifted from a stale ~50 s estimate before.
 #
 # Regression coverage:
 #   - Change 1: stale "ours" symlink (target under .../DinoStack/...) is
@@ -102,6 +104,18 @@
 #     exists to cover. If the post-case adapter rebuild
 #     (scripts/build-all.sh) fails, a loud warning is printed naming the
 #     possible canary contamination instead of failing silently.
+#   - DS-148 (Tier-3 Skeptic Minor x2, hardening the case (m)/(n) scaffold):
+#     both cases now route through the shared `_with_mutated_source` helper
+#     instead of each carrying its own backup/mktemp/trap/restore block, and
+#     the mutating region is guarded by a cross-process `mkdir`-based lock
+#     (`_mut_lock_acquire`/`_mut_lock_release`) so two concurrent runs of this
+#     suite against the SAME checkout can no longer corrupt the tracked
+#     source permanently (run B backing up run A's already-mutated file, then
+#     writing that corruption back as "restored"). The lock is keyed to
+#     REPO_DIR, so separate worktrees never contend with each other - only
+#     two processes racing the same checkout do. Case (m)'s no-rebuild
+#     restore and case (n)'s rebuild-with-soft-fail-warning restore are both
+#     preserved via the helper's optional restore-hook argument.
 # ---------------------------------------------------------------------------
 set -uo pipefail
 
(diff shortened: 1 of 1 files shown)


tyson-solara6:

**Blocks merge (independent of code review):** this PR currently shows `mergeStateStatus: DIRTY` / `mergeable: CONFLICTING` against `main` - needs a rebase before it can land.

**Code review:** the mkdir-based lock + backup-after-acquire reordering in `.claude/tests/install-converge.test.sh:155-236` is sound and correctly closes the concurrent-run corruption race described in the PR. Two very narrow, non-blocking notes for awareness:
- A signal arriving in the brief gap between the explicit `_mut_restore_now` call and the trap reset back to `_cleanup` could cause a stale trap to invoke `_mut_restore_now`/`_mut_lock_release` a second time, potentially freeing a lock a different process has since legitimately acquired. Vanishingly unlikely, and it's a pre-existing restore-then-reset-trap pattern, just now compounded with lock semantics.
- The lock path is keyed to `${TMPDIR:-/tmp}` + `sha256(REPO_DIR)`. If two concurrent invocations run under different `TMPDIR` values (e.g. divergent CI sandboxing), the lock silently provides no protection. Worth a doc caveat.

Nothing else found; the case (m)/(n) restore-hook behavior is preserved correctly.

tyson-solara6:
Superseded by the rebased branch: this one predates #604 and its stale-lock reclaim was a non-atomic check-then-delete that a Skeptic demonstrated two waiters could both pass. Closing rather than force-pushing over it.

PR #614, opened 2026-08-09
rename(content): bin/agentic-* to bin/ds-* in content/references

tyson-solara6:
Renames `bin/agentic-*` command references to `bin/ds-*` in `content/references/**` - the last content chunk. **102 occurrences across 12 files.**

With this, the `content/` tree is complete. The only remaining `agentic-*` tool references anywhere under `content/` are a single `agentic-factory` (a sibling repo's track name, not a tool here) and that residue is now pinned by an executable assertion rather than prose.

## Verification of the rename itself

Round 1 verified it by line-set normalization: normalize the tool names on the 94 removed lines, set-compare against the 94 added lines, and confirm **zero only-in-minus and zero only-in-plus**. No collateral change, no dropped line, no near-miss identifier. `ae_base_branch_sync` (a bash function) correctly untouched while `bin/ds-base-sync` beside it was renamed.

The pattern-coupling check that caught real defects in three prior units came back clean here: full `codex-skills.py` inventory at base and head, set-compared after normalizing - no pattern silently stopped matching.

## Where the review effort actually went

The rename was clean in round 1. Three further rounds went into a single `AGENTS.md` bullet describing which build scripts hardlink which files - and that bullet was wrong three consecutive times, each version written as the fix for the previous one's wrongness, each wrong about something different: the script list, then the file list, then the count and the mechanism.

The fix was to stop correcting it. The prose census is deleted; what remains is the derived instruction - grep the target `build.sh` for `hardlink_from_content(`, a bare `ln ` without `-s`, or `os.link(` - which was independently verified to find every call site, with no hardlink creation anywhere outside `*/build.sh`.

A test now backs it, and the last round was spent making that test's own description honest. It initially claimed to derive the call-site set "via this same grep" and to redden when a script gains or loses a hardlink call. Both were false: its patterns miss every real `hardlink_from_content "$src"` call site (it was matching the helper's *definition*), and under CI conditions - always a fresh checkout, where git has not preserved hardlinks - a script losing its call passes green.

The description now names two demonstrated guarantees and two explicit non-guarantees, each verified by constructing and running the case rather than reasoning about it.

## Also fixed

- The residue assertion masked protected tokens by prefix, so `agentic-engineering-doctor` slipped through silently. Now an exact-set membership test, proven by injection.
- The test's hardcoded destination census is gone - Set B is derived by `find` + inode comparison, which needs no path convention and so cannot break when Unit 6 renames the skill directory.
- Grep patterns no longer match commented-out lines.
- `content/project-scaffolding.yml`'s manifest renamed to `bin/ds-migrate`; its applied content is deliberately untouched, since `bin/ds-migrate` reads that file from the DinoStack install and never copies it into consumer projects.

## Known follow-up

The test's two sides are still not fully independent, and it cannot detect a hardlink call gained or lost within a single CI run. Closing that needs either a double build per adapter or a committed inode fingerprint - real design work, deliberately left out of this unit's scope and worth its own ticket.

## Review rigor

- **Brief / Plan path:** `docs/planning/ds-rename-program/` (Plan tier, 7 units; gitignored, local-only). Unit 5 block under `architect-plan.md` §spawn_brief_must_state.
- **Skeptic rounds (tier):** 4 rounds, Tier 2.
- **Findings summary:** R1 - 1 Major, 4 Minor. R2 - 1 Major, 1 Minor. R3 - 1 Major, 4 Minor. R4 - closing. Every Major after round 1 was the same `AGENTS.md` bullet, resolved by deleting the census rather than correcting it a fourth time. The rename itself carried no Major at any point.

## North Star alignment

**Advances "produce verifiable outcomes autonomously."** A prose claim that had gone stale three times is replaced by an executable check, and the claim that remains is scoped to what that check actually demonstrates.

**Advances "low friction."** The methodology's reference layer now names one CLI prefix throughout.

**No pillar regression.** Compat aliases keep every documented invocation working.


tyson-solara6:

**Blocks merge (not a logic problem in this diff):** `mergeStateStatus: DIRTY` / `mergeable: CONFLICTING`, confirmed genuine - this branch is 2 commits behind `origin/main` (#611, #612), and #612 edited the same generated-adapter files this PR's build regenerated. Needs a rebase onto current `main` followed by `bash scripts/build-all.sh` before merge, per the guidance this PR itself adds to `AGENTS.md`.

**Code review:** verified mechanically rather than by eyeballing - normalized line-diffing every changed `content/**` file to confirm no missed/over-rename, grepped for any leftover `agentic-` token in added lines (only the intentionally-untouched `agentic-factory` sibling-repo reference remains, as documented), byte-diffed `content/references/*.md` and `content/project-scaffolding.yml` against all 8 adapter mirrors post-patch (all identical), and did a full clone + `scripts/build-all.sh` + `python3 -m pytest bin/tests/` run (1052 passed) plus confirmed the `bin/agentic-*` compat symlinks still resolve. No correctness issues or out-of-scope findings.

PR #616, opened 2026-08-09
fix(wrap): route /ds-wrap Part B durable facts to the canonical MEMORY.md (DS-90)

tyson-solara6:
PR 2 of 2 for DS-90. Retargets `/ds-wrap` Part B to write the canonical `MEMORY.md`, bounds its growth, and repairs the shipped docs that described the old routing.

## The defect

`/ds-wrap` Part B - the command whose entire job is session-end durable-fact capture - wrote to `<cwd>/.agentic/memory.md`, which is gitignored, `@`-imported by nothing, and read by no agent at session start. The canonical `<cwd>/MEMORY.md` that **is** `@`-imported got nothing from it.

Two measurements made it concrete:

- **149 entries** sit stranded in the staging file, accumulated over seven weeks by synchronous `/ds-wrap` - in the repo that authors this methodology.
- `.agentic/compression-state.json` tracked **only** the staging file. Part E - the sole outflow mechanism in the methodology - had been spending Worker+Skeptic compression effort on a gitignored scratch file nothing reads, while the always-loaded file went uncompressed and ungated.

## What changes

| area | change |
|---|---|
| Part B target | `<cwd>/MEMORY.md`, capped 3/run; `.agentic/memory.md` becomes deferred-wrap **daemon staging**, drained by Part B |
| Part E targets | root `MEMORY.md` added as **primary** - its size is per-session resident cost |
| Reachability | five sites that let a run skip the drain are now staging-aware, each citing `bin/tests/reach_model.py` by invariant |
| `sid` scheme | staged `s0…` / fresh `f0…` prefixes - see below |
| Doc-sync | `conductor-operating-rules.md` lock scope, `conventions.md`, `conventions-detail.md`, `delegation-detail.md`, `capture-classification.md`, two agent manifests, `docs/index.html`, one slide deck, `docs/wrap-deferred.md` |

**A live defect found and fixed along the way.** `drain()` validates `sid` uniqueness across staged **and** fresh entries combined, but the contract only ever defined a scheme for staged ones. A Worker numbering fresh entries from their own 0-based index collides `"0"` with `"0"` on run one, `drain()` raises, staging is left untouched - and every subsequent run aborts identically. A permanent deadlock. Separately, `_validate` tests truthiness, so an integer index `0` raises a misleading "missing required 'sid'". The `s`/`f` prefix scheme fixes both, and it is the scheme `test_drain_invariants.py` already assumes - so no merged file needed an edit.

## Discharges `AGENTS.md`'s module-wiring rule

PR 1 merged `reach_model.py` and `drain_model.py` with **zero** `content/` callers. This wires both: `reach_model.py` via the R1 citation in Part B Step 0 and R2 at the three Part E skip sites; `drain_model.py` via Step 0's normative drain reference and Output 2's `sid` contract.

## Review

Skeptic returned 0 Critical, 4 Major - all consequences of the retarget opening the zero-substance route to writing memory. All fixed:

1. That route's confirmation still mandated a literal string saying nothing was captured, while Part B was writing up to 3 entries. Now conditional.
2. **Part G skipped on that route** because "no session activity happened" - so the write sat uncommitted, and the next run was genuinely zero-substance and skipped again. No `/ds-wrap` run would ever commit it. Part G's skip now reuses Part B's own skip predicate rather than branching on route.
3. A new lock-scope enumeration claimed the lock guards "exactly" three files, omitting `decisions.md`, `.agentic/memory.md`, and `.agentic/memory-pending.md`.
4. `conventions.md` still described `.agentic/memory.md` as `/ds-wrap`-internal scratch.

Conductor verification then caught a fifth: the corrected lock-scope paragraph **contradicted a pre-existing sentence four lines above it** asserting the lock guards "exactly one thing." Shipping that would have put a live self-contradiction into the file that documents contradiction-resolution. Reconciled by pointing the older sentence forward to the non-exhaustive list rather than duplicating it - two enumerations in one file is how the defect arose.

The Skeptic also verified the engineer's one disclosed judgment call against the merged model and found it correct: `reach_model.py` explicitly forbids `s`/`e` terms in `ROUTE_PREDICATES`, so the zero-substance disqualifiers belong in that route's downstream logic, not its trigger.

## Verification

- **1,051 pytest** green; **47/47** hooks-JS tests green including `test-wrap-write-discipline.js`
- Adapter drift clean using the pathspec copied verbatim from `adapter-sync.yml`; all 4 skill symlinks relative
- Resident budget OK, `THRESHOLD` untouched; no `content/sections/**` edits so no baseline regeneration
- `grep -cF 'Output 3 (AGENTS.md updates)'` = 3 - the reachability fix that closes a prose/model divergence where the shipped criteria were **more permissive than the model**
- DS-128's four `aw-site:` markers intact
- `docs/index.html`'s `Write(**/memory*.md)` / `Edit(**/memory*.md)` globs **byte-identical** - deliberately not broadened

## North Star alignment

**Advances "produce verifiable outcomes autonomously."** A durable fact learned in a session now lands in the file every future session actually reads. Before this, session-end capture wrote to a store nothing loads - the autonomy was real but the outcome was unverifiable, because the artifact was unreachable.

**Advances "guard operator attention."** Part E now bounds the `@`-imported file, so the fix does not trade a routing bug for unbounded resident growth. The five reachability fixes mean the drain cannot be silently skipped.

**Works for everyone.** The routing is per-project and resolves at runtime; the DinoStack-repo carve-out (its own `MEMORY.md` is gitignored) is preserved, exactly as `wrap-ticket` already handles it.

**One attention cost, deliberately not paid for the operator.** `docs/index.html`'s published permission globs were left unbroadened. Adding `**/MEMORY.md` would silence overwrite prompts on operators' committed knowledge file - which Part E can now rewrite wholesale - in a recipe they paste into their own settings. The prompt is a guard worth keeping by default; the prose now says so and shows how to opt out. You cannot un-paste someone's settings.

## Known limitations

- The 149-entry backlog drains at 2-3/run: **50-75 wraps**. Bulk migration was descoped after five review rounds without convergence; it returns as its own ticket.
- The drip is inert on a fresh default-config project (`deferred_wrap_daemon` defaults `false`) - load-bearing only for backlog-carrying and daemon-enabled projects. Stated in the shipped prose, not concealed.
- Part E compression rewords entries, so the append-writers' substring dedup can re-append a fact after compression. Accepted knowingly: the status quo is compressed *never* and growing monotonically. A follow-up ticket owns hardening the dedup.


tyson-solara6:
Superseded by a rebase onto current `main` (past #614's `bin/agentic-*` -> `bin/ds-*` rename in `content/references`, which conflicted with this branch's lock-scope rewrite). Same work, replayed and re-verified on `feat/ds-90-wrap-memory-routing-v3`. Closing per the close+rebase convention rather than bundling.

PR #618, opened 2026-08-09
refactor(DS-split): extract Tracker Writeback Helper into content/references/tracker-writeback.md

tyson-solara6:
## Summary

DS split unit 1 of 6 (context-economy split of `content/commands/ds-implement-ticket.md`). Moves the `## Tracker Writeback Helper` block (invocation contract, forward-only guard's category-rank/pipeline-sub-rank algorithm, diagnostic-enrichment sub-step, failure/skip logging formats - lines 480-548 of the pre-move file) out of `content/commands/ds-implement-ticket.md` into a new `content/references/tracker-writeback.md`, behind a trigger-pointer. Zero loss of effectiveness - the move is byte-identical for 65 of 66 lines; the one non-verbatim line carries three deltas: a pre-existing em dash normalized to a hyphen per repo convention, plus two dangling-deictic repairs ("below" -> the explicit file citation) made necessary by content that moved out of the file during the split.

- Leaves a trigger-pointer at the source location containing the literal path `content/references/tracker-writeback.md`.
- Updates all 10 in-file `## Tracker Writeback Helper` cross-references (Setup, Phase 2c, Phase 11, and the W1-W7 call sites) plus the 3 external citers (`ds-ticket-status-sync.md`, `ds-wrap.md`, `ds-init-project.md`).
- Repoints `bin/tests/test_tracker_writeback_ranking_spec.py` and `bin/tests/test_tracker_dev_complete_spec.py`'s `canonical_block` fixtures and moved-content assertions at the new file; content that stayed inline (Setup, Phase 2c, Phase 11's own Inputs list) still reads the command file.
- Adds a pointer-propagation regression test across the 7 command mirrors (`test_command_mirrors_carry_pointer_to_helper_reference`) - `check-adapter-sync` diffs regenerated-vs-committed content from the same build run, so it cannot catch a build regression that silently drops the pointer; this test can.
- Replaces the `.hermes/SKILL.md` byte-identity check with a containment check, since that aggregate wraps reference docs in `### ` (not `## `), which over-runs the shared `_extract_block()` heading-terminated extraction.
- Adds the module-manifest header to the new reference file per `content/rules/module-manifest.md`.
- Bumps the reference-doc count (34 -> 35) across `README.md`, `CONTRIBUTING.md`, `ADAPTERS.md`, and `docs/slides/contributing-slides.md` (+ rebuilt `.html`) per the docs-currency obligation.

**Post-review fixes (this round):**
- Repointed `POINTER_TEXT` in `test_command_mirrors_carry_pointer_to_helper_reference` from the bare `content/references/tracker-writeback.md` path (which recurs 8x per mirror, 14x in `.hermes/SKILL.md`, via unrelated cross-references inside the moved block - vacuous, could never go red for the regression it exists to catch) to the extraction-site-unique literal `Full reference (invocation contract`. Mutation-tested both failure modes: deleting the extraction-site pointer line -> red; deleting the entire helper region -> red; restored -> green.
- Repaired two dangling deictics in `content/references/tracker-writeback.md` left over from the split ("7 new sites below" and "see the edits to those files below") that pointed at content no longer present in the file.
- Corrected the `ADAPTER_PATHS` comment's hardlink claim: same-inode identity only holds after `build.sh` has run locally - a fresh clone or worktree checkout gives `content/references/`, `.cursor/`, `.gemini/`, and `.copilot/` four distinct inodes carrying an identical git blob until rebuilt. The list still asserts over exactly one independent build-time surface either way.

Doc-sync: predicate not triggered (no reality-asserting change beyond the already-attested 34->35 reference-doc count bump from the original commit).

## North Star alignment

This change is a pure prose relocation behind a progressive-disclosure trigger-pointer - it reduces the always-resident cost of `content/commands/ds-implement-ticket.md` for every session that doesn't need the Tracker Writeback Helper's algorithm detail, while keeping it fully available on read-on-trigger for the sessions that do (Phase 11, W1-W7, `/ds-ticket-status-sync`, `/ds-wrap` Part F). It works for every project regardless of tracker configuration - the moved content and its call sites are identical in behavior before and after, verified by `bin/tests/test_tracker_writeback_ranking_spec.py` and `bin/tests/test_tracker_dev_complete_spec.py` staying green (93/93 and 4 previously-broken tests fixed) plus mutation-testing every new/modified assertion.

## Test plan

- [x] `python3 -m pytest bin/tests/ -q` - 1079 passed (one identity/lock race test flaked in the full run, confirmed passing in isolation and unrelated to this diff)
- [x] All `bin/tests/test_*.sh` shell tests - pass
- [x] All `hooks/tests/test-*.js` (incl. the two wrap-lock tests) - pass
- [x] `bash scripts/build-all.sh` (11 adapters) then `git diff --exit-code -- <adapter dirs>` - clean, idempotent
- [x] `python3 scripts/codex-skills.py inventory --repo .` regenerated + `.codex/build.sh` - clean
- [x] `scripts/check-symlinks-relative.sh` - all 4 skill symlinks stay relative
- [x] `scripts/check-command-file-budget.sh`, `check-resident-budget.sh`, `check-skill-embed-budget.sh` - all OK
- [x] `git diff origin/main...HEAD -U0 | grep '^+.*—'` - zero added em dashes
- [x] Mutation-tested every new/modified test assertion (pointer-propagation test, hermes containment test, byte-identity split, and the repointed `POINTER_TEXT` anchor): broke each on purpose, confirmed red, restored, confirmed green
- [x] `python3.12 scripts/prose-move-impact.py` run against a pristine pre-move checkout, confirming the 9 breaking assertions + 6 doc citations it flagged are exactly what this PR resolves (verified post-hoc by grep + the green test suite, since the tool's `--range` argument is only meaningful against the pre-move tree; NOTE: this script returns `Verdict: FAIL` even against `origin/main` since UNRESOLVED consumers count as failures by design - it is a pre-move consumer inventory, not a pass/fail gate)


tyson-solara6:
Superseded by #620. This branch conflicted irrecoverably with the `agentic-engineering` -> `dinostack` skill rename (#619) across 31 files; force-push is denied in this repo, so the work was re-applied on a fresh branch rather than rebased. #620 carries the identical twice-reviewed change plus the three Minors from this PR's round-2 review.

PR #619, opened 2026-08-09
rename(skill): agentic-engineering skill and directory to dinostack

tyson-solara6:
Renames the skill name and directory from `agentic-engineering` to `dinostack` across 7 skill directories, 6 build scripts, the installers, the Codex generator, a required CI gate, the runtime activation hook, and `content/` prose. Final unit of the `ds-*` rename program.

Motivation: skill names occupy a flat per-user namespace with no vendor scoping, so a third-party skill named `agentic-engineering` would silently collide. `dinostack` is also shorter, so it is budget-favorable.

## The boundary that defines this unit

Only the skill's **install-path identity** changes. These deliberately do not:

- `~/.claude/agentic-engineering.json`, `~/.agentic/agentic-engineering-config.json` (config filenames)
- `agentic-engineering: opt-in` / `opt-out` / `-profile:` / `-preset:` (activation markers)
- `<!-- BEGIN/END managed-by-agentic-engineering -->` (managed-block delimiters)
- the `_ae_stale_*` allowlist contents, and `agentic-factory`

Every one of those lives in a file on the operator's machine or in a consumer repo we do not control. Renaming any of them changes behavior in projects nobody edits.

## What review caught, and why it matters

**Three of the protected literals were renamed anyway,** by stray sed hits, and each would have failed silently:

The **managed-block delimiters** were renamed in three adapters' install and uninstall paths. Since existing installs carry the old delimiter, upgrade would take the append branch: a second managed block added, the first stranded with `@skills/agentic-engineering/...` imports pointing at a directory the same run deletes, the `skill_auto_load` migration disarmed, and uninstall unable to remove it. Confirmed live on this machine - `~/.claude/CLAUDE.md` still carries the old marker. Resolved by keeping the old literal, matching the precedent already set by the activation markers.

The **activation marker** `agentic-engineering:` was renamed to `dinostack:` in `docs/configuration-reference.md` and `ds-init-project.md` - including, on the next pass, in `content/sections/01-activation-preflight.md`, the file that *defines* it. An operator following the public reference would have written a marker nothing reads, and on update an existing `opt-in`/`opt-out` inside a managed block would no longer be preserved, silently flipping activation state.

**The `skill_auto_load` migration** matched only the current skill-path literal, so a genuinely pre-rename block on disk could never trigger it - the one population it exists for.

## Self-heal, built here

Unit 0 built the `~/.local/bin` self-heal that made the binary rename survivable. Nothing equivalent existed for skill directories, so this unit adds per-adapter stale-directory pruning, in two shapes:

- **Symlink-destination** adapters remove the symlink node after an ownership check. Unlinking never traverses the target.
- **Real-directory** adapters use a new `scripts/lib/prune-stale-skill-dir.sh` with an ownership predicate: every entry must be a symlink resolving inside `$REPO_DIR`, or one of the specific filenames that adapter's installer generates. Any unrecognized entry aborts and leaves the directory. Removal is `find -mindepth 1 -delete` then `rmdir` - never `rm -rf`.

The predicate initially keyed on the checkout's *name* rather than `$REPO_DIR`, fitted to a fixture called `fake-DinoStack` - which meant it would never fire for the oldest installs, whose checkouts were literally named `agentic-engineering` and are the most likely to have a stale directory. Now compared against `realpath($REPO_DIR)`, with the prefix-but-not-child case verified refused.

## Tests that had to be made real

The wiring assertions added to prove each prune is *invoked* rather than merely defined matched a `command -v` definition check - so they passed on exactly the state they existed to reject, in all seven adapters. Now tightened to the invocation, with per-adapter RED reproduced.

A guard was added because the converge suite deleted the real repo's `.git/hooks/pre-commit` - `HOME` sandboxing does not contain it, since `resolve_git_hooks_dir` uses `git rev-parse --git-path hooks`. The guard's first version added `INT TERM` to the trap, which made the interrupted path *worse*: `_cleanup` has no `exit`, so SIGINT ran the handler and the script continued, restoring the hook and then deleting it again over the remaining ~20 install calls. Corrected and proven by executing SIGINT, normal completion, and mid-suite failure. `.cursor`'s converge suite had the same unguarded defect and is now guarded too.

## Verification

- 1077 pytest; both converge suites green (57/0 and 20/0); zsh failure set identical to `main`
- Zero `managed-by-dinostack` repo-wide; every remaining `dinostack` occurrence audited individually and confirmed to be a skill-name, path, or brand-log use with no consumer matching it
- The completeness sweep used `git ls-files` plus a binary-safe reader, because `grep -rl` silently drops files it deems binary - which is how one file escaped the first pass
- `build-all.sh` then adapter drift clean AND `git status --porcelain --untracked-files=all` empty, confirming no old skill path resurrects as untracked files
- `codex-skill-sync` validates 4 skills under the new name; methodology baseline regenerated and drift-checked

## Review rigor

- **Brief / Plan path:** `docs/planning/ds-rename-program/` (Plan tier, 7 units; gitignored, local-only). Unit 6 block under `architect-plan.md` §spawn_brief_must_state.
- **Skeptic rounds (tier):** 4 rounds, Tier 3 (irreversible on-disk user state, 11-adapter blast radius).
- **Findings summary:** R1 - 1 Critical, 2 Major, 1 Minor. R2 - 2 Critical, 2 Major, 2 Minor. R3 - 2 Major, 3 Minor. R4 - closing. Every Critical was a protected literal renamed by a stray sed hit; each was invisible to CI because the fixtures had been renamed alongside the code.

## Known follow-ups

- `bin/tests/test_hooks_snapshot_migration.sh` and `bin/tests/test_local_bin_ds_prefix_install.sh` touch the real pre-commit hook through unguarded `install.sh` calls. Pre-existing, self-documented in the former's own manifest.
- No uninstall path removes a pre-rename skill directory; the prune is install-side only, so uninstalling directly from a pre-rename install leaves an orphaned symlink.
- `scripts/lib/precommit.sh` re-points the real hook at a worktree path when run from inside one, which dangles when the worktree is removed. This fired repeatedly across the program.

## North Star alignment

**Advances "works for everyone."** The collision hazard is closed, and the self-heal means no existing install is left with an orphaned skill directory. The three protected-literal reverts are the concrete version of that promise - each would have changed behavior in projects the operator never touched.

**Advances "low friction."** One name for the skill across every adapter.

**Trade-off, stated plainly:** the skill directory is now `dinostack` while the config filenames, activation markers, and managed-block delimiters remain `agentic-engineering`. That inconsistency is deliberate - those literals live in files we do not control - but it is a real seam a future contributor will trip over, and it is now recorded in the plan's boundary table rather than left to be rediscovered.


tyson-solara6:
Automated review note (no formal verdict — self-authored PR):

This PR's own diff to `hooks/stop-context.js` is a one-line label rename (`agentic-engineering` → `dinostack`) and looks fine.

However, the **CodeQL check on this PR is failing** with a high-severity "insecure randomness" alert at `hooks/stop-context.js:863` (`session_uuid: sessionId || generateUuid()`, where `generateUuid()` uses `Math.random()`). That line is **pre-existing code this PR did not touch** — it's just in a file this PR happens to edit. `session_uuid` here is a local telemetry identifier, not used for auth/session-security purposes, so this is very likely a false positive for this specific use, but it's worth a deliberate look (or an explicit CodeQL suppression with rationale) rather than leaving the check red.

Could not auto-file a Jira ticket for this — the mcp-atlassian Jira MCP tools were not available in this run's environment. Flagging here instead so it isn't lost.

PR #340, opened 2026-06-27
refactor(methodology): progressive-disclosure kernel split, rebuilt on current main (supersedes #314)

tyson-solara6:
## What

Progressive-disclosure kernel split of the always-on methodology. Trigger-only detail moves out of the always-loaded `content/sections/**` and `content/rules/**` into `content/references/**` behind read-on-trigger pointers; the governance kernel stays inline.

**Assembled METHODOLOGY body: 123,698 -> 84,741 chars (~31.5%).** Plus `module-manifest.md` delisted from 5 always-on adapter surfaces and `code-standards`/`conventions` rules shrunk - total always-on cut is larger.

## Why

Every session loads the full methodology before the first tool call, tripping the ">40k chars will impact performance" warning and driving mid-ticket auto-compaction. This cuts the always-on cost while keeping every governance rule that fires before the agent decides to read anything.

## What stays inline (kernel)

Risk tables, the Direct-OK/Spawn-Worker decision table, default-and-proceed, the finding-verification guardrails, the Open-Questions hard gate, activation steps 1-4, the QA-fires/diff-read rules, the Tier/role-model one-liners. Partition rule codified in `design-goals.md` Goal 4 (three-question test).

## Scope

- Shrink sections 01, 02, 04, 07, 08, 09, 12 and rules `code-standards`/`conventions` to kernel + pointers
- 7 new `content/references/*` files (verbatim relocation)
- Fix dead `SKILL.md` + `.codex/skill/SKILL.md` -> `agent-methodology.md` references
- Delist `module-manifest.md` from 5 always-on adapter surfaces (incl. the Hermes glob-skip)
- Rebuild all 10 adapters; regenerate `scripts/.methodology-baseline.sha256`
- Sections 03/05/06/11 left as-is (already kernel-sized or their reference files are supplements, not full copies - shrinking would drop governance)

## Verification

- Architect plan + 3 design-review Skeptic passes (original effort) + a fresh integration Skeptic on this diff: governance preserved, relocations byte-faithful, all pointers resolve, and teammate feature #256 (cross-harness teams + role-model routing) content intact
- Bidirectional no-orphan scan: 0 forward / 0 backward
- All 10 adapter builds exit 0; `check-methodology-drift`, `check-codex-skill-sync` pass; `adapter-sync` clean
- 8 commits, all DCO-signed, no merge commits

## Note

[...]

PR #624, opened 2026-08-09
refactor(DS): extract QA loop state (Phase 6b/8.5) into content/references/qa-loop-state.md

tyson-solara6:
## Summary

Unit 5 of 6 in the `content/commands/ds-implement-ticket.md` progressive-disclosure split (Units 1 and 6 already landed: `tracker-writeback.md`, `handoff-evaluation.md`).

- Extracts **Phase 6b: QA Gate (conditional)** (was lines 1978-2111 on `main`) and **Phase 8.5: QA evidence (conditional)** (was lines 2304-2462 on `main`) into a new `content/references/qa-loop-state.md`, following the retained-heading + pointer-paragraph pattern from Units 1/6.
- The `site: W3` tracker-writeback structured tag is kept verbatim in the kernel command file (it is located there by an unguarded `text.index()` in `bin/tests/test_tracker_lifecycle_sites_spec.py`), with a one-sentence pointer immediately adjacent to it, plus a new co-location regression test that would fail if the two are ever separated.
- `content/commands/ds-implement-ticket.md` size after round-2 fixes below: 322,971 B, well under the 371,000 B budget (48,029 B headroom). This grew back up from the round-1 figure of 320,264 B because round-2 restores executable spawn-contract/telemetry-emit paragraphs and a Tracker Writeback Helper caller-enumeration block to the kernel - see "Round 2 fixes" below.

## Companion edits

- `bin/tests/test_ticket_rework_ledger.sh`: widened the SPEC scan (new `_present_any` helper) for the three `QA_STATUS` literals whose sole source moved to `qa-loop-state.md`. `overwriting the Phase 6 state` was intentionally left untouched - it also occurs in Phase 6 proper, outside every move range.
- `bin/tests/test_tracker_lifecycle_sites_spec.py`: added `test_w3_tag_colocated_with_qa_loop_state_pointer`.
- `bin/tests/test_tracker_dev_complete_spec.py`: added `content/references/qa-loop-state.md` to `EXPECTED_TRACKER_STATE_FILES` (the W5 BLOCKED-target citation that moved with the body).
- `bin/tests/test_prose_move_impact.py`: `test_regression_pattern_slot_literal_not_dropped_by_meaningfulness_filter` depended on the (now genuinely extracted) Phase 6b body still containing 3 of `mark-blocked-and-continue`'s occurrences inside its `NINE_RANGES` fixture. Repointed that one test to `## Batch state contracts (binding)` (still un-extracted, still contains the literal) so the regression coverage stays real rather than stale.
- `bin/tests/test_loop_state_site_coverage.sh` already carries the shared gate-body widening from Unit 6 (confirmed present, not re-edited).

## Pre-move consumer inventory

Ran `python3.12 scripts/prose-move-impact.py` pre-move. 11 BREAKING rows reported; disposition:
- 4 resolved by the companion edits above (3 `QA_STATUS` literals + `site: W3`).
- 7 were false positives of the tool's single-match resolution - each literal (`$BRANCH_NAME`, `mark-blocked-and-continue`, `overwriting the Phase 6 state`, `Phase 0`, `forward_only_guard: true` x2, `regardless of whether`) has other occurrences outside both move ranges, confirmed by direct grep against the live file before committing.

## Docs-currency

Reference-doc count 36 -> 37: updated `README.md`, `CONTRIBUTING.md` (incl. alphabetical enumeration), `ADAPTERS.md`, `docs/slides/contributing-slides.md` (both sites), rebuilt slides via `bash scripts/build-slides.sh`.

Doc-sync: triggered (new reference doc, command-file structural change) -> updated README.md, CONTRIBUTING.md, ADAPTERS.md, docs/slides/contributing-slides.md + regenerated .html.

## Round 2 fixes (Skeptic findings on this PR)

Multiset-diff audit method (net occurrence totals are unfalsifiable against substitution - a multiset diff over `.codex/skill-compatibility.yml`'s `(source_token, resolution_mode)` pairs is required instead) surfaced that Unit 1 (`tracker-writeback.md` extraction, already merged to `main`) had shipped live Codex breakage, and that this PR's own `qa-loop-state.md` extraction dropped two more executable spawn-contract literals into a file `scripts/codex-skills.py`'s transform never scans (`content/references/**` is a symlinked resource tree, not a transform input - only `content/commands/*.md`, `content/SKILL.md`, and assembled `METHODOLOGY.md` are).

- **MAJOR 1 (Unit 1 regression, fixed in-scope here per the same precedent as Unit 6's fold-in):** restored the 4x `/ds-wrap` + 4x `` `/ds-ticket-status-sync` `` + 1x `.agentic/tracker-states.json` caller-enumeration tokens that Unit 1's extraction dropped from the transform-visible kernel, as a new "Caller enumeration" bullet list in `content/commands/ds-implement-ticket.md` §"Tracker Writeback Helper". `content/references/tracker-writeback.md`'s own copies now point back to that kernel list instead of repeating the raw tokens. One remaining single-occurrence delta (`content/commands/ds-implement-ticket.md` cited from `content/commands/ds-wrap.md`, 6->5) is a legitimate, not-a-defect change: that citation now correctly points at `content/references/tracker-writeback.md`, the algorithm's actual current location.
- **MAJOR 2 (gate blindness):** added `test_qa_loop_spawn_contract_paragraph_pinned_in_kernel` to `scripts/test/test_codex_skills.py`, which pins the kernel source directly (independent of the reference-file allowlist mechanism the existing gate uses). Mutation-tested: reverting `content/commands/ds-implement-ticket.md` to the pre-restoration `6ac8871c` revision turns it red; restoring the fix turns it green.
- **MAJOR 3 (`Agent` -> `spawn_agent` rewrite dropped):** restored the QA-loop telemetry-emit paragraph (the two `Agent` tool-call bracket sentences for qa-engineer Step 1 and fix-engineer Step 4) to the kernel as "Telemetry emit (V1) - QA loop", and removed the redundant raw `isolation: "worktree"` literal from `qa-loop-state.md`'s own Step 1 (which already deferred to the kernel but was still repeating the literal). `qa-loop-state.md` now carries zero raw `isolation:`/`Agent`-tool-call literals outside its manifest-comment description of what the kernel retains.
- **Minor - Unit 6 restoration incomplete:** `content/references/handoff-evaluation.md`'s own manifest note claimed the two resume-banner literals were "kept in the kernel command file (not duplicated here)", which was false - both literals are also printed inline inside the file's own illustrative print-example fenced blocks. Corrected the note to state the kernel is the source of truth and the inline copies are for print-example readability only.
- **Minor - inaccurate "verbatim" claim:** the round-2 restorations above are close paraphrases of the pre-extraction prose, not byte-for-byte verbatim; this PR body does not claim verbatim restoration for them.
- **Minor - stale PR body:** this body was regenerated (byte count, fixture-repoint target) rather than hand-patched.
- **Minor - four-condition predicate for Units 2/3/4:** the round-2 finding states this predicate exists only in an unpushed local commit message from the reviewing session. It is not present in any commit reachable from this branch, `main`, or any other local ref checked (`git log --all`), and `docs/planning/` is gitignored and not visible to a worktree-isolated engineer per this repo's own convention (`AGENTS.md` "docs/planning/ is gitignored..."). Could not locate or restate it here - flagging for the conductor to supply from its own session context before Units 2/3/4 proceed.
- **Residual uncited cross-file deictic** at `qa-loop-state.md`'s Step 4 ("apply the ... subsection in Phase 6") was left as-is: it already names the target subsection ("Batch-mode escalation routing (mark-blocked-and-continue)") by full title one sentence earlier in the same paragraph, so "that subsection in Phase 6" is a same-paragraph back-reference, not a dangling cross-file deictic requiring a separate fix.

## North Star alignment

Continues the DS-143 progressive-disclosure kernel split (PR #340 pattern): keeps the always-loaded resident set lean while pushing trigger-only QA-loop mechanics into a read-on-trigger reference doc, without changing runtime behavior - a pure prose relocation with mechanically re-verified consumers. Round 2 restores the small subset of that content Codex actually needs to execute (spawn-contract and telemetry-emit literals), keeping the trade-off honest rather than silently shipping Codex breakage for resident-set savings.

## Test plan

- [x] `bash scripts/prose-move-impact.py` pre-move inventory run and every BREAKING row disposed (see above)
- [x] Mutation-tested every new/modified assertion (co-location test, `_present_any` calls, `EXPECTED_TRACKER_STATE_FILES` entry, `test_qa_loop_spawn_contract_paragraph_pinned_in_kernel`): broken -> confirmed red -> restored -> confirmed green
- [x] `python3 -m pytest bin/tests/ -q`
- [x] All `bin/tests/test_*.sh`
- [x] `hooks/tests/test-*.js` (excl. wrap-lock, run separately); wrap-lock tests pass too
- [x] `bash scripts/build-all.sh` then `git diff --exit-code` over the adapter-sync pathspec - clean
- [x] `bash scripts/check-symlinks-relative.sh` - all 4 dinostack skill symlinks relative
- [x] `.codex/skill-compatibility.yml` regenerated
- [x] `bash scripts/check-command-file-budget.sh` / `check-resident-budget.sh` / `check-skill-embed-budget.sh` - all OK
- [x] `python3 scripts/test/test_codex_skills.py` (full run, ~7.5 min)
- [x] No added em dashes (`git diff -U0 origin/main...HEAD | grep '^+.*—'` on own diff)
- [x] `bash scripts/build-slides.sh` - idempotent, no drift
- [x] Multiset diff of `.codex/skill-compatibility.yml` `(source_token, resolution_mode)` pairs across 1e777841 -> 1ec0023c (Unit 1) and against `HEAD` re-run post-fix to confirm the 4x/4x/1x/1x MAJOR-1 tokens are restored


tyson-solara6:
Abandoning this extraction on measured evidence, after three review rounds.

The move lost 30 Codex-rewritten occurrences; three fix commits recovered 5. The residue violates the split plan's own predicate 24 times (15 condition-2 `.agentic/*`/`.claude/*` path rewrites, 9 condition-4 `$AE_REPO_DIR/` prefixes), with `content/references/qa-regression-obligation.md` driven to zero across the entire transform surface.

Economics had also inverted: the reference file (26,606 B) is larger than the 21,125 B removed from the kernel, net reduction 16,752 B with 20.7% clawback and climbing.

Conclusion: the QA-loop block is too densely woven into the Codex transform to extract without either breaking Codex or restoring most of it. This is a third distinct failure mode alongside the Q1 semantic test and gate scanned-set loss - recorded as transform density.

The three genuine repairs this PR contained (Unit 1's live regression fix, Unit 6's completion, and two gates) are preserved in #628.

PR #629, opened 2026-08-10
perf(ds-wrap): defer learnings.md read and parallelize Part E compression

tyson-solara6:
Two behavior-preserving latency fixes to `/ds-wrap`. Measured drivers, not guesses.

## What

**1. Defer the `.agentic/learnings.md` full read (Step 0 -> Step 1).**
The file is ~291 KB / ~73K tokens and was read in full at Step 0, *before* Step 0.5 decided the route. Only paths that spawn a draft Worker consume it. Light-path and zero-substance wraps paid the full read for content they never used. The read now happens immediately before the Step 1 Worker spawn, conditioned on "a draft Worker is being spawned" rather than on a re-derivation of Step 0.5's criteria (which would have duplicated criteria the file states are defined in exactly one place). The cheap existence probe stays in Step 0's batch.

Verified by grep that only two sites ever consumed the content: the Step 1 prompt embed and Part B step 3's dedup check. Three stale cross-references updated.

**2. Parallelize Part E's per-target compression.**
Targets are independent files with independent backups and independent `compression-state.json` sub-keys, but were processed strictly one at a time. When 2+ targets pass the gate, their compression Workers now spawn together in one message, and each target's Skeptic spawns as soon as that target's own Worker returns.

Concurrency is at the spawn/review layer only. Step 4's on-disk write sequence (backup -> snapshot -> prune -> overwrite -> state-update) stays main-agent-only and serialized both within and across targets. Every Worker brief, Skeptic brief, sign-off, and write now carries the target's absolute path so a sign-off cannot be applied to the wrong file. The 3-reroute cap is stated explicitly as per-target.

Both `MEMORY.md` and `.agentic/memory.md` are over-gate in this checkout today, so this halves Part E on the next qualifying run.

## Gates

- `scripts/build-all.sh`: 11 adapters clean; drift pathspec copied verbatim from `adapter-sync.yml`
- `pytest bin/tests/`: 1112 passed
- `bin/tests/test_*.sh`: 35 files, all passed
- hooks JS suite: 45 files passed, including `test-wrap-md-partA-parity.js` (19/0) and `test-wrap-context-format-golden.js` (20/0) - both byte-pinning tests green, so Part A's whole-flow lock and the merge-algorithm reference citation are untouched
- `check-command-file-budget.sh` / `check-resident-budget.sh` / `check-skill-embed-budget.sh`: OK
- `check-symlinks-relative.sh`: OK
- Em-dash check on branch diff: empty

`content/sections/**` untouched, so no methodology-baseline regen.

Doc-sync: predicate not triggered. Checked `docs/index.html`, all `docs/slides/*.md`, `README.md`, `CONTRIBUTING.md`, `content/SKILL.md` - none assert anything about Step 0's read timing or Part E's per-target sequencing; they describe `/ds-wrap` at a capability level only.

## North Star alignment

**Works for everyone (universality):** both changes are structural reordering of an existing flow. No operator identity, tracker, workspace, or local setup is referenced. A project with a small or absent `learnings.md` sees the same behavior it does today; a project with one over-gate compression target sees the same Part E it does today.

**Enforcement floors preserved:** neither change removes a gate. Part E's compression Skeptic still runs per target with its unchanged fact-loss brief and its 3-reroute cap. The whole-flow lock is untouched, as the two byte-pinning tests confirm. This is a latency change at the spawn layer, not a review-rigor change.


tyson-solara6:
Automated review pass (self-authored PR, no formal review verdict per policy).

**Skeptic findings:**

1. **Deferred `learnings.md` read (Minor residual risk):** No missed call site found — the PR's own pinned test (`test_learnings_md_read_is_gated_behind_the_draft_worker_spawn`) and the updated Part B step 3 dedup-check wording both confirm the read now happens at Step 1 for both draft-Worker spawn sites. The test only pins the markdown prose, not runtime behavior — an inherent limitation of prompt-as-code, not new here.

2. **Parallel Part E compression — `compression-state.json` race (Major):** Step 4(e) states the read-modify-write on `compression-state.json` "never races" because "step 4 is serialized," but the only documented *read* of that file is the one-time Step-0 snapshot used for gating. If step 4(e) writes back using that stale Step-0 snapshot for every target instead of re-reading the file's current state immediately before each write, target A's sub-key update gets clobbered when target B's step-4(e) runs after it. Worth spelling out explicitly that step 4(e) does a fresh read immediately before its write, not a reuse of the Step-0 snapshot.

3. **Minor:** No explicit check-and-reject step if a Worker's/Skeptic's self-reported `Target:` line mismatches its assigned path — only "never apply to a different target's file," with no verification gate.

**North Star alignment:** ALIGNED — latency-only, behavior-preserving change; reduces autonomous execution time without adding operator-facing surface area.

Recommend addressing the compression-state.json read-modify-write ordering (#2) before merge, since it's a plausible data-corruption path when 2+ targets pass the parallelization gate.

tyson-solara6:
Automated review pass (self-authored PR, no formal verdict per self-review rules). Adversarial read via Skeptic against the full diff.

**Findings: Major x2, Minor x1**

- **Major — escalation quiescing changes cross-target failure semantics without disclosing it as a behavior change.** In `content/commands/ds-wrap.md`'s new "Escalation quiescing" rule, once ANY concurrently-running Part E target exhausts its own re-route/format-reinvocation cap, step 1 now bars starting a new round for every OTHER target too — even a healthy sibling still within its own 3-re-route budget. In the prior sequential model each target's budget was fully independent. This is a real robustness regression (more compression-skip outcomes than before), not just a latency change, and the PR's "behavior-preserving" framing doesn't call it out. Worth an explicit decision: either decouple sibling failure from each other's cap, or document this as an accepted trade-off.
- **Major — `Target:` echo-back is validated for presence only, not equality to the dispatched path.** Part E's Worker/Skeptic briefs require a `Target: <path>` line, but step 3's validation only checks the line exists, never that it matches the path that target was actually dispatched to. With N Workers/Skeptics spawned together returning in arbitrary order, a response that echoes a sibling's path would still pass format validation — the only guard is an advisory "never apply..." sentence, not a mechanical equality check. Recommend adding an explicit path-equality assertion in step 3.
- **Minor** — doc-sync predicate plausibly trips (Part E's compression flow materially changed) but couldn't be independently verified without a live checkout; worth a grep pass over README/SKILL.md docs for stale sequential-only Part E descriptions.

Both are fixable without re-architecting; recommend addressing before merge given they touch this PR's own headline claim of being behavior-preserving.

tyson-solara6:
Closing in favor of #639, which re-derives this PR's uncontested half - deferring the `.agentic/learnings.md` read to the draft-Worker spawn - cleanly against current `origin/main`.

**Why split rather than continue here.** This PR bundled two independent changes. The learnings-read deferral drew zero findings across five review rounds and adds 1,112 B. The Part E compression-parallelism half drew one blocking finding per round: three of them were the same defect relocating rather than recurring (lock released too early -> non-total outcome enumeration -> non-total again on format-validation failure), then a report string asserting a reason that did not happen, then a silent deletion of the general non-quiescing report clause that only a whole-PR pass caught. It grew `content/commands/ds-wrap.md` from 123,091 to 136,140 B (+10.6%) - a file no budget gate measures - to guard a win on a path that fires only when 2+ compression targets are over-gate.

**The Part E work is recorded, not discarded.** Full failure history and three carry-forward constraints are in the per-path-locking plan: it is only safe once compression targets hold disjoint per-path locks with a guarded publish; the escalation-exit contract must give every in-flight sibling a terminal state by construction rather than by enumeration; and report strings asserting a specific reason are a recurring truthfulness hazard - derive the reason from the terminal state instead (that form also measured 237 B smaller than the string list it replaced). Captured as `.agentic/learnings.md` KNW-20260810-009.

Nothing here is lost - branch `perf/ds-wrap-latency` retains all six commits for reference.

PR #635, opened 2026-08-10
feat(hooks): guard worktree-isolated agent reads into the primary checkout (DS-150)

tyson-solara6:
## Summary

DS-150 Unit C: adds `hooks/enforce-worktree-read.py`, a PreToolUse(`Read`) hook that denies a worktree-isolated subagent's `Read` when the target resolves into the PRIMARY checkout instead of the agent's own isolation worktree.

The design is pinned by live payload capture (Claude Code v2.1.220, 4 records across 2 sessions), not inference:

- **Containment predicate** - `caller_root` from the payload's `cwd` field (never `CLAUDE_PROJECT_DIR`, which would compare the primary root against itself; never `os.getcwd()`, since the payload field is authoritative). `primary_root` from `CLAUDE_PROJECT_DIR`. All three operands (target, `caller_root`, `primary_root`) `realpath`-normalized before the test, because isolation worktrees live *inside* the primary root and an unnormalized prefix test is insufficient.
- **Caller discrimination** - main session vs subagent is decided by the **absence** of `agent_id`/`agent_type` (measured: subagent payloads carry 12 top-level keys, main-session 10, the difference being exactly those two). This matches what `enforce-shippable-edit.py` and `enforce-orchestrator-singularity.py` already do, so no new assumption is introduced.
- **Exemptions** - config-driven via `worktree_read_guard_exemptions` in `.agentic/config.json`, shipping empty. Zero built-in entries.
- **Kill-switch** - `AE_WORKTREE_READ_GUARD_DISABLE=1`. Fail-open on every error path.
- **Registered** with the guarded `test -f ... && python3 ... || exit 0` command form, matching the `enforce-turn-shape.py` precedent - an unguarded `python3 <missing>.py` exits 2, and exit 2 on PreToolUse blocks the call, which on a `Read` matcher would deny every read in every session including the conductor's.

Also wired into `bin/ds-doctor`'s `MANAGED_HOOK_BASENAMES` so an orphaned registration is diagnosed rather than certified healthy.

Adding the config key moved two repo-wide counts (behavioral toggles 21 -> 22, `enforce-*.py` hooks 8 -> 9), so this PR also syncs every asserting site and the gates that police them.

## North Star alignment

- **Works for everyone (universality).** The guard resolves everything from the hook payload and `CLAUDE_PROJECT_DIR` at runtime - no operator identity, workspace, tracker, or local path is baked in. The exemption list ships empty and is project-local; an `rclone.conf` exemption inherited from an upstream handoff was deliberately dropped after `grep -rn rclone content/` returned zero here, rather than being carried in as dead config. Fail-open plus a kill-switch means a project or harness where the guard does not apply degrades to prior behavior rather than breaking.
- **Guard operator attention.** The guard converts a silent, late-surfacing failure - an isolated agent reasoning against files outside its own branch, discovered days later as "why did this engineer act on code that isn't in its diff?" - into an immediate, actionable deny at the point of the read.
- **Produce verifiable outcomes autonomously.** The hook gates on payload facts only, never on inferred session capability, per `hooks/AGENTS.md`. Every assertion added here is mutation-verified rather than assumed, and the `_ENFORCER_SUBCOUNT_SITES` sweep now covers `hooks/lib/enforcement_log.py` - closing the specific gap that let a stale subcount survive two review passes.
- **No enforcement floor removed.** This adds a guard; it relaxes nothing. The flat prohibition that no hook may deny conductor `Read`/`Grep`/`Glob` to force delegation is preserved - the guard fires only on a genuinely worktree-isolated subagent, and the main session is exempt by construction.

## Test plan

- [x] `hooks/tests/test-enforce-worktree-read.py` - 35 assertions: main-session allow, non-isolated-subagent allow, own-worktree allow, cross-boundary deny, outside-primary-root allow, symlink normalization on all three operands, kill-switch, full fail-open matrix (absent/malformed `tool_input`, absent `cwd`, absent `CLAUDE_PROJECT_DIR`, unparseable stdin), config-absent/malformed, and configured exemptions including a segment-alignment negative case.
- [x] `bin/tests/test_install_worktree_read_guard.sh` - asserts the generated `Read` matcher carries the `test -f` guard; git-level sandbox so the test cannot repoint the primary checkout's `.git/hooks/pre-commit`.
- [x] `bin/tests/test_agentic_doctor.sh` Test 12 - a missing hook script is reported FAIL.
- [x] `bin/tests/` pytest: 1112 passed. `test_tracker_writeback_ranking_spec.py`: 64 passed, matching `main`.
- [x] `hooks/tests/test-*.py` / `test-*.js` / `test-*.sh` suites clean (`test-stdin-guard.js` fails identically on `main` - missing local `node_modules`, environmental).
- [x] `scripts/build-all.sh` + `git diff --exit-code` over the verbatim `adapter-sync.yml` pathspec: zero drift. `check-methodology-drift.sh`, `check-symlinks-relative.sh`, and the resident / skill-embed / command-file budget gates all OK.

Every assertion added or changed across four review rounds was mutation-verified: broken on purpose, confirmed red, restored, confirmed green.


tyson-solara6:
Automated review pass (self-authored PR, no formal verdict per self-review rules). Security-sensitive change (a containment hook), so this ran both a Skeptic pass and a dedicated security-auditor pass in parallel; findings below are reconciled across both.

**Findings: Major x1 (both reviewers agree), High-severity concerns flagged by security-auditor but assessed as likely-mitigated/Minor by Skeptic (reconcile before merge), Minor x2**

- **Major (both reviewers) — TOCTOU race between the hook's decision-time `realpath()` resolution and the actual `Read` tool's file open, undocumented.** `enforce-worktree-read.py` computes containment once at PreToolUse time; the actual file open happens afterward in a separate process. A subagent with concurrent Bash access could get an ALLOW for a path currently inside its own worktree, then symlink-swap that path to point into the primary checkout before the real open executes. This is inherent to any path-string-based PreToolUse hook, but it isn't mentioned anywhere in the module's "Accepted limitations" section, so it reads as an oversight rather than an accepted risk. Recommend adding it explicitly to the accepted-limitations list (full mitigation isn't really available at this layer).
- **High per security-auditor / Minor per Skeptic — kill-switch (`AE_WORKTREE_READ_GUARD_DISABLE=1`) self-disable scope.** Security-auditor flagged that if a constrained subagent's own tool calls can set this env var in a way that reaches the hook's subprocess environment, it could disable its own containment boundary — exactly the actor the control is meant to restrain. Skeptic assessed this as likely architecturally sound (the hook reads from the long-lived host harness process's environment, not from a subagent's own Bash-call shell) but agreed it's undocumented. **Needs one explicit verification**: confirm empirically whether a subagent's Bash-tool env exports actually propagate to the hook's subprocess env, and document the answer either way in the manifest.
- **High per security-auditor / Minor per Skeptic — trust in the payload's `cwd` field as authoritative for `caller_root`.** Security-auditor questioned whether `cwd` is genuinely unforgeable by the subagent it's meant to constrain, versus being an empirical-only guarantee. Skeptic didn't find a concrete forgery path but agreed the trust assumption is implicit rather than stated. Recommend documenting this as an accepted platform-level trust assumption (the payload is constructed by the Claude Code harness, not the subagent) rather than leaving it implicit.
- **Minor — caller-discrimination brittleness.** The main-session/subagent split is pinned to a small measured payload sample (v2.1.220, 4 records, 2 sessions). If a future harness version changes payload shape, this hook silently and permanently misclassifies affected calls. Worth one sentence in the manifest calling out that the failure mode is total, silent inertness if the sample turns out non-representative.
- **Informational, not a defect** — the module's own docstring already discloses that Bash/Grep/Glob aren't guarded by this hook (only `Read` is), so this shouldn't be represented as closing containment generally — noting so it isn't mistaken for a hidden gap.

Positive controls confirmed by both reviewers: segment-aware exemption matching correctly avoids the `/foo` vs `/foobar` bug (tested), fail-open discipline is comprehensive and consistent, the guarded shell-registration wrapper prevents a missing script from blocking every Read, and test coverage (35 assertions) is genuinely adversarial on most axes (symlink aliasing on multiple operands, segment-boundary exemption negatives, null-byte fail-open).

Recommend resolving the TOCTOU documentation gap and the two reconciled High/Minor items (kill-switch propagation, cwd trust) before treating this as a hard isolation boundary rather than a best-effort guard.

tyson-solara6:
Superseded by #642. `main` advanced by five merges (#632, #633, #634, #636, #638) during review and `gh pr update-branch --rebase` hit conflicts; force-push is denied in this repo, so the work was ported to a fresh branch cut from current `main` rather than rebased in place.

Content is unchanged from the sign-off on this PR (identical 50 files / +1496 / -129). The conflicting adapter artifacts were regenerated on the new base rather than hand-merged, since `.hermes/SKILL.md` embeds every command and a stale-base regeneration can silently revert a sibling PR without `check-adapter-sync` detecting it.

PR #636, opened 2026-08-10
fix(DS-156): bind conductor turn shape to warrant (Unit 1 of 2)

tyson-solara6:
## Summary

DS-156 Unit 1 of 2. Rewrites `content/references/conductor-turn-format.md` to bind conductor turn shape to the turn's warrant, replacing the uniform fixed-slot template that shipped in DS-122 / DS-151 / DS-155.

**Problem.** Those three tickets added a structured block at the TOP of the conductor's turn and left everything after it as ungoverned prose. Structure was added; prose was not removed. Operator's complaint, verbatim: "all it did was add structured outputs to the top and then just still all the normal prose responses is still there."

**Design.**

- **Execution turns** (Answer warrant absent, at least one of Decision/Stoppage/Completion present): the structured block IS the turn. Closed whitelist of permitted line shapes, plus a `STATUS_LINE_MAX_CHARS` (200) length bound on slot lines and the position-1 identity line, so a narrative paragraph cannot hide inside a legitimate `State:` line.
- **Answer turns** (Answer warrant present, wins precedence): prose IS the payload. No status slots, no shaping cap. `## Operator decisions` still goes last if it co-occurs.
- **Relevance discipline** replaces the old line-count framing on Answer turns: six named bans (question restatement, opening preamble, narrating intent, unasked alternatives, closing recap, self-assessment). A long answer to a hard question passes; a three-line answer with a preamble fails.
- **Posture split:** the structural check `_execution_prose_flag` is specified BLOCKING; the phrase-list check `_answer_relevance_flag` stays ADVISORY.

**Unit 2 implements the hook.** Every symbol this PR names (`_execution_prose_flag`, `_answer_relevance_flag`, `ANSWER_BODY_BUDGET`, `STATUS_LINE_MAX_CHARS`) is a contract, not shipped behavior. Public docs carry a uniform greppable `DS-156 CONTRACT, NOT YET SHIPPED` clause across 29 files for removal when Unit 2 lands.

## Review history

Five Skeptic rounds. Two defect classes closed structurally rather than by rewording:

1. **Absolute coverage claims** in the hook-contract section were false on a new axis each round (label, then fence). Fixed by deleting the claims entirely and replacing them with a `Known uncovered shapes` table.
2. **Table rows conflating contract with observed fact.** Fixed with a `Status` column marking each row `observed` (exercisable against live source) or `contract` (Unit 2 behavior).

Final round: 0 Critical, 0 Major, 1 non-blocking Minor (the literal `200` appears alongside the symbol in two table rows; cosmetic drift surface). Sign-off granted.

## North Star alignment

**Pillar 1, guard operator attention** - directly advances it. The founding defect was a change that added capability without reducing what the operator must read, which pillar 1 names as a regression. Execution turns become structurally incapable of carrying a narrative paragraph.

**Pillar 2, verifiable outcomes** - advances it. Advisory enforcement shipped three times on this exact problem and the prose never went away; a hook that logs and exits 0 produces no verifiable outcome. The structural check is specified blocking on that measured 3-for-3 record.

**Pillar 3, low friction** - a bounded cost, explicitly named in the artifact. A false positive on a blocking Stop hook stalls a turn. Mitigated by scoping blocking to the one mechanically unambiguous predicate (is every status-region line a recognized slot line) and leaving the curated phrase-list check advisory until measured.

**Pillar 4, universality** - neutral. No operator identity, tracker, or harness assumption introduced.

## Test plan

- `python3 hooks/tests/test-enforce-turn-shape.py` - 105/105 (hook untouched; confirms the Unit 1/Unit 2 scope boundary)
- `bash scripts/build-all.sh` + adapter-drift diff over the verbatim `adapter-sync.yml` pathspec - clean, `git status --porcelain` empty
- `check-methodology-drift.sh`, `check-resident-budget.sh`, `check-skill-embed-budget.sh`, `check-codex-skill-sync.sh`, `check-symlinks-relative.sh` - all OK
- `npm ci --prefix scripts` + `build-slides.sh` + `git diff --exit-code -- docs/slides` - byte-reproducible
- Merge-result check against current `main` (including #632's completion-warrant widening): auto-merges clean, 125/125 hook tests pass on the merge result, no semantic contradiction

Doc-sync: triggered - updated `docs/index.html`, `docs/slides/autonomy-slides.md` (+ rebuilt `.html`), `README.md`, `docs/components.md`, `docs/configuration-reference.md`, `docs/safe-configuration.md`, `hooks/AGENTS.md`, `.claude/install.sh` (comment), and `content/references/{risk-config-and-tiers,conventions-detail,events-log}.md`, `content/commands/ds-init-project.md` in the same PR.


tyson-solara6:
Automated review pass (self-authored PR, no formal verdict per self-review rules). This is Unit 1 of 2 (spec-only; the enforcing hook is a separate future PR) — confirmed via diff inspection that no `.py` hook logic changed, only the reference doc and its doc-sync sites.

**Findings: Major x1, Minor x2**

- **Major — the execution-turn whitelist doesn't close the exact loophole this PR exists to close.** The spec whitelists any `State:`/`Running:`/`Blocked:` line up to `STATUS_LINE_MAX_CHARS` (200 chars), with no cardinality/dedup check in the blocking `_execution_prose_flag` contract. The "1-3 status lines per turn" cap is explicitly a prose convention, not mechanically enforced — the only count backstop (`_volume_flag`, budgeted at 10 lines) is advisory-only. So a turn built from 8-9 separate `State: <sentence>` lines (each independently valid, each under 200 chars, ~1600 chars of narrative capacity) passes the blocking check on every line and never trips the advisory one — this is exactly the "stuffing prose across many short slot-shaped lines" attack the review brief was built to probe, and it isn't listed in the "Known uncovered shapes" table. Since Unit 2 will implement this contract as written, recommend closing this in Unit 1: either make the per-label line cap part of `_execution_prose_flag`'s blocking domain, or add it honestly to the uncovered-shapes table if the decision is to defer it.
- **Minor** — 4 of the 6 Answer-turn relevance bans are explicitly non-mechanized (semantic judgment only, no hook signal now or after Unit 2). Reasonable given the honesty the spec otherwise aims for, but worth the operator's awareness that "produce verifiable outcomes" is only half-advanced for Answer turns.
- **Minor** — doc-sync (29 files carrying the "CONTRACT, NOT YET SHIPPED" marker) was spot-checked across several sites and found consistent, but not exhaustively diffed against every mirror; low-probability residual drift can't be fully ruled out from this pass.

Recommend closing the Major finding (or explicitly accepting it as a documented gap) before this spec becomes Unit 2's implementation contract.

PR #637, opened 2026-08-10
feat(DS-154): add the standing learnings-capture instruction (Unit B)

tyson-solara6:
Unit B of DS-154. Adds the standing instruction that tells agents to capture learnings in flight, and collapses the `learnings_candidate[]` schema to a single definition.

Nothing is wired yet - Unit C inserts the pointers into the agent files. Unit A (`bin/ds-learning-shard`, #633) shipped the CLI this instruction refers to.

`/ds-wrap` is untouched and is not going away. The goal is that it stop being *necessary*.

## What this adds

**`content/references/learnings-capture-instruction.md`** - the "watch for learnings" instruction, written once. Roughly 20 agent files will point at it rather than each carrying a copy, because copies drift: the Elevated-signal table is duplicated across more files than a grep suggests and took three rounds to reconcile once.

It covers both agent classes, which are genuinely different:
- **Write-capable agents** call `bin/ds-learning-shard append` the moment a learning occurs, in flight.
- **Read-only agents** (declaring `disallowedTools: [Edit, Write, Agent]`) cannot write and cannot spawn, so they populate `learnings_candidate[]` in their return digest instead.

The 7/11 split was verified by frontmatter grep, not assumed.

## The `learnings_candidate[]` extraction found more than expected

The ticket assumed one declaration. There were **five**: `content/agents/engineer.md` twice (YAML and JSON-Schema), `content/references/digest-return-pattern.md`, and inline restatements in `content/agents/investigator.md` and `content/agents/debugger.md`.

Worse, the two agent-file copies restrict `kind` to three values, **omitting `decision`** - a live divergence from the canonical four, not merely duplication. The first three are collapsed to pointers here. The two agent-file copies are Unit C's scope and its brief names them explicitly rather than treating all 19 files as uniform pointer insertions.

The definition lives in the new instruction file rather than `subagent-protocol.md`, which does not mention the field at all today - putting it there would have created a third site instead of collapsing the existing ones.

## Base freshness

Built on `24287263`, the current `main` tip. Two earlier attempts were discarded rather than merged: each regenerated `.hermes/SKILL.md` from a base that went stale mid-build, silently reverting a concurrently-merged PR. The aggregate file embeds every command and reference, so this happens to any `content/`-touching branch that outlives the gap between merges, and `check-adapter-sync` cannot see it - it regenerates and compares, and stale output is self-consistent.

Verified for this branch: `.hermes/SKILL.md` removes 23 lines relative to `main`, and all 23 are the `learnings_candidate` schema block being replaced by a pointer. A control run against the discarded branch returns a higher count, so the comparator is live rather than vacuously empty.

Filed as **DS-155** - this class needs a real detector, not a manual pre-merge diff.

## Verification

- `bin/tests/` **1150 passed**; `test_reference_doc_count_sync.py` and `test_ds_primary_name_sweep.py` green.
- hooks Python 15 files / hooks JS 44 files, 0 failures, including both prose-pinning tests.
- Adapter drift clean using the pathspec copied verbatim from `adapter-sync.yml` (hand-typing it is unsafe - `git diff --exit-code` exits 0 silently on a nonexistent path).
- `check-resident-budget` 391 B / 600 B. `check-skill-embed-budget` 127184 B, floor 100000, ceiling 139160. `check-symlinks-relative` 4/4 relative. `check-methodology-drift`, codex skill sync, slides drift all clean.
- Reference-doc count 36 to 37 across all five sites a test enforces, plus `docs/index.html`, which was independently stale and now reads unambiguously.

## North Star alignment

- **Guard operator attention** - one instruction source instead of twenty copies; agents capture in flight rather than depending on an operator remembering `/ds-wrap` before a session ends.
- **Works for everyone** - the instruction reads a caller-supplied `SESSION_KEY` from the spawn brief and never an environment variable. `CLAUDE_CODE_SESSION_ID` is Claude-only; `AGENTIC_SESSION_ID` and `CLAUDE_SESSION_UUID` are dead, and `bin/ds-migrate` is silently degraded because it reads them.
- No pillar regressed; no new operator ceremony.

Doc-sync: triggered (new `content/references/` file). Updated `docs/index.html`, `CONTRIBUTING.md`, `README.md`, `ADAPTERS.md`, `docs/slides/contributing-slides.md` (rebuilt via `scripts/build-slides.sh`), and `content/SKILL.md`. Decks grepped by hand for `learnings_candidate`, `ds-learning-shard`, and `digest-return`: zero further hits.


tyson-solara6:
Automated review pass (self-authored PR, no formal verdict per self-review rules). Adversarial read via Skeptic, verified against the live repo (frontmatter capability split across all 18 agent files, base-commit contents, exact diff line counts, and the CI test that derives its expected reference-doc count live rather than from a hardcoded literal) rather than trusting the PR body's claims alone.

**Findings: 0 Critical, 0 Major, 1 Minor. Clean bill of health.**

- **Minor (pre-existing, out of scope for this PR)** — `.github/copilot-instructions.md` / `.copilot/build.sh` carry a separate, already-incomplete curated reference-doc list (missing several real docs like `base-branch-sync.md`, `evidence-on-disk.md`, `handoff-evaluation.md` even before this PR) that isn't covered by the reference-doc count-sync test and remains un-updated for the new `learnings-capture-instruction.md` file. This PR doesn't introduce the staleness, just doesn't fix it either. Note: could not auto-file this as a tracked follow-up ticket — the Jira MCP integration isn't available in this environment — flagging here so it isn't lost; worth a manual ticket or a follow-up PR to either bring that list in sync or fold it into the existing count-sync test's coverage.

All four verified claims (four-value `kind` enum consolidation, the write-capable/read-only 7/11 capability split, the `.hermes/SKILL.md` 23-line-only diff, and the reference-doc count bump 36→37 across all five asserted sites) checked out exactly as described. No hidden 6th duplication site found. No code/hook depends on the collapsed inline schema text, so the collapse is behavior-neutral.

tyson-solara6:
Superseded by a rebuild on current `main`. This branch's generated artifacts would revert 107 lines of #636 from `.hermes/SKILL.md` plus one line each from three other files - the DS-155 aggregate-revert class. Hand-authored content is unchanged and carries the same Skeptic sign-off. Continues in the v4 branch.

PR #639, opened 2026-08-10
perf(ds-wrap): defer learnings.md read to the draft-Worker spawn

tyson-solara6:
Defers the `.agentic/learnings.md` full read from Step 0 to immediately before the Step 1 draft-Worker spawn. Routes that never spawn a draft Worker stop paying for content they never use.

Supersedes the first half of #629, which is being closed. The other half of that PR (Part E compression parallelism) is withdrawn - see below.

## What

The live `.agentic/learnings.md` is ~295 KB / ~73K tokens and was read in full at Step 0, *before* Step 0.5 determined the route. Only routes that spawn a draft Worker consume that content - the light path and the ordinary zero-substance path skip Steps 1-3 entirely.

The read now happens immediately before the Step 1 spawn, conditioned on "a draft Worker is being spawned" (which covers both the standard path and the zero-substance staging-drain exception) rather than on a re-derivation of Step 0.5's criteria - `ds-wrap.md` states those criteria are defined in exactly one place, so restating them would be a drift defect even while accurate. The cheap existence probe stays in Step 0's batch.

Six edit sites across five hunks, all located by anchor text: the Step 0-pre "unconditionally reads" sentence, the Step 0-pre short-circuit skip list, the Step 0 read bullet, the new read at the Step 1 spawn, and two stale `(read in Step 0)` cross-references.

**Consumer derivation.** Grepped `learnings` across `content/commands/ds-wrap.md` (18 hits), `content/references/wrap-context-format.md` (0 hits), and `content/references/conductor-operating-rules.md` (24 hits, one of them `.agentic/learnings.md` at `:226`, none a read of its content during `/ds-wrap`). Exactly two sites consume the *content*: the Step 1 prompt embed and Part B step 3's dedup check. Every other hit is existence-only or a verbatim copy (watched-paths mtime signature, legacy migration, the Step 0 probe, Part G's `cp`). Confirmed no route reaches Part B step 3 without first passing a draft-Worker spawn: the light path skips Part B outright, the staging-drain exception spawns the Worker from the same template, and the standard path spawns at Step 1.

## Why this is a fresh branch rather than #629

#629 bundled this with Part E compression parallelism. That second change drew one blocking finding per review round across five rounds - three of them the same defect relocating rather than recurring, then a silent deletion that only a whole-PR pass caught - and grew `ds-wrap.md` by ~13 KB (123,091 -> 136,140 B) to guard a win on a path that fires only when 2+ compression targets are over-gate.

This change alone adds 1,112 B and has drawn zero findings. Part E parallelism is deferred to the per-path locking work, where the concurrency model is designed rather than layered onto the whole-flow lock; the full failure history and carry-forward constraints are recorded there.

## Gates

- `pytest bin/tests/`: 1154 passed
- 38 `bin/tests/test_*.sh`: all pass
- hooks JS suite: 45 files, 0 failures; `test-wrap-md-partA-parity.js` 19/0, `test-wrap-context-format-golden.js` 20/0, `test-wrap-write-discipline.js` 27/0
- `check-command-file-budget.sh`, `check-resident-budget.sh`, `check-skill-embed-budget.sh`: all OK
- `build-all.sh`: 11 adapters; drift pathspec copied verbatim from `adapter-sync.yml`; skill symlinks confirmed relative post-commit

New `bin/tests/test_wrap_learnings_read_deferral_spec.py` pins four invariants: Step 0 no longer reads the file, the short-circuit list no longer names it, the read is gated behind the draft-Worker spawn, and the condition does not restate Step 0.5's criteria. All four verified to fail against `origin/main` by swapping in that content; each pinned string mutated individually, each reddening exactly its own assertion. Auto-discovered by `bin-tests.yml`.

`content/commands/ds-wrap.md`: 123,091 -> 124,203 B.

Doc-sync: predicate not triggered (no public-facing doc asserts learnings.md read timing).

## North Star alignment

**Works for everyone (universality):** a pure reordering of an existing flow. No operator identity, tracker, workspace, or local setup is referenced. A project with a small or absent `learnings.md` behaves exactly as it does today; the saving scales with whatever the project's own file happens to be.

**Enforcement floors preserved:** no gate is removed or weakened. The dedup obligation is unchanged - the same content reaches the same two consumers, just later. The whole-flow lock is untouched, as the two byte-pinning wrap tests confirm. This is a read-timing change, not a review-rigor change.


tyson-solara6:

Adversarial read found this clean. Verified the read is genuinely deferred from Step 0 to immediately before the Step 1 draft-Worker spawn, consistently mirrored across all format adapters (.claude, .codex, .cursor, .gemini, .github, .hermes, .openclaw, .opencode) plus the canonical `content/commands/ds-wrap.md`. No staleness risk - there's no file-mutating activity between the old and new read points, and routes that skip Steps 1-3 never needed the content anyway. Test coverage is good: `test_wrap_learnings_read_deferral_spec.py` uses positional assertions and documents two mutation-tested gaps it closes (read relocated to EOF; read relocated past the field it populates).

**Nit (pre-existing, not introduced here):** Step 0.5's zero-substance staging-drain exception says "still spawn the draft Worker and Skeptic" without an explicit pointer to Step 1's block - relies on an existing doc convention rather than an explicit cross-reference.

No blocking issues.

North Star alignment: ALIGNED - pure reordering, zero behavior change, cuts unnecessary read cost.

tyson-solara6:
Superseded by #647 - same three signed-off commits, rebased onto current `main` (`51fb9d03`).

`origin/main` moved twice while this PR's 24 checks were running, and `gh pr update-branch --rebase` then failed on the generated codex inventory. Force-push is blanket-denied by the harness permission layer, so the resolution is a new branch rather than a rewrite of this one.

The conflict was resolved by **regenerating** `.codex/skill-compatibility.yml` and the four `RESOURCE-MAP.json` files, not by hand-merging or taking a side. A textually-merged generated inventory can end up matching neither branch's real content, and this repo has already shipped a stale-base regeneration that silently reverted 51 files across three merged PRs.

Verified after the rebase: `.hermes/SKILL.md` hunks stay inside the `/ds-wrap` region, with #642's `enforce-worktree-read` and #638's DS-154 reference both intact.

Review history carries over - this content was signed off here (Critical 0, Major 0), plus a follow-up delta closing two Minors that was also signed off. Nothing is lost; this branch retains all three commits.

PR #640, opened 2026-08-10
fix(precommit): point hook symlink at the common repo, not the worktree

tyson-solara6:
`install_precommit_hook` resolved the *hooks directory* correctly for a linked worktree but symlinked `<repo_dir>/hooks/pre-commit` into it. Run from a worktree, that repointed the main checkout's `.git/hooks/pre-commit` at a path about to be deleted - and **git silently does not run a dangling hook**, so commits kept succeeding with no pre-commit checks and no error.

Observed dangling four separate times in one session on a live checkout.

## Fix

Adds `resolve_hook_src <repo_dir>`: compares `git rev-parse --path-format=absolute --git-dir` against `--git-common-dir`.

- **Equal** (ordinary checkout) - target stays `<repo_dir>/hooks/pre-commit`, byte-for-byte as today.
- **Different** (linked worktree) - target becomes `$(dirname <git-common-dir>)/hooks/pre-commit`, the common repo's own hooks file, which outlives any worktree.
- Any resolution failure falls back to today's behavior, preserving the never-abort-the-caller contract.

The hooks-*directory* resolution is untouched - that was DS-58's fix and it is correct.

`uninstall_precommit_hook` calls the same resolver. Its ownership check only removes a symlink pointing at exactly the install-side target, so leaving it on the old path would have silently stopped uninstall from working. The file's manifest header calls out that install and uninstall are symmetric; DS-58 had to fix both sides for the same reason.

## Verification

Pre-fix red against `origin/main`'s `scripts/lib/precommit.sh` (`e23ca35d`), test file held at the fixed version: **20 passed, 5 failed**. The failures are the bug itself - Test 2/3 show the symlink pointing at the ephemeral worktree rather than the main repo, and Test 6 shows it dangling after `git worktree remove`: `test -e` false, `test -x` false, and the hook fails to execute (rc=127). Restoring the fix: **25 passed, 0 failed**.

New coverage in `bin/tests/test_precommit_worktree.sh`:
- **Test 6** - creates a real linked worktree, installs from it, removes the worktree, then asserts the main checkout's hook still resolves **and still executes**. `test -L` alone would pass against the bug, because a dangling symlink is still a symlink.
- **Test 7** - SHA-256 of the real checkout's own hook before and after, proving the test never escapes into the live repo.
- Tests 2/3 corrected to expect the common repo's `hooks/pre-commit` as the target.
- A re-install no-op check.

Real-hook checksum `90924a28…854482` verified identical before and after every run, three times across the session.

## Gates

- `pytest bin/tests/`: 1150 passed
- 37 `bin/tests/test_*.sh`: all pass
- hooks JS: 47/47 including `test-wrap-acquire-lock.js` 80/80 and `test-wrap-release-lock.js` 33/33
- Both shells: `bash` 25/0, `zsh` 25/0 (checked deliberately - `status` and `path` are read-only in zsh and have caused real bugs here; neither appears in the changed code)

`test_precommit_worktree.sh` already existed at a path the `bin-sh-tests` glob discovers, so no workflow edit. `scripts/lib/precommit.sh` is referenced by no adapter `build.sh`, so no adapter rebuild.

Doc-sync: predicate not triggered. The manifest header in `scripts/lib/precommit.sh` documents `resolve_hook_src`, the install/uninstall symmetry, and the failure modes.

## Related

Complementary to `fix/precommit-guard-adoption`, which stops five `bin/tests/*.sh` files from mutating the real hook at all. This PR makes the target durable when the hook is repointed legitimately; that one stops tests from repointing it in the first place. Neither subsumes the other.

## North Star alignment

**Works for everyone (universality):** the ordinary non-worktree install is byte-for-byte unchanged, so no existing setup shifts. The worktree case derives its candidate from git (`--git-dir` vs `--git-common-dir`) rather than from a hardcoded path - but `dirname(--git-common-dir)` **is** still an assumed layout, and it does not hold for bare repos, `--separate-git-dir`, or submodule worktrees. The fix gates that assumption on the candidate file actually existing rather than eliminating it: those three layouts fall back to today's behavior with a warning. That residual is documented in the module's Failure-modes section and asserted by Test 8, not left implicit.

**Enforcement floors preserved:** this restores an enforcement floor rather than relaxing one - a silently-skipped pre-commit hook is a gate that has stopped running without telling anyone. The never-abort-the-caller soft-fail contract is retained, and the new test asserts execution rather than existence so the floor cannot regress into a link that merely looks present.

### DIFF
--- a/bin/tests/test_precommit_worktree.sh
+++ b/bin/tests/test_precommit_worktree.sh
@@ -59,6 +59,24 @@
 #     asserts the main checkout's hook still resolves AND executes -
 #     confirmed to fail against the pre-fix source (see git history/PR for
 #     the baseline failure output this test was written to catch).
+#   - Major 1 (dangling-at-install for bare/separate-gitdir/submodule
+#     worktrees): Test 8 exercises a bare repo + `git worktree add`, where
+#     dirname(--git-common-dir) has no hooks/pre-commit of its own.
+#     resolve_hook_src falls back to the worktree's own hooks/pre-commit
+#     rather than emitting a target dangling from install time. Test 8 ALSO
+#     asserts the KNOWN, undisclosed-no-longer residual: this fallback
+#     target is still inside the worktree, so removing the worktree leaves
+#     the hook dangling anyway (see the library's Failure-modes "KNOWN
+#     RESIDUAL" bullet) - not a regression, not fixed by this PR, disclosed
+#     rather than silently left unverified.
+#   - Major 2 (uninstall orphaning a legacy-targeted hook): Test 9.
+#   - Minor 1 (install_precommit_hook's `[[ ! -f "$hook_src" ]]` guard was
+#     reachable but had zero coverage): Test 10 - an ordinary repo with no
+#     hooks/pre-commit file at all, the only shape that exercises
+#     resolve_hook_src's unchecked fallback branches.
+#   - Minor 3 (legacy_hook_src used the raw, non-canonicalized repo_dir):
+#     Test 11 - install with a trailing-slash spelling, uninstall with the
+#     canonical spelling of the same directory.
 #   This test re-creates worktree fixtures for all of the above to prevent
 #   regression.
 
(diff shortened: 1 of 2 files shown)


tyson-solara6:

This PR is unusually well pre-hardened - 25 tests including nested-worktree, ordinary-checkout regression, uninstall symmetry, and foreign-hook non-clobbering. Two real gaps remain:

**Worth-fixing:** the bare-repo + worktree layout is made non-dangling at install time but not truly fixed. Test 8 shows `resolve_hook_src` falls back to the worktree's own `hooks/pre-commit` when the bare repo's parent has no `hooks/` dir - but that fallback target lives *inside* the worktree, so removing that worktree later reproduces the exact dangling-symlink bug this PR exists to fix, just for a narrower layout. This is likely an acceptable residual limitation (no durable location exists for a bare repo without a working tree), but it's currently undocumented and unverified by a test that actually removes `BARE_WT` afterward.

**Worth-fixing:** `bin/ds-doctor:783` still independently hardcodes `<repo_dir>/hooks/pre-commit` with no shared resolver. The PR's own comment argues this is harmless today only because ds-doctor's separate hardcoded `.git/hooks` path already fails first for worktrees - i.e. this PR's correctness quietly depends on that unrelated bug's failure ordering never changing. Worth a comment at `bin/ds-doctor:783` pointing future readers back here.

**Nit:** the new "source doesn't exist" guard in `install_precommit_hook` writes to stdout while `resolve_hook_src`'s fallback warning goes to stderr - inconsistent, no functional impact.

No migration-breakage concerns - existing stale symlinks self-heal via the pre-existing `_ae_is_ours` repoint logic on next install.

North Star alignment: ALIGNED - fixes a silently-dangling pre-commit hook that let commits succeed with zero checks and no error.

PR #641, opened 2026-08-10
fix(bin/ds-update): pin ds-doctor --fix cwd to the resolved repo_dir

tyson-solara6:
`_run_doctor()` called `subprocess.run(["ds-doctor", "--fix"])` with **no `cwd=`**, so it inherited the invoking shell's working directory rather than the config-resolved `repo_dir` every other operation in the file is scoped to. **Correction (post-review):** the original framing of this PR asserted that `ds-doctor --fix` could repair the wrong checkout from the wrong cwd. That is **false** and has been corrected in the code's docstring. Every `--fix` target is HOME- or config-derived (`~/.claude/settings.json`, `~/.local/bin`, `<repo_dir>/.git/hooks`), with `repo_dir` read from `~/.agentic/agentic-engineering-config.json` - none is cwd-sensitive. The sole cwd-dependent path in `ds-doctor` is a read-only `role-models.yml` probe at `:947`, gated behind `--cross-harness`, which `ds-update` never passes.

So this change has **zero behavioral effect on any current invocation**. It is kept as consistency and defense-in-depth: every other subprocess call in the file passes `cwd=str(repo_dir)` explicitly, and this one is what would scope that `--cross-harness` probe correctly if it is ever wired through.

## Wider than first thought

This was reported as a single dormant call site. It is **three**, all sharing the same omission:

- the already-up-to-date early return (`:584`)
- the no-rebuild-needed early return (`:630`)
- the normal Step 14 path (`:720`)

It stayed inert only because all 16 `invoke_updater` call sites in the existing suite pass either `--no-doctor` or `--check`, and `--check` returns before the doctor branch is reached. Nothing about the code prevented it - the test suite simply never took a path that reaches `_run_doctor`.

## Fix

`repo_dir` is threaded into `_run_doctor(repo_dir)` and all three callers, so there is one resolution point, and the subprocess runs with `cwd=str(repo_dir)`. This matches the file's existing convention - the adapter install loop already does `subprocess.call(cmd, cwd=str(repo_dir))`, and that call was used as the model.

**Sibling audit:** no other `cwd`-less subprocess call exists in the file. The only others are `_git()`, which scopes via `git -C <repo_dir>` (the file's established pattern for git specifically), and the adapter install loop, already correct.

## Verification

New regression test T18 in `bin/tests/test_agentic_update.sh`: stubs `ds-doctor` on PATH to record the cwd it was invoked in, invokes the updater from a *different* cwd **without** `--no-doctor`, and asserts the recorded cwd equals the fixture `repo_dir` rather than the invoking shell's.

Proven red against the unfixed baseline (`origin/main` at `e23ca35d`, reverted via stash with the test held at the fixed version): T18 failed **both** assertions -

```
ds-doctor invoked with cwd '<OTHER_DIR>', expected repo_dir '<FAKE_REPO>'
ds-doctor cwd equals the invoking shell's cwd
```

60 passed, 2 failed. Restoring the fix: 62 passed, 0 failed.

The test also checksums the real `bin/agentic-update` before and after to prove sandbox containment - this repo has a documented history of tests escaping into the live checkout, and a test written to catch a sandbox-escape hazard should not be able to commit one.

## Gates

- `bin/tests/test_agentic_update.sh`: 62/0 under **bash** and 62/0 under **zsh**
- `pytest bin/tests/`: 1150 passed
- hooks JS suite: 45/45 files pass

Two observations reported as seen rather than chased: `test-wrap-daemon.js` hit a tool-call timeout on first run and passed clean standalone (154/154) - it is a slow multi-subprocess test, not a failure. And `test-stdin-guard.js` passed cleanly (78/78) despite a prior report of it failing identically on `main`.

Scope confirmed by `git diff --stat origin/main...HEAD`: two `bin/` files only, no adapter dirs, no `content/`, no docs. `bin/agentic-update` is a symlink to `ds-update`; the real target was edited.

Doc-sync: predicate not triggered. The module manifest header already described cwd-scoped behavior in general terms and needed no correction.

## North Star alignment

**Works for everyone (universality):** the fix resolves `repo_dir` from the same config path every other operation already uses, so behavior is identical regardless of where the operator invokes the updater from, what their checkout is called, or which adapter they run. Nothing operator-specific is introduced.

**Enforcement floors preserved:** no gate is removed or relaxed. This restores an intended scoping guarantee that was silently absent - a repair command escaping its resolved target is the failure this closes, and the new test asserts the scoping rather than the absence of a symptom.

### DIFF
--- a/bin/ds-update
+++ b/bin/ds-update
@@ -581,7 +581,7 @@ def main() -> int:
         _reset_version_check_cache()
         # Skip to doctor
         if not args.no_doctor:
-            return _run_doctor()
+            return _run_doctor(repo_dir)
         return 0
 
     # ------------------------------------------------------------------
@@ -627,7 +627,7 @@ def main() -> int:
         print("No rebuild needed (no adapter source changed).")
         _reset_version_check_cache()
         if not args.no_doctor:
-            doctor_rc = _run_doctor()
+            doctor_rc = _run_doctor(repo_dir)
             if doctor_rc != 0:
                 return doctor_rc
         print("Done.")
@@ -717,7 +717,7 @@ def _run_adapter_installs(
     # Step 14: doctor
     # ------------------------------------------------------------------
     if not args.no_doctor:
-        doctor_rc = _run_doctor()
+        doctor_rc = _run_doctor(repo_dir)
         if doctor_rc != 0:
             return doctor_rc
 
@@ -725,11 +725,18 @@ def _run_adapter_installs(
     return 0
 
 
-def _run_doctor() -> int:
-    """Run ds-doctor --fix. Return 0 on success, 2 on unfixable, 0 on other errors."""
+def _run_doctor(repo_dir: Path) -> int:
+    """Run ds-doctor --fix, scoped to repo_dir. Return 0 on success, 2 on
+    unfixable, 0 on other errors.
+
+    cwd is pinned to the resolved repo_dir (not inherited from the invoking
+    shell) so --fix's filesystem side effects land on the checkout this
+    updater is scoped to, matching every other subprocess call in this file.
+    """
     try:
         result = subprocess.run(
             ["ds-doctor", "--fix"],
+            cwd=str(repo_dir),
             # Inherit stdio so doctor output flows to the terminal
         )
         if result.returncode == 2:
--- a/bin/ds-update
+++ b/bin/ds-update
@@ -61,7 +61,11 @@ Upstream deps: Python 3 stdlib only (argparse, json, os, pathlib, re,
 
 Downstream consumers: end users running from any cwd; CI pipelines; shell
                       aliases; any non-TTY automation that cannot launch the
-                      interactive scripts/update.js TUI.
+                      interactive scripts/update.js TUI. The ds-doctor --fix
+                      subprocess is invoked with cwd pinned to the resolved
+                      repo_dir (see _run_doctor) for consistency with this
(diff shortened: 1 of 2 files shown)


tyson-solara6:

No blocking issues. Verified against the live `_load_repo_dir()` on main: `repo_dir` is already realpath-resolved and symlink-safe before being threaded into `_run_doctor`, so relative-path/symlink invocation isn't a gap here. `subprocess.run([...], cwd=str(repo_dir))` is list-based (no `shell=True`), so no quoting/injection risk. Confirmed the only cwd-sensitive consumer inside `ds-doctor` (the `--cross-harness` role-models probe) is never invoked by `ds-update`, so no other caller depends on the old inherited-cwd behavior.

**Worth-fixing:** the two new exception branches (`FileNotFoundError.filename`-based disambiguation between missing `repo_dir` vs. missing `ds-doctor` binary, and the new `NotADirectoryError` handler) are untested - T18a/b/c only exercise the happy-path cwd-threading, not a deleted or non-directory `repo_dir`. Given this PR's stated bar of proving fixes red-against-baseline, these new branches should get the same treatment.

**Nit:** disambiguating `FileNotFoundError` via `.filename` is a CPython/POSIX implementation detail rather than documented public API - fine for POSIX-only bash tooling, but not portable in principle.

North Star alignment: ALIGNED - closes a latent cwd-dependent failure mode with no new operator-facing surface.

PR #642, opened 2026-08-10
feat(hooks): guard worktree-isolated agent reads into the primary checkout (DS-150)

tyson-solara6:
## Summary

DS-150 Unit C (rebased onto current `main`; supersedes #635): adds `hooks/enforce-worktree-read.py`, a PreToolUse(`Read`) hook that denies a worktree-isolated subagent's `Read` when the target resolves into the PRIMARY checkout instead of the agent's own isolation worktree.

The design is pinned by live payload capture (Claude Code v2.1.220, 4 records across 2 sessions), not inference:

- **Containment predicate** - `caller_root` from the payload's `cwd` field (never `CLAUDE_PROJECT_DIR`, which would compare the primary root against itself; never `os.getcwd()`, since the payload field is authoritative). `primary_root` from `CLAUDE_PROJECT_DIR`. All three operands (target, `caller_root`, `primary_root`) `realpath`-normalized before the test, because isolation worktrees live *inside* the primary root and an unnormalized prefix test is insufficient.
- **Caller discrimination** - main session vs subagent is decided by the **absence** of `agent_id`/`agent_type` (measured: subagent payloads carry 12 top-level keys, main-session 10, the difference being exactly those two). This matches what `enforce-shippable-edit.py` and `enforce-orchestrator-singularity.py` already do, so no new assumption is introduced.
- **Exemptions** - config-driven via `worktree_read_guard_exemptions` in `.agentic/config.json`, shipping empty. Zero built-in entries.
- **Kill-switch** - `AE_WORKTREE_READ_GUARD_DISABLE=1`. Fail-open on every error path.
- **Registered** with the guarded `test -f ... && python3 ... || exit 0` command form, matching the `enforce-turn-shape.py` precedent - an unguarded `python3 <missing>.py` exits 2, and exit 2 on PreToolUse blocks the call, which on a `Read` matcher would deny every read in every session including the conductor's.

Also wired into `bin/ds-doctor`'s `MANAGED_HOOK_BASENAMES` so an orphaned registration is diagnosed rather than certified healthy.

Adding the config key moved two repo-wide counts (behavioral toggles 21 -> 22, `enforce-*.py` hooks 8 -> 9), so this PR also syncs every asserting site and the gates that police them.

## North Star alignment

- **Works for everyone (universality).** The guard resolves everything from the hook payload and `CLAUDE_PROJECT_DIR` at runtime - no operator identity, workspace, tracker, or local path is baked in. The exemption list ships empty and is project-local; an `rclone.conf` exemption inherited from an upstream handoff was deliberately dropped after `grep -rn rclone content/` returned zero here, rather than being carried in as dead config. Fail-open plus a kill-switch means a project or harness where the guard does not apply degrades to prior behavior rather than breaking.
- **Guard operator attention.** The guard converts a silent, late-surfacing failure - an isolated agent reasoning against files outside its own branch, discovered days later as "why did this engineer act on code that isn't in its diff?" - into an immediate, actionable deny at the point of the read.
- **Produce verifiable outcomes autonomously.** The hook gates on payload facts only, never on inferred session capability, per `hooks/AGENTS.md`. Every assertion added here is mutation-verified rather than assumed, and the `_ENFORCER_SUBCOUNT_SITES` sweep now covers `hooks/lib/enforcement_log.py` - closing the specific gap that let a stale subcount survive two review passes.
- **No enforcement floor removed.** This adds a guard; it relaxes nothing. The flat prohibition that no hook may deny conductor `Read`/`Grep`/`Glob` to force delegation is preserved - the guard fires only on a genuinely worktree-isolated subagent, and the main session is exempt by construction.

## Test plan

- [x] `hooks/tests/test-enforce-worktree-read.py` - 35 assertions: main-session allow, non-isolated-subagent allow, own-worktree allow, cross-boundary deny, outside-primary-root allow, symlink normalization on all three operands, kill-switch, full fail-open matrix (absent/malformed `tool_input`, absent `cwd`, absent `CLAUDE_PROJECT_DIR`, unparseable stdin), config-absent/malformed, and configured exemptions including a segment-alignment negative case.
- [x] `bin/tests/test_install_worktree_read_guard.sh` - asserts the generated `Read` matcher carries the `test -f` guard; git-level sandbox so the test cannot repoint the primary checkout's `.git/hooks/pre-commit`.
- [x] `bin/tests/test_agentic_doctor.sh` Test 12 - a missing hook script is reported FAIL.
- [x] `bin/tests/` pytest: 1112 passed. `test_tracker_writeback_ranking_spec.py`: 64 passed, matching `main`.
- [x] `hooks/tests/test-*.py` / `test-*.js` / `test-*.sh` suites clean (`test-stdin-guard.js` fails identically on `main` - missing local `node_modules`, environmental).
- [x] `scripts/build-all.sh` + `git diff --exit-code` over the verbatim `adapter-sync.yml` pathspec: zero drift. `check-methodology-drift.sh`, `check-symlinks-relative.sh`, and the resident / skill-embed / command-file budget gates all OK.

Every assertion added or changed across four review rounds was mutation-verified: broken on purpose, confirmed red, restored, confirmed green.


tyson-solara6:
Automated review pass (self-authored PR, no formal review verdict since GitHub blocks self-review). This PR adds a security-relevant isolation-boundary control, so it got a dedicated security-auditor pass in addition to the general review.

**High - guard scope excludes all non-Read tools (CWE-863, incomplete mediation).** `hooks/enforce-worktree-read.py` and its `.claude/install.sh` registration only match `tool_name == "Read"`. A worktree-isolated agent can trivially read into the primary checkout via `Bash` (`cat`, `python -c open(...)`), `Grep`, or `Glob` - none of which are gated. The PR's own docstring calls this an "accepted limitation," but since Bash is the single most common tool subagents use for exploration, this is the primary bypass of the stated isolation goal and deserves more prominence than a footnote - worth a follow-up to extend coverage (at minimum Bash) or to explicitly reframe this as "advisory-strength, not a hard security boundary" in the docs.

**Medium - case-insensitive filesystem containment bypass (CWE-178).** `_resolve_target`/`_relpath_or_none` do literal string/component comparison after `realpath`, which doesn't normalize casing on case-insensitive-but-case-preserving filesystems (macOS APFS default, NTFS). A `file_path` differing only in case from a `CLAUDE_PROJECT_DIR` segment can resolve to `rel_target=None` (looks "outside primary_root") while the OS opens the same inode inside it - a fail-open ALLOW on a read that's actually in-bounds. Suggest `os.path.samefile`/inode-based containment instead of pure string comparison, plus a test case.

**Medium - TOCTOU via symlink swap (CWE-367).** Containment is checked once via `realpath` at decision time; the actual open happens later in the harness's Read implementation. A symlink swapped between check and open could dodge either a deny or an allow. Low practical severity given per-turn serialization, but worth noting since the design rests on realpath-at-a-point-in-time.

**Good, confirmed:** kill-switch and missing-hook-file paths correctly fail open (avoiding the "missing script exits 2 -> blocks everything" trap); `caller_root` is sourced only from the payload's `cwd`, never `os.getcwd()`/`CLAUDE_PROJECT_DIR`, avoiding a self-comparison bug; telemetry is isolated in its own try/except so a broken logging path can't suppress a deny; 35 unit tests cover symlink normalization and a fail-open matrix for malformed payloads.

North Star alignment: ALIGNED - converts a silent isolation violation into an immediate deny without adding routine operator babysitting, though see the High finding above on how much protection it actually provides today.

PR #648, opened 2026-08-10
fix(DS-156): implement blocking execution-turn shape check (Unit 2 of 2)

tyson-solara6:
## Summary

DS-156 **Unit 2 of 2**. Implements the `content/references/conductor-turn-format.md` §9 hook contract that Unit 1 (#636) specified, and removes all 32 `DS-156 CONTRACT, NOT YET SHIPPED` markers now that the behavior exists.

**Why this ticket exists.** DS-122, DS-151 and DS-155 each shipped green and none fixed the operator's complaint: structured output was added at the top of the conductor's turn while all the prose stayed. Unit 1 rewrote the spec; this unit makes it real.

## What ships

- **`_execution_prose_flag` - BLOCKING.** The first blocking path in this hook. Two branches: the sole-stoppage branch is predicate-identical to the `_forced_yield_flag` it replaces (raw-line domain, `Waiting:` lines only, no length test); the general branch permits only `State:`/`Running:`/`Blocked:`/`Waiting:` lines in the status region, with slot lines bounded by `STATUS_LINE_MAX_CHARS = 200`. The same bound applies to the position-1 identity line on both branches - a LENGTH test only, never shape validation (DS-155 rejected two identity-line shape predicates and that stays out of scope).
- **`_answer_relevance_flag` - ADVISORY.** Curated phrase-list detectors for ban 2 (opening filler) and ban 5 (closing recap) only. Bans 1/3/4/6 stay prose rules; a real answer necessarily reuses nouns from the question it answers, so no reliable mechanical heuristic separates that from restatement.
- **`_volume_flag`** re-budgets to `ANSWER_BODY_BUDGET = 50` on Answer turns as a runaway-generation backstop, not a shaping constraint. `BASE_BODY_BUDGET = 10` and its free pools are byte-identical for execution turns.
- **Three-way partition** (Answer / execution / zero-warrant), exhaustive and non-overlapping. Zero-warrant stays with `_status_only_flag` at its existing advisory posture; no zero-warrant turn can reach the blocking path.
- **Blocking mechanism:** the process always exits 0 and blocks via a `{"decision":"block"}` stdout payload. A non-zero exit would be silently swallowed by the registration wrapper's `|| exit 0` in `.claude/install.sh` - the spec now says so explicitly, because the previous wording asserted the opposite and would have produced a hook that ships green and never fires.

## Verification

The Skeptic ran the founding-complaint shape through the live hook rather than citing a test name:

| input | verdict |
|---|---|
| preamble at position 1 + 90-word `State:` line | BLOCK |
| Decision-warrant turn + one narrative sentence | BLOCK |
| sole-stoppage + fenced narrative aside | BLOCK |
| identity line 199 / 200 / 201 chars | QUIET / QUIET / BLOCK |
| slot line 200 / 201 chars | QUIET / BLOCK |
| 140-char `Waiting:` line, both branches | QUIET (no length bound imported) |
| zero-warrant, 4 prose lines | ADVISORY only |

Nine false-positive probes were all quiet: multi-ticket `State:` line, long branch name in the identity line, a 7-`Waiting:` forced yield, `Waiting:` on the general branch, a fenced block in the status region, multi-line `## Operator decisions` items, an Answer+Decision combo, a realistic `[phase: complete]` turn, and a `Status: DONE` relay.

Two residuals are disclosed as `observed` rows in the Known uncovered shapes table rather than silently fixed: one line of arbitrary prose at position 1 under 200 chars (closing it needs an identity-line shape predicate, out of scope), and an incidental quoted fragment granting the best-effort Answer warrant, which downgrades an otherwise-blocking execution turn to advisory.

## North Star alignment

**Pillar 1, guard operator attention** - this is the unit that delivers it. An execution turn is now structurally incapable of carrying a narrative paragraph.

**Pillar 2, verifiable outcomes** - advisory enforcement shipped three times on this problem and the prose never went away. The structural check now blocks on that measured record; the phrase-list check stays advisory until its fire rate is measured.

**Pillar 3, low friction** - a false positive on a blocking Stop hook stalls a turn, so blocking is scoped to the one mechanically unambiguous predicate, with the loop guard capping consecutive blocks at 2. Thirteen probe shapes were run to confirm legitimate turns pass.

## Test plan

- `hooks/tests/test-enforce-turn-shape.py` - 125 → 143. Existing scenarios were re-derived to the new expectations, not deleted; two additions were mutation-verified non-vacuous (`ds156-i2` and the rewritten `s8b`).
- `pytest bin/tests/ -q --timeout=60 --timeout-method=thread` - 1150/1150
- Hooks JS suite, verbatim CI loop - 45/45
- `build-all.sh` + adapter-drift over the verbatim `adapter-sync.yml` pathspec - clean, `git status --porcelain` empty

`bin/tests/test_tracker_writeback_ranking_spec.py` is touched deliberately: this change moves `enforce-turn-shape.py` into the `"deny"` group as well as `"allow_advisory"`, so the pinned enforcer enumeration and its stale-count sweep both needed updating. Three mutations confirmed the sweep still goes red on a genuinely stale count.

Doc-sync: triggered - all 32 not-yet-shipped markers removed and the surrounding sentences rewritten to present tense across `README.md`, `docs/{index.html,components.md,configuration-reference.md,safe-configuration.md}`, `docs/slides/autonomy-slides.md` (+ rebuilt `.html`), `hooks/AGENTS.md`, `.claude/install.sh` (comment only), and `content/references/{risk-config-and-tiers,conventions-detail,events-log}.md`, `content/commands/ds-init-project.md`.


tyson-solara6:
Superseded: rebased again onto current main (#644 landed mid-flight). Force-push is denied in this repo, so the rebased work ships on a new branch.

PR #650, opened 2026-08-10
fix(hooks): detect a stale hooks snapshot so merged hook fixes stop shipping dormant

tyson-solara6:
## Summary

Hooks run from a per-checkout snapshot directory, not from the git checkout. That is deliberate (DS-54): a bare `git pull` must never silently rewire a hook command a live session already resolved, after an incident where a hooks change bricked live delegation the instant it landed.

The consequence was undocumented and unhandled: **a merged hook fix stays dormant until an adapter install runs.** Two hook fixes merged earlier today and neither took effect until a manual reinstall.

A detector already existed (`hooks/lib/hooks-staleness-core.sh`, wired into `session-start-wrap.sh`) but fires only at a fresh session's first tool call. This adds the missing trigger points.

## What changed

- **`ds-doctor`** gains a hooks-snapshot staleness check, reusing the existing classifier rather than reimplementing it. `--fix` calls the same `sync_hooks_snapshot` that `install.sh` already runs unconditionally.
- **`ds-base-sync`** prints an advisory line after every base-branch sync. Per `AGENTS.md` it runs unconditionally at the `/ds-implement-ticket` Phase 12 tail, making it the one guaranteed post-merge trigger. Read-only; exit code captured and restored so it can never change the result.
- **`ds-update`** no longer skips silently when `old_head == new_head`. That path previously returned early past both the adapter loop *and* its own hooks-changed note, so a user who pulled manually first got a silent no-op - the exact scenario that produced this PR.
- **`hooks_source_paths()`** extracted into `scripts/lib/hooks-snapshot.sh` as the single source of the hook-source path list, now driving both the hash and the copy in `sync_hooks_snapshot`, plus the staleness classifier and `ds-update`. It was previously hand-copied in three places.

**The DS-54 invariant is preserved and was verified**: every path that can mutate the snapshot requires an explicit human command. A sweep across `hooks/`, `scripts/`, `content/` and every adapter directory confirms no hook, cron, session-start, or wrap flow invokes `ds-doctor --fix`. The `/ds-update` command's `--no-doctor` contract is respected, not subverted.

## Adapter asymmetry, now documented

`.cursor` and `.opencode` symlink hook config directly into the checkout, so they get hook fixes live on `git pull` and have no dormancy gap. Claude Code, Codex, Gemini and Kimi are snapshotted and can run stale. This is counterintuitive - the primary harness is the stale one - and was previously written down nowhere.

## Review

Five Skeptic rounds; sign-off granted at round 5. Findings resolved along the way, each verified by execution rather than by reading:

- The snapshot classifier reported `OK ... hooks snapshot is current` when it **crashed**, because the exit code was never inspected. A detector whose failure mode is claiming everything is fine.
- `--fix` reported `half_applied` as resolved and exited 0 when `sync_hooks_snapshot` cannot fix that state, leaving an adapter wired to the live checkout - the DS-54 hazard state itself.
- The first fixture could not detect list drift at all: it contained none of the hashed paths, so any list produced the same empty-input digest. Proven by corrupting the list and watching 64/64 still pass.
- `content/references/base-branch-sync.md`'s CLI contract had gone stale against the shipped stdout.

Verified by the reviewer's own execution: snapshots byte-identical before and after the extraction, across hostile `repo_dir` forms (spaces, trailing slash, symlinked component, relative); `chmod 700` on `ds-identity` still fires; the bash 3.2 empty-array guard correct at all four sites against real `/bin/bash` 3.2.57; a six-way drop-one mutation table pinning each source path individually, non-vacuous on destination as well as presence.

- `bin/tests` bash 807/0; `test_hooks_snapshot.sh` 44/44; pytest 1149; JS 45/0; all three budget gates OK; adapter drift clean
- The four zsh failures are pre-existing and were confirmed identical at `origin/main`; none touches a file in this diff

## North Star alignment

**Produce verifiable outcomes.** A merged hook fix silently doing nothing is the opposite of a verifiable outcome. This closes the gap at three trigger points and makes the state inspectable on demand.

**Works for everyone.** Resolution is per-operator throughout - no baked-in identity, workspace, or path. The consumer list in the module manifest is now derived by a sweep command embedded in the file itself, so the next reviewer runs it rather than hand-listing; hand-listing is what left `.codex/uninstall.sh` missing twice.

**Enforcement floors are not removed by convenience.** DS-54's invariant was treated as load-bearing throughout: no addition here lets a passive action rewire hooks under a running session. The one friction cost - a developer with uncommitted `hooks/` edits now gets the adapter-install loop on every "already up to date" `ds-update` - is named in the code and both AGENTS files rather than left to be discovered.

Doc-sync: predicate triggered - `content/references/base-branch-sync.md` updated for the advisory line and the capture/restore exit-code contract, 4 regenerated adapter copies committed, `.codex` inherits via relative symlink; `hooks/AGENTS.md` gained a "Merged is not live" section; `AGENTS.md` and `bin/AGENTS.md` updated.


tyson-solara6:
Superseded: `main` advanced with DS-150 (#642, #644), which touches `bin/ds-doctor`, its test suite, and the generated `.hermes/SKILL.md`, so this branch now conflicts. Force-push is denied in this repo, so the work is being replayed onto current `main` as `fix/hooks-snapshot-staleness-detection-v2` with the conflicts resolved (generated files regenerated rather than hand-merged). Closing rather than bundling, per AGENTS.md. Skeptic sign-off carries over: 0 Critical, 0 Major across five rounds at `a53e8eea`.

PR #652, opened 2026-08-10
fix(DS-152): stop a scratch worktree from dangling the repo's shared pre-commit hook

tyson-solara6:
## Summary

`ds-doctor` found this repo's `.git/hooks/pre-commit` pointing at `/tmp/skeptic-621-SB4Cfv/hooks/pre-commit` - a scratch worktree deleted weeks earlier. The hook had been silently dead ever since.

Root cause: `install_precommit_hook <repo_dir>` resolved the hook **destination** to the shared common `.git/hooks` (correct - hooks are shared across all worktrees) but sourced the hook **body** from `<repo_dir>/hooks/pre-commit`, i.e. whatever directory it was handed. Running any adapter's `install.sh` from an ephemeral scratch worktree therefore repointed the entire repo's shared hook at a directory about to be deleted, and `_ae_is_ours()` had no durable-vs-ephemeral distinction so it did this without warning.

Scratch worktrees of exactly that shape are routine under `content/references/qa-regression-obligation.md`. The bug reproduced itself twice more during the work on it.

## What changed

- **`resolve_primary_checkout`** - the hook body is now always sourced from the primary checkout, never the invoking `repo_dir`. Sourcing from a worktree only made sense if the design provided per-worktree hook *content*, and it does not: `hooks/pre-commit` early-exits its whole adapter-build section inside any worktree.
- **Existence check before symlinking** - install refuses, loudly and non-fatally, rather than writing a link to a nonexistent target.
- **Canonicalization** - `git rev-parse --git-path hooks` returns a *relative* `.git` from a primary checkout but a *canonical absolute* path from a linked worktree, so raw string comparison silently failed on any path with a symlinked component.
- **Legacy residue cleanup** - a hook installed by the old code, including one whose target worktree is already deleted, is now recognized and removed when it lies under the repo's own worktree directories.
- **Relative link targets are anchored against the hooks directory**, per POSIX, instead of the process CWD.

## Nothing shipped broken

The dead hook caused no bad merge. `hooks/pre-commit` skips its adapter-build and staging section inside any worktree, so it was never load-bearing for the worktree-isolated PR workflow - and `check-adapter-sync` rebuilds all 11 adapters server-side on every PR regardless of local hook state. A full `build-all.sh` against `main` produces zero diff.

## Review

Six Skeptic rounds. Findings resolved, each reproduced by execution rather than argued:

- **The test suite could dangle the live repo's shared hook.** With `GIT_DIR` set in the environment, git ignores `-C <dir>` entirely, so the fixture harness silently operated on the ambient repo and wrote a symlink into it pointing at a deleted fixture. Reproduced, then closed by unsetting the git env vars and adopting the existing hook guard.
- **A Critical infinite loop.** The walk-up over a dangling path never terminated on a slash-free component, hanging `uninstall_precommit_hook` and all three adapter uninstall scripts with unbounded memory growth. Reproduced at rc=124 under both shells.
- **A regression test that tested nothing.** The hang test wrapped its call in `timeout` and asserted `rc != 124` - but `timeout` is not a stock macOS binary, so without it the call returns 127 and the assertion passes while verifying nothing. Now resolves `timeout`/`gtimeout` explicitly and hard-fails under `CI`.
- **The uninstall path was hardened to fail closed**, since its failure direction is deletion.

Two findings were the same defect relocated rather than removed, and three fixtures were built to match the fix instead of the defect - each caught only because the reviewer executed the repro rather than reading the diff.

- 54/54 in all four {bash, zsh} x {symlinked, non-symlinked} combinations, at equal assertion counts
- `bin/tests` 38/38 suites; pytest 1149; JS 44/44; all three budget gates OK
- Every new assertion mutation-tested; the anchor and the termination guard proven independently load-bearing

**CI cannot catch this class.** macOS `TMPDIR` is symlinked and ubuntu's `/tmp` is not, so a symlink-sensitive assertion only ever fails on a developer's machine. The suite is therefore run from both a symlinked and a non-symlinked root.

## North Star alignment

**Works for everyone.** The bug was pure path-spelling portability: it fired for any checkout reached through a symlinked home, dev directory, or macOS `TMPDIR`, and was invisible to CI and to anyone on a non-symlinked path.

**Produce verifiable outcomes.** A silently dead pre-commit hook is the worst kind of green - present, executable-looking, and doing nothing. This makes every failure path loud and non-fatal, and closes a test-suite escape that could corrupt the operator's real repository.

**Enforcement floors are not weakened.** The uninstall ownership predicate was widened only as far as measurement justified, and was verified against six adversarial targets (unrelated repo, user script, husky-style path, non-repo path, foreign symlink, target destruction) that must all be preserved. The deletion blast radius is bounded to the hook symlink itself, never its target.

Doc-sync: predicate not triggered - internal shell library plus its test suite; no public count, list, or path assertion in `README.md`, `CONTRIBUTING.md`, `content/SKILL.md`, or `docs/` is affected.

### DIFF
--- a/bin/tests/test_precommit_worktree.sh
+++ b/bin/tests/test_precommit_worktree.sh
@@ -865,19 +865,18 @@ case "$DANGLE_LEGACY_HOOKS_DIR" in
 esac
 DANGLE_LEGACY_HOOK_DST="$DANGLE_LEGACY_HOOKS_DIR/pre-commit"
 
-# resolve_primary_checkout canonicalises repo_dir (via _pc_canonicalize_dir)
-# before uninstall_precommit_hook passes it to _pc_is_legacy_sibling_hook's
-# path-prefix check, so the fixture's target path must be built from the
-# SAME canonical form - not the raw $DANGLE_LEGACY_MAIN string, which can
-# differ on a symlinked TMPDIR/HOME (e.g. macOS's
-# /var/folders -> /private/var/folders).
-DANGLE_LEGACY_MAIN_CANON="$(cd "$DANGLE_LEGACY_MAIN" && pwd -P)"
-
-# The dangling target's path lies under the repo's own known worktree root
-# (.claude/worktrees/) even though that directory no longer exists on disk -
-# simulating a Claude-harness worktree that was already deleted without its
-# legacy hook being uninstalled first.
-DANGLE_LEGACY_TARGET_DIR="$DANGLE_LEGACY_MAIN_CANON/.claude/worktrees/agent-deleted-example"
+# DS-152 round 4 (Major, was relocated here in round 3): the dangling
+# target's path is built from the RAW $DANGLE_LEGACY_MAIN string (as
+# `.claude/install.sh`'s pre-fix `REPO_DIR="$(cd ... && pwd)"` - logical
+# pwd, NOT `pwd -P` - would actually have spelled it for any checkout
+# reached through a symlinked component), NOT a pre-canonicalised form. This
+# is the shape the defect lives in: on a symlinked TMPDIR/HOME (e.g. macOS's
+# /var/folders -> /private/var/folders) this raw spelling differs, on disk,
+# from `primary_checkout`'s canonical form even though both name the exact
+# same real directory. A fixture built from the canonical form (round 3's
+# shape) cannot exercise this - see _pc_canonicalize_missing_dir below for
+# the fix that makes comparing these two spellings actually work.
+DANGLE_LEGACY_TARGET_DIR="$DANGLE_LEGACY_MAIN/.claude/worktrees/agent-deleted-example"
 mkdir -p "$DANGLE_LEGACY_HOOKS_DIR"
 ln -s "$DANGLE_LEGACY_TARGET_DIR/hooks/pre-commit" "$DANGLE_LEGACY_HOOK_DST"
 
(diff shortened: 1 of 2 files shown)


tyson-solara6:
Superseded by #640, which landed the same root-cause fix from a concurrent session (independent investigation of the same `ds-doctor` symptom; no DS-152 cross-reference). Closing rather than rebasing: the two implementations use different mechanisms (`-ef` inode identity vs string equality, `resolve_primary_checkout` vs `resolve_hook_src`), so a textual merge would produce conflicting parallel helpers rather than a coherent design.

Verified against current `origin/main`, with executed repros:
- **Covered by #640**: sourcing the hook body from the common repo, canonicalization of existing paths, and the "never write a NEW dangling link" guard.
- **Not covered, live on main**: orphan cleanup for a hook naming an already-deleted worktree (reproduced - it survives uninstall permanently, since both of #640's legacy candidates derive from the invoking `repo_dir`); the test suite's sandbox escape under an ambient `GIT_DIR` (reproduced - `git -C` is ignored and the install writes into the ambient repo's hooks dir); and the warning for an already-dangling hook on every run.
- **Not applicable**: the Critical infinite loop this branch fixed was self-inflicted by its own walk-up logic, which `main` never performs. Confirmed rc=0, no hang, against `origin/main`.

The uncovered items will land as a targeted delta on top of #640 rather than as this branch.

PR #654, opened 2026-08-10
fix(DS-157): suppress status-only nag on no-identity-line completion reports

tyson-solara6:
## Summary

DS-157. `hooks/enforce-turn-shape.py` fired its advisory status-only finding on a genuine completion report - observed live, on a turn reporting DS-156's own merge. This suppresses that false nag without granting the Completion warrant.

**Mechanism.** `_LEADING_COMPLETION_RE` is anchored to string-start and only ever inspects position 1. In the observed turn the identity line was `Both units shipped.` - `shipped` is a recognized past participle, but the pattern also requires a completion-adjacent word within 40 characters and none follows before the period. The actual declaration, `**DS-156 is done.**`, sat in the first body paragraph, which that anchor deliberately never scans (DS-156 round 1 established that scanning the body reopens a false-positive class: completed sub-items inside still-in-progress turns).

## Why this is not a warrant widening

Completion is an **execution** warrant. Granting it routes a turn onto `_execution_prose_flag`, which is BLOCKING and permits only `State:`/`Running:`/`Blocked:`/`Waiting:` slot lines.

All six true positives found in measurement are multi-paragraph prose reports. Verified by execution: simulating the warrant-granted state and calling `_execution_prose_flag` directly returns a blocking finding. **Widening the warrant would have converted today's advisory nag into a hard block on exactly the turns it was meant to help** - strictly worse than the status quo.

So the fix recognizes the shape one level down, inside `_status_only_flag` only (`_has_body_completion_declaration`), suppressing the advisory while never setting `warrants["completion"]`. Test `y6c` pins that distinction directly: it simulates the warrant-granted state and asserts `_execution_prose_flag` still blocks, so a future refactor cannot quietly fold this signal into the warrant dict.

## Measurement

Same corpus and method as #632, which set the standard for this class of change. `~/.claude/projects`: 3,564 files, 3,901 main-agent candidate final turns, 970 currently flagged status-only.

| round | change | newly captured | false positives |
|---|---|---|---|
| 1 | scan any unfenced body paragraph | 15 | **4 (26.7%)** - a tally header and a sub-item trap, reproducing #632 round 1's defect class one position over |
| 2 | restrict to the first unfenced body paragraph + tally-header exclusion | - | 2 |
| 3 | extend the continuing-work veto with "in progress", "review(s) running", "running on", "still open" | 6 | **0** |

Final: 6 of 970 (0.6%) newly suppressed, 0 false positives on independent re-labelling of all six. Each added veto phrase traces to a real false positive found in measurement, not to speculation.

## North Star alignment

**Pillar 1, guard operator attention** - a guard that fires on correct turns trains the conductor to ignore its own feedback channel, which the module's docstring names as worse than no hook at all. This removes a false nag on 0.6% of status-only fires.

**Pillar 2, verifiable outcomes** - the change is measured against a 3,901-turn corpus with false positives counted per round, not asserted.

**Pillar 3, low friction** - materially improved. The alternative (widening the warrant) would have hard-blocked prose completion reports; this suppresses an advisory and touches no blocking path.

## Test plan

- `hooks/tests/test-enforce-turn-shape.py` - 143 → 151. The eight additions cover both the newly-recognized shapes and the false-positive shapes that were rejected; the rejected ones are what stop a future widening from reopening this.
- `pytest bin/tests/ -q --timeout=60 --timeout-method=thread` - 1155 passed
- Hooks JS suite, verbatim CI loop - 45 files, 0 failed
- `build-all.sh` + adapter-drift over the verbatim `adapter-sync.yml` pathspec - clean

Doc-sync: triggered - updated `content/references/conductor-turn-format.md` (§2 Completion-warrant bullet, the `_status_only_flag` hook-contract bullet, and the Known-uncovered-shapes history) and rebuilt all 11 adapters. `docs/index.html`, `docs/slides/*.md`, `README.md`, `docs/configuration-reference.md` and `docs/safe-configuration.md` describe only the general blocking/advisory split, which this change does not alter.


tyson-solara6:
Superseded by round-2 findings: the round-1 commit widened a regex shared with the warrant path, costing 18 completion-warrant false-positive losses that CI could not see. Force-push is denied here, so the corrected work ships on a new branch.

PR #655, opened 2026-08-10
fix(DS-157): suppress status-only nag on no-identity-line completion reports

tyson-solara6:
## Summary

DS-157. `hooks/enforce-turn-shape.py` fired its advisory status-only finding on a genuine completion report - observed live, on a turn reporting DS-156's own merge. This suppresses that false nag without granting the Completion warrant.

**Mechanism.** `_LEADING_COMPLETION_RE` is anchored to string-start and only ever inspects position 1. In the observed turn the identity line was `Both units shipped.` - `shipped` is a recognized past participle, but the pattern also requires a completion-adjacent word within 40 characters and none follows before the period. The actual declaration, `**DS-156 is done.**`, sat in the first body paragraph, which that anchor deliberately never scans (DS-156 round 1 established that scanning the body reopens a false-positive class: completed sub-items inside still-in-progress turns).

## Why this is not a warrant widening

Completion is an **execution** warrant. Granting it routes a turn onto `_execution_prose_flag`, which is BLOCKING and permits only `State:`/`Running:`/`Blocked:`/`Waiting:` slot lines.

All six true positives found in measurement are multi-paragraph prose reports. Verified by execution: simulating the warrant-granted state and calling `_execution_prose_flag` directly returns a blocking finding. **Widening the warrant would have converted today's advisory nag into a hard block on exactly the turns it was meant to help** - strictly worse than the status quo.

So the fix recognizes the shape one level down, inside `_status_only_flag` only (`_has_body_completion_declaration`), suppressing the advisory while never setting `warrants["completion"]`. Test `y6c` pins that distinction directly: it simulates the warrant-granted state and asserts `_execution_prose_flag` still blocks, so a future refactor cannot quietly fold this signal into the warrant dict.

## Measurement

Same corpus and method as #632, which set the standard for this class of change. `~/.claude/projects`: 3,564 files, 3,901 main-agent candidate final turns, 970 currently flagged status-only.

| round | change | newly captured | false positives |
|---|---|---|---|
| 1 | scan any unfenced body paragraph | 15 | **4 (26.7%)** - a tally header and a sub-item trap, reproducing #632 round 1's defect class one position over |
| 2 | restrict to the first unfenced body paragraph + tally-header exclusion | - | 2 |
| 3 | extend the continuing-work veto with "in progress", "review(s) running", "running on", "still open" | 6 | **0** |

Final: 6 of 970 (0.6%) newly suppressed, 0 false positives on independent re-labelling of all six. Each added veto phrase traces to a real false positive found in measurement, not to speculation.

## North Star alignment

**Pillar 1, guard operator attention** - a guard that fires on correct turns trains the conductor to ignore its own feedback channel, which the module's docstring names as worse than no hook at all. This removes a false nag on 0.6% of status-only fires.

**Pillar 2, verifiable outcomes** - the change is measured against a 3,901-turn corpus with false positives counted per round, not asserted.

**Pillar 3, low friction** - materially improved. The alternative (widening the warrant) would have hard-blocked prose completion reports; this suppresses an advisory and touches no blocking path.

## Test plan

- `hooks/tests/test-enforce-turn-shape.py` - 143 → 151. The eight additions cover both the newly-recognized shapes and the false-positive shapes that were rejected; the rejected ones are what stop a future widening from reopening this.
- `pytest bin/tests/ -q --timeout=60 --timeout-method=thread` - 1155 passed
- Hooks JS suite, verbatim CI loop - 45 files, 0 failed
- `build-all.sh` + adapter-drift over the verbatim `adapter-sync.yml` pathspec - clean

Doc-sync: triggered - updated `content/references/conductor-turn-format.md` (§2 Completion-warrant bullet, the `_status_only_flag` hook-contract bullet, and the Known-uncovered-shapes history) and rebuilt all 11 adapters. `docs/index.html`, `docs/slides/*.md`, `README.md`, `docs/configuration-reference.md` and `docs/safe-configuration.md` describe only the general blocking/advisory split, which this change does not alter.

---

## Round 2 (Skeptic BLOCK: 2 Major, 2 Minor)

Round 1's safety analysis covered `_has_body_completion_declaration` but missed that the same commit widened `_CONTINUING_WORK_PHRASE_RE` - a regex **shared** with `_classify_warrants`, which grants the blocking-adjacent Completion warrant. CI could not see this; only corpus measurement could.

Full-corpus both-directions measurement, each phrase in isolation (~3,937 main-agent final turns, 324 matching any round-1 phrase):

| phrase | warrant losses | verdict |
|---|---|---|
| `still open` | 10/10 false positives - backlog and PR-list vocabulary describing *other* work | dropped |
| `in progress` | 7/7 false positives - a tracker status value for a *different* ticket | narrowed to the bold sub-heading shape `**In progress:**`, its one real motivating case's actual form; re-measured 0/7 |
| `review(s) running` | 0 | kept |
| `running on` | 0 | kept |

Net: 17 of 18 measured completion-warrant false-positive losses eliminated, 1 genuine loss correctly retained, 0 status-only regression. Three-direction delta against the pre-DS-157 baseline: status-only newly-quiet 0 / newly-firing ~2; warrant gained 0 / lost 1 (down from 18 under the round-1 set); `_execution_prose_flag` block/pass delta 0/0.

`in progress` being *worse* than the phrase the review flagged is why the round-1 measurement's one-directional method was the real defect, not any single phrase.

Also fixed: the stale module manifest that still claimed three veto phrases after four were added - the omission that concealed the shared-regex effect from a reader of the header; the one-directional measurement docstring, rewritten to report all three directions plus the exec-flag delta; `y6c` retitled to describe what it actually pins.

Tests 151 → 155. `y7` pins that a genuine completion turn mentioning unrelated open work keeps its warrant. `y8a/b/c` pin the latent sole-stoppage routing flip demonstrated in review - characterization tests, not a fix; that residual is disclosed, not closed.


tyson-solara6:
Superseded by round-3 fixes (regression test pinning the in-progress narrowing, wording correction, and a Known-uncovered-shapes row). Force-push is denied here, so the work ships on a new branch.

PR #659, opened 2026-08-10
fix(DS-158): retry session-log lock timeout instead of silently dropping the record

tyson-solara6:
## Summary

DS-158. A `hooks-js-tests` failure on an unrelated documentation-only PR (#657) turned out to expose a real production defect in telemetry writes, not just a flaky test.

**The defect.** `bin/ds-identity`'s session-log append acquires an exclusive `flock` with a bounded timeout. That call sits inside an 8-attempt retry loop whose `except` clause caught only `FileNotFoundError` (for log rotation), so a lock-timeout `RuntimeError` **escaped the loop entirely** and landed in the outer `except (OSError, RuntimeError): return False, False`. A timeout therefore got exactly one attempt, no retry, and was swallowed. `hooks/stop-context.js:writeTelemetrySafely` calls `write-hook` once and does not retry either.

Net effect: two genuine concurrent conductor sessions on the same project could permanently drop one session's telemetry record. Rare, silent, and exactly the soft-failing-telemetry pattern that has hidden real breakage in this repo before.

## The fix

- **The timeout is now retried, not dropped.** `_lock_fd_exclusive` raises a dedicated `_LockTimeoutError(RuntimeError)`, and `_write_jsonl_safely`'s attempt loop catches it alongside `FileNotFoundError` - caught by the base class rather than the subclass, so it stays correct against any RuntimeError-based lock helper including a test double.
- **Budget is bounded on both axes.** Attempts share the existing `max_attempts = 8`; wall-clock is bounded by `SESSION_LOG_LOCK_BUDGET_SECONDS` computed once at entry. Attempts alone cannot bound worst-case latency (which the JS subprocess ceiling depends on); wall-clock alone cannot guarantee a minimum number of tries. Backoff is linear-capped and itself bounded by the remaining budget, so it is spent *inside* the ceiling rather than added on top.
- **Total ceiling held at parity.** The per-attempt timeout was replaced by a total budget plus a per-attempt cap, deliberately keeping the overall ceiling where round 1 left it. The fix is that a timeout is now retried within the budget, not that the budget grew.

**Surfacing was already correct** - verified rather than assumed. `hooks/stop-context.js` already calls `recordHealth('writeSessionLog', status.project === true, ...)` on every invocation, and `flushHealth(cwd)` runs unconditionally (not gated on `AE_IDENTITY_DEBUG`), writing `.agentic/.telemetry-health.json`. Since `status.project` is exactly the append's return value, a give-up was already observable without a debug flag. No Python-side change needed.

**Caller-side retry judged not warranted**, documented inline: re-spawning the subprocess would double worst-case Stop-hook latency while racing the same contention the first attempt just spent its full budget failing to clear. The Python side is the right layer.

## Verification

A real forced-overlap subprocess race **cannot discriminate** the two versions - both bound total lock-wait to the same ceiling by construction, so a stall over it fails identically and a stall under it succeeds identically. The defect lives only in the narrow one-shot-vs-retry boundary, which a wall-clock race cannot reliably land inside without flakiness. That is stated plainly rather than papered over with a demonstration that proves nothing.

The deterministic equivalent was used instead: `_lock_fd_exclusive` monkeypatched to raise two forced timeouts before delegating to the real flock, run against the literal pre-fix commit via `git stash`.

- **Pre-fix:** `AssertionError: a transient lock timeout must not drop the record` - the defect reproduces.
- **Post-fix:** passes, with `attempts["n"] == 3` confirming the retry actually occurred and the record readable back from the log.

Plus an OS-level sanity check with a genuine `fcntl.flock` held for 2s by a second real process: the writer waited ~2.02s, acquired, and wrote.

Two regression tests added and wired into the runner - one pinning the retry path, one pinning that give-up stays soft-fail (returns `False`, never raises) and stays bounded.

`hooks/tests/test-stop-context-identity.js:891-937` is untouched and still asserts both the project-scoped and global-scoped logs receive both complete records.

## North Star alignment

**Pillar 2, verifiable outcomes** - the primary win. A silently dropped telemetry record produces no verifiable outcome at all, and the drop path is now retried and, on give-up, observable through the existing health record.

**Pillar 1, guard operator attention** - indirect but real: this class of flake reds CI on PRs that did not cause it, which costs an operator a diagnosis cycle each time.

**Pillar 3, low friction** - worst-case Stop-hook latency is unchanged; the ceiling was held at parity and backoff spends the budget rather than extending it.

## Test plan

- `pytest bin/tests/` - 1157 → 1159 (two new regression tests)
- `node hooks/tests/test-stop-context-identity.js` × 20 clean and × 20 under full-core CPU saturation - 20/20 both
- Hooks JS suite, verbatim CI loop - 45 files, 0 failures
- `test-enforce-turn-shape.py` 157, `test-turn-charge-model.py` 18
- methodology-drift, resident-budget, skill-embed-budget, symlinks-relative, command-file-budget - all OK

Doc-sync: predicate not triggered - no doc states a concrete lock-timeout figure (re-checked), and both the retry behavior and the health-record surfacing are internal mechanisms with no documented user-facing contract.

### DIFF
--- a/bin/ds-identity
+++ b/bin/ds-identity
@@ -135,6 +135,16 @@ PENDING_CAP = 100
 PENDING_SCAN_LIMIT = PENDING_CAP + 1
 PENDING_DIRECTORY_SCAN_LIMIT = (PENDING_CAP * 4) + 1
 PENDING_INTERNAL_STALE_SECONDS = 60
+# DS-158 round 2: total wall-clock budget for the session-log lock retry loop
+# in _write_jsonl_safely, spent across all attempts (not per attempt), not
+# per attempt in isolation. Kept at parity with round 1's flat single-attempt
+# 5.0s wait (not shrunk) - the fix here is that a lock timeout now gets a
+# retry at all within this budget, not that the budget itself grew, since
+# the JS caller's subprocess ceiling (see hooks/stop-context.js
+# writeTelemetrySafely's spawnSync `timeout` comment) sets the same total
+# ceiling either way. Bumping one obliges bumping the other, by construction.
+SESSION_LOG_LOCK_BUDGET_SECONDS = 5.0
+SESSION_LOG_LOCK_PER_ATTEMPT_CAP_SECONDS = 1.0
 
 
 def _contains_unicode_control(value: str) -> bool:
@@ -704,6 +714,15 @@ def _acquire_validated_file_lock(
         os.close(parent_fd)
 
 
+class _LockTimeoutError(RuntimeError):
+    """One flock() attempt exhausted its bounded wait.
+
+    DS-158 round 2: kept distinct from a bare RuntimeError so a caller-side
+    retry loop can catch specifically this transient, retryable condition
+    without also absorbing an unrelated RuntimeError it did not anticipate.
+    """
+
+
 def _lock_fd_exclusive(fd: int, *, timeout: float) -> None:
     """Acquire one fd lock with a monotonic bounded retry loop."""
     deadline = time.monotonic() + timeout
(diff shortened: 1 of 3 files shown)


tyson-solara6:
Superseded by round 3: shared deadline across both appends, partial-success checkpoint with an UNKNOWN health state, and tests pinning the wall-clock bound. Force-push is denied here.

PR #664, opened 2026-08-11
feat(DS-154): roll up prior-session shards at activation preflight (Unit F)

tyson-solara6:
Unit F of DS-154, and **the unit that actually retires the "remember to run `/ds-wrap`" problem.**

Merged: **A** `bin/ds-learning-shard`, **B** the capture instruction, **C** pointers in 18 agent files, **D** `SESSION_KEY` in spawn briefs. In review: **E** (rollup at Phase 11e, #662) which covers ticket sessions.

Ad-hoc sessions had no drain at all. This adds one.

`/ds-wrap` is untouched and is not going away. It stops being *necessary*, which was the goal.

## Why session start, not session end

It fires automatically on the very next session of any kind, on every one of the 11 adapters, because every adapter has an activation-preflight moment by definition. No daemon, no `--resume`, no harness-specific hook. That property is the whole point: an earlier design for this ticket was rejected precisely because it worked on one harness only.

## The step

New **Step 7** in `content/sections/01-activation-preflight.md`, running only when Step 4 resolved to active. Exactly one call, `ds-learning-shard rollup --repo <cwd>`, which prints a JSON array and exits 0 on every path.

- **Empty array is the common case: stop, print nothing, spawn nothing.** This runs at the top of every session, so anything chatty or expensive on an empty store would be resented and disabled.
- On a non-empty array, classify each entry through the existing guardrail-first table at `content/references/capture-classification.md` and forward only `Capture: MUST` to `learnings-agent`. No parallel classification path.
- Soft-fail absolutely: a missing binary, non-zero exit, or unparseable output is a silent no-op and never blocks session start.

Detail deferred to a new §Step 7 in `content/references/activation-detail.md`, keeping the always-loaded section to a trigger-pointer. The former Step 7 renumbered to 8, with the Step 4 breadcrumb updated.

## One design point worth checking

The call is **unscoped** - no `--session-key`. That is correct rather than lazy: at preflight the current session has appended nothing, so an unscoped rollup can only see prior-session shards, and the key of interest belongs to a session that has already ended. `.rolled-up.json` bookkeeping is keyed on the set of CLI-owned entry ids, so a repeat is a non-event rather than a double-emit.

## Verification

- `check-resident-budget` **OK, 391 B / 600 B, unchanged** - it measures `claude-managed-content.md`, not `content/sections/**`, so this addition does not touch it. Detail still went to a reference file rather than inline.
- `check-methodology-drift` OK after baseline regeneration (`content/sections/**` changed).
- `check-skill-embed-budget` 128,240 B, floor 100,000, ceiling 139,160, headroom 10,920.
- `bin/tests/` 1157 passed; hooks JS 47, hooks Python 17, hooks shell all pass.
- Adapter drift clean with the pathspec copied verbatim from `adapter-sync.yml`; `check-symlinks-relative` 4/4 relative; `check-codex-skill-sync` clean after a `scripts/codex-skills.py inventory` refresh per the documented procedure.
- `git merge-tree --write-tree origin/main HEAD` exits 0, and the merge **result** drops zero lines introduced by Unit D. A control against the raw branch tip reports 65, so that check is live rather than vacuously clean.

Known pre-existing failure, unrelated: `test_agentic_doctor.sh` T11 (Python 3.14 ANSI-colored argparse help).

## North Star alignment

- **Guard operator attention** - this is the step that means a forgotten `/ds-wrap` no longer costs a session's learnings. Nothing is required of the operator.
- **Works for everyone** - placed at the one moment every adapter has, using a plain CLI call. No capability, environment variable, or hook.
- **Verifiable outcomes** - the empty-store path is specified as silent and cheap, so the common case is observably free.

Doc-sync: triggered (`content/sections/`, `content/references/`); `docs/index.html`, `docs/slides/*.md`, `README.md`, `CONTRIBUTING.md` and `content/SKILL.md` checked. No update required: no reference file was added or removed, so the "37 reference docs" count holds, and the activation paragraph in `docs/index.html` covers config and marker resolution only.


tyson-solara6:
Superseded by a rebuild on current main: this branch conflicted with Unit E (#662) in the regenerated .codex RESOURCE-MAP files. Hand-authored content is unchanged. Continues in feat/ds-154-session-start-rollup-v2.

PR #675, opened 2026-08-11
fix(DS-return-contract r7): pin Shape-3 permissive-widening + doc numeral/nomenclature fixes

tyson-solara6:
## Summary

Round 7 (final round) on the subagent-return-contract Unit 0 branch. Closes one Major and two Minors from Skeptic review:

- **Major:** `SHAPE3_BARE_STATUS_TOKEN_RE` (added in round 6 to resolve a `BLOCKED`-line over-strictness Minor) had its strict direction pinned by the checker snapshot but its permissive direction unpinned - widening the regex from `r"^[A-Z][A-Z0-9_]*$"` to `r"^.+$"` (accepting arbitrary unbounded narrative on any non-`Label: value` line) left all 50 prior contract+mutation tests green. Adds a new mutation operator, `shape3_bare_status_token_regex_permissive_widening`, using a synthetic Shape-3 seed and a multi-word/mixed-case bare-line mutant. Verified both directions by hand: RED against the widened regex, GREEN against the restored narrow regex. Bumps the CI collect floor 25 -> 26 in `.github/workflows/bin-tests.yml`.
- **Minor 1:** Fixed a contradictory Shape-4 form-count numeral (prose said "exactly THREE" while enumerating four forms - 1, 2/3, 5) across `content/references/subagent-return-contract.md`, the three adapter copies, `.hermes/SKILL.md` (all regenerated via `bash scripts/build-all.sh`), and the matching code comment in `bin/tests/test_agent_return_contract_spec.py`.
- **Minor 2:** Retired stale "pointer" nomenclature left over from round 6's form-5 deletion in two present-tense docstrings, and renamed the round-5 mutation ID `shape2_pointer_requires_md_or_schema_reference` (which named a requirement round 6 deleted) to `shape2_undeclared_narrative_placeholder_not_recognized` - it still catches a genuinely unbounded narrative placeholder, only its old name was stale.

## Scope note

This unit ships the subagent-return-contract checker/gate only. All 87 real-corpus violations across 17 agent files surfaced by this checker are deliberately deferred to Units 1/3/4 - this PR does not fix any of them. `engineer.md`'s `pr_description_body` 2000-char cap is also Unit 1's scope, not this one's.

## North Star alignment

[...]

PR #677, opened 2026-08-11
fix(DS-163): make the W1 In-Progress tracker writeback observable

tyson-solara6:
## Problem

Operator report: "when I use `/ds-implement-ticket` the ticket is not getting transitioned to In Progress before the agent starts doing the work."

Investigation found the ordering was already correct. PR #517 moved the `In Progress` writeback ("W1") into Phase 1, ahead of Phase 2 (codebase read), Phase 3 (architect), Phase 3b (orchestration-planner), and Phase 5 (engineer), and left an anti-regression note pinning that Phase 5 has no W1 site.

The real defect is that W1 is unenforceable and unobservable. The Tracker Writeback Helper is spawned fire-and-forget; the conductor never awaits or verifies it. No hook, gate, or log records whether W1 fired, was skipped by its gate, or was never attempted at all. Worse, the `TRACKER == none` skip was *completely silent* - zero output, zero log line. To an operator that is indistinguishable from an ordering bug, which is exactly how it was reported.

## Change

- **A `tracker_writeback` breadcrumb** at the Phase 1 W1 site, covering three outcomes distinctly (`skipped`, `dispatched`, `dispatch_failed`) with a reason code on skip (`tracker_none`, `ticket_id_format`, `prefix_mismatch`, `fetch_failed`), resolved mutually exclusively. Emitted via the existing `bin/ds-emit` path, soft-fail throughout: a missing or failing `ds-emit` is a genuine no-op and never blocks Phase 1.
- **A one-line advisory** scoped to the `tracker_none` skip only, so the silent case becomes self-diagnosable. The other three skip reasons emit no advisory. It is not a stop-and-ask and not an `## Operator decisions` item.
- **Mechanism-pinned regression tests** in `bin/tests/`, plus an AST guard against duplicate test-function shadowing and a `bash -n` gate over the command file's fences with a precise allowlist of the 5 pre-existing value-position placeholders.
- Ordering is deliberately unchanged; the Phase 5 anti-regression note is now pinned by test.
- Doc-sync: `content/references/events-log.md`, `content/references/tracker-writeback.md` (the `TRACKER == none` skip is now documented as recorded, not silent), `content/sections/09-events-log.md`, `content/sections/12-protocol-details.md`, `docs/events-log-telemetry.md`, and `bin/ds-emit`'s manifest.

## Scope limit, stated plainly

This detects the case where the conductor **reaches** the W1 gate and the gate declines. It does **not** detect a conductor that never reaches the W1 prose at all - a breadcrumb emitted by that prose cannot record its own non-execution. Both `content/references/events-log.md` and `docs/events-log-telemetry.md` say so explicitly rather than implying full coverage. Closing that remaining case needs a deterministic hook, which is a separate piece of work.

## Review

Four Skeptic rounds. Round 1 found a Critical (`W1_DISPATCH_OUTCOME` referenced but assigned nowhere - the `dispatched` outcome was unrecordable, and the expansion could abort the step under `set -u`, defeating the "never blocks Phase 1" promise) and three Majors, the worst being that every regression assertion passed with the entire feature deleted. Round 2 verified the Critical fixed by independent mutation and found the coverage overclaim had relocated into the public doc. Round 3 found a false event-type count this diff itself introduced. Round 4 granted sign-off at 0 Critical / 0 Major after re-deriving the count, executing both emit fences under `set -euo pipefail` across all six outcome cases (exactly one event each, correct `outcome`/`reason`), and re-running six mutations to confirm each reddens a mechanism test.

Three Minors deferred by explicit decision: two bare expansions in `if` conditions (safe under the file's conductor-substitution model), an incomplete illustrative list at `ds-implement-ticket.md:1902` (already incomplete before this diff), and Phase 11e's missing `AGENTIC_LOOP_KEY` export (pre-existing on `main`, unrelated phase).

Rebased onto current `main` after PR #674 and #675 landed. The 5 conflicts were all generated files and were resolved by full regeneration, never hand-merged; both sibling changes verified to coexist, with no whole-file deletions and no hunk unrelated to this change.

## North Star alignment

Advances **verifiable outcomes**: a tracker transition that previously left no trace either way now produces a checkable record, and the most common silent-skip cause is self-diagnosable by the operator instead of presenting as a phantom ordering bug.

Does not regress **attention economy**: one advisory line on exactly one reason code, and nothing at all on the paths that were already correct.

Does not regress **works for everyone**: no dependency on a particular operator, tracker, workspace, or harness. Projects resolving `TRACKER=none` get one advisory line instead of silence; open-goal synthetic ticket ids still correctly never fire W1.

## Verification

- `pytest bin/tests/` 1242 passed; hooks JS (46 files), hooks Python (16), hooks sh, `bin-sh-tests` - all 0 fail
- `scripts/build-all.sh` then adapter-drift diff over the verbatim CI pathspec - clean
- `check-methodology-drift.sh` OK; resident 391/600 B; skill-embed 130,015 B (floor 100,000 / ceiling 139,160); command-file 358,790/371,000 B
- `.codex/skill-compatibility.yml` byte-identical to a fresh regen; all four skill symlinks relative
- `bash -n` over every fence: same 5 pre-existing failures as `main`, zero new
- All 5 commits DCO-signed

Doc-sync: triggered (new `tracker_writeback` event-type surface) -> updated `content/references/events-log.md`, `content/references/tracker-writeback.md`, `content/sections/09-events-log.md`, `content/sections/12-protocol-details.md`, `docs/events-log-telemetry.md`, `bin/ds-emit` manifest, and `scripts/.methodology-baseline.sha256`, plus regenerated adapter copies. `docs/index.html` and `docs/slides/*.md` grepped by hand - neither references the event-type set.


tyson-solara6:
Automated review pass (no code issues found — no formal verdict since this is our own PR).

One thing worth a human glance: the nit-scan pass flagged that `bin/ds-emit`'s header comment dropped the line referencing `content/rules/agent-methodology.md` (trivial-path edits) as a downstream consumer when the consumer list was updated in this PR. That's a content-accuracy question rather than a mechanical typo, so it wasn't auto-fixed — worth confirming whether that consumer reference should still be there.

tyson-solara6:
Superseded by the v2 branch: the stack needed rebasing onto current main plus a gitleaks allowlist fix, and force-push is blanket-denied in this repo.

PR #680, opened 2026-08-11
fix: single-source the shipped consumer .gitignore and make negation-drift detection behavioral

tyson-solara6:
## Problem

`/ds-init-project` Step 9 shipped consumer projects a **default-allow** enumerated gitignore denylist, while `content/project-scaffolding.yml` (the `ds-migrate` adoption route) used a **default-deny** umbrella. Two consequences:

1. The ephemeral set is open-ended, so a denylist can never be completed. Three consecutive sweeps each missed paths the next round found - including `.agentic/worktrees/` (full nested checkouts) and `.agentic/identity.yml` (carries a developer handle).
2. The two adoption routes disagreed by construction. `findings.md`, `learnings-agent.session`, `tier-map.yml`, and `presets.yml` were measured tracked via init and ignored via migrate.

Separately, prose at 15+ sites asserted files were "gitignored under the `.agentic/` umbrella". True inside DinoStack (which does use `/.agentic/*`), false in every consumer project. `content/commands/ds-init-project.md` already warned this bug class is invisible from inside this checkout.

## Change

- **Step 9 now delegates to `ds-migrate apply`** against the manifest, so both routes read one source and cannot diverge.
- **Drift detection is behavioral, not syntactic.** `ds-migrate check` asks git whether each manifest-negated path is ignored, rather than inferring it from gitignore pattern syntax. This closed five separate spelling evasions (comment-preceded umbrellas, root-anchored `!/.agentic/x`, `/**/` prefixes, `.agentic/**/**` suffixes, and body forms like `[.]agentic/*`), each of which had produced `check: ok` while knowledge files were silently uncommittable.
- **`apply` fails loudly (exit 3)** when the postcondition fails, and records a `[scaffolding-sync] FAILED:` notice naming the affected paths - previously a success notice was recorded while negations stayed defeated.
- Repairs a pre-existing bare `.agentic/` directory-form ignore, which defeats every negation and which `check` used to report as `ok`.
- False umbrella prose removed at 15+ sites, with a CI-wired gate (`scripts/check-no-false-umbrella-claims.sh`) against re-introduction.

## Known limit

For the 2 **directory-form** negations only (`!.agentic/session-log/` and its `/**` twin), probe paths must be synthesized. A defeater keyed to an unguessed name returns `ok` while no matching file exists yet. Real on-disk files are discovered and caught at depth 1-3, so it is self-healing on the next `check` and loses no data (session-log is append-only; worst case is one deferred Phase 8 cycle). The 11 exact-path negations need no guessing and are unaffected. This limit is stated in the function docstring, the operator-facing prose, and a pinned test. The closer - a point-of-use check at the commit site, where the exact filename is known - is deliberately follow-up scope.

## Verification

- 20-case gitignore-spelling matrix: zero false-OK; identical output across the round-11 and round-12 commits.
- Both adoption routes produce byte-identical outcomes; 13 knowledge paths tracked (incl. nested `session-log/`), 0 runtime files leaked; v5 and v6 both converge to v7; `apply` byte-stable across 3 runs.
- `check` verified strictly read-only (full-tree hash + `git status --porcelain --ignored` identical before/after, 80/80 cases).
- Oracle is `git add -A` + `git status --porcelain` on a fresh index throughout, never `git check-ignore` exit codes.
- Every fix mutation-proven red then restored. `pytest bin/tests/` 1211 passed; 43/43 shell tests; hooks JS suite clean apart from two failures reproducing identically on `main`.

Twelve adversarial review rounds; ten found something real.

## North Star alignment

- **Works for everyone (universality):** the core defect was DinoStack's own umbrella making consumer-project behavior invisible from inside this repo. Both adoption routes now derive from one manifest, so a teammate adopting via `ds-migrate` and one via `/ds-init-project` get identical results.
- **Verifiable outcomes autonomously:** drift detection observes real git behavior instead of modeling gitignore syntax, and a silent postcondition failure now becomes a durable, named record.
- **Guard operator attention:** honest claims replace overclaims - the known limit is documented where a reader will encounter it rather than asserted away.


tyson-solara6:
Superseded by #682. Same signed-off work (Skeptic sign-off at `a3279c7f`, 12 review rounds), ported onto current `main` because `gh pr update-branch --rebase` conflicted on four generated `.codex/skills/*/RESOURCE-MAP.json` files and force-push is unavailable here. Those artifacts were regenerated on the new base rather than hand-merged. Port equivalence verified mechanically: identical 102-file set and byte-identical per-file numstats against the signed-off diff.

PR #681, opened 2026-08-11
test(security): pin the gitleaks allowlist scope with a plant-and-scan guard

tyson-solara6:
## Problem

`.gitleaks.toml` configures the `scan` job, a **required status check** on `main`. PR #678 added a deliberately narrow allowlist to it for false positives in the generated `.codex/skill-compatibility.yml`. Nothing anywhere pinned that config: no test in `bin/tests/`, `hooks/tests/`, or `scripts/` referenced it. A future edit could silently widen the allowlist, or replace the triple-scoped entry with a blunt whole-file exemption, and no gate would notice. A silent hole in a required security control is the worst kind, because nothing ever re-surfaces it.

## Change

`bin/tests/test_gitleaks_allowlist_scope.sh` - a plant-and-scan guard that asserts by **running gitleaks**, never by reading the TOML. Each scenario builds a throwaway `git init` scratch repo, copies the real unmodified config in, commits a probe, runs `gitleaks git .` (matching the CI job's history-scan mode), and inspects the JSON report. The real repo's history and the real `evals/` tree are never touched; `evals/` is git-untracked here (decision #203) and invisible inside an isolation worktree, so every eval-boundary probe path is constructed fresh in the scratch repo.

Nine assertions across five groups:

- **0** - a correctly-shaped `source_token` pair IS suppressed. This one is load-bearing: the real file is JSON with double-quoted keys despite its `.yml` extension, and the allowlist regex anchors on that exact shape. A YAML-unquoted fixture never triggers the exemption at all, which would make every "still caught" assertion pass for the wrong reason.
- **1** - a non-exempt key (`other_token`) in the allowlisted file is still caught.
- **2** - a `source_token` secret in a different `.codex/` file is still caught (path leg).
- **3** - a `github-pat`-shaped value under `source_token` in the allowlisted file is still caught (rule leg).
- **4a/4b/4c** - the eval boundary holds: `evals/fixtures/` and `evals/skill-comparison/tasks/` suppressed, `evals/skill-comparison/` outside `tasks/` caught.
- **5a/5b** - a secret under a real-code path (`hooks/`, `bin/`) is caught. This closes a gap a reviewing Skeptic found by mutation: the first `[[allowlists]]` block is unconstrained, so adding `hooks/.*` to it silently exempts an entire directory from every rule, and the original 7 assertions all stayed green. That is exactly what the config's own comment forbids and exactly what this guard exists to catch.

Also: every suppression assertion now requires gitleaks to have actually produced a valid report, so a per-scenario tooling failure can no longer read as "suppressed".

`.github/workflows/bin-tests.yml` gains a gitleaks install step for the `bin-sh-tests` job, reading `GITLEAKS_VERSION` and `GITLEAKS_SHA256` out of `.github/workflows/gitleaks.yml` at runtime so the two pins cannot drift.

`.gitleaks.toml` is byte-identical to `main` - this adds coverage for that control, it does not revise it.

## Review

Two rounds. The first Skeptic ran 12 independent mutations rather than accepting the engineer's table and granted sign-off at 0 Critical / 0 Major, but surfaced the real-code-path gap above plus three accuracy defects (a header docstring claiming coverage the guard lacked, three individually-vacuous suppression legs, a mislabeled manifest dependency). All four are fixed here, each mutation-proven red-then-restored with `.gitleaks.toml` diffed byte-identical after every cycle.

Worth recording: dropping `targetRules` alone does **not** broaden the control, contrary to the original docstring. `github-pat`'s `Match` field is the bare token and never includes the `"source_token":` prefix, so the narrow regex leg structurally never matches it. `targetRules` is defense-in-depth, not the load-bearing leg. The header now says so.

Also confirmed by measurement, since it came up as a proposed alternative: renaming the `*_token` fields in `scripts/codex-skills.py` would **not** retire the allowlist. `gitleaks git .` scans full committed history, and every historical commit carrying the old field names stays in `main`'s history permanently. The allowlist is required regardless of what future commits name the field.

## North Star alignment

Advances **verifiable outcomes**: converts a required security control with zero regression coverage into one that is mutation-verified, so a future broadening reddens CI instead of landing silently.

Does not regress **attention economy**: no new operator-facing output; one shell test in an existing suite.

Does not regress **works for everyone**: no dependency on a particular operator, tracker, or workspace. The guard hard-fails under `${CI}` if gitleaks is absent rather than skipping silently, so it cannot go green having asserted nothing.

## Verification

- Guard: 9 passed under **bash** and 9 under **zsh** (this repo has shipped gates that check nothing under zsh while printing ALL PASS)
- `pytest bin/tests/` 1242 passed; full `bin/tests/test_*.sh` loop, 42 files, all pass
- Negative branches fired deliberately: malformed TOML and a forced scratch-commit failure both produce failures, not silent passes
- `git diff origin/main HEAD -- .gitleaks.toml` empty
- No em dashes in added lines; both commits DCO-signed

Doc-sync: predicate not triggered. No count, list, or path assertion in `README.md`, `CONTRIBUTING.md`, or `content/SKILL.md` references `bin/tests/` membership.

## Known follow-up, deliberately not in scope

The download/checksum/extract block is now duplicated between `gitleaks.yml` and `bin-tests.yml`. The drift-critical pin values are single-sourced and both copies verify the same checksum, so divergence cannot produce a wrong binary. Extracting `scripts/install-gitleaks.sh` becomes worth it if a third copy ever appears.

### DIFF
--- a/.github/workflows/bin-tests.yml
+++ b/.github/workflows/bin-tests.yml
@@ -139,6 +139,28 @@ jobs:
       - run: pip install pyyaml
       - name: install zsh (required for bash/zsh parity assertion in test_check_resident_budget.sh)
         run: sudo apt-get update && sudo apt-get install -y zsh
+      - name: install gitleaks CLI (required by test_gitleaks_allowlist_scope.sh)
+        # test_gitleaks_allowlist_scope.sh hard-fails under CI if gitleaks is
+        # missing on PATH (a silently-skipped security regression guard is
+        # indistinguishable from a passing one), so this step must always
+        # succeed here. Version and checksum are read from
+        # .github/workflows/gitleaks.yml at CI time rather than hardcoded, so
+        # bumping that pin cannot silently drift from this install step.
+        run: |
+          set -euo pipefail
+          version="$(grep -E '^[[:space:]]*GITLEAKS_VERSION:' .github/workflows/gitleaks.yml | head -1 | sed -E 's/^[^:]*:[[:space:]]*//')"
+          sha256="$(grep -E '^[[:space:]]*GITLEAKS_SHA256:' .github/workflows/gitleaks.yml | head -1 | sed -E 's/^[^:]*:[[:space:]]*//')"
+          if [ -z "$version" ] || [ -z "$sha256" ]; then
+            echo "ERROR: could not read GITLEAKS_VERSION/GITLEAKS_SHA256 out of .github/workflows/gitleaks.yml" >&2
+            exit 1
+          fi
+          tarball="gitleaks_${version}_linux_x64.tar.gz"
+          url="https://github.com/gitleaks/gitleaks/releases/download/v${version}/${tarball}"
+          curl -sSfL "$url" -o "$tarball"
+          echo "${sha256}  ${tarball}" | sha256sum -c -
+          tar -xzf "$tarball" gitleaks
+          sudo install -m 0755 gitleaks /usr/local/bin/gitleaks
+          gitleaks version
       - name: run shell bin tests
         # This step measured 4m56s against the old 5-minute JOB timeout on a
         # single PR, and was cancelled outright on main under queue pressure.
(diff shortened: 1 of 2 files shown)


tyson-solara6:
Superseded. The guard's probe fixtures were literal secrets, so gitleaks flagged the guard's own file and the required `scan` check failed. Rebuilt with runtime-generated probes on a clean branch - this branch is being deleted because its history permanently carries the literal fixtures and gitleaks scans committed blobs.

PR #686, opened 2026-08-11
feat(DS-return-contract Unit 3): migrate qa-engineer to Shape-2 pointer return, rewire Phase 8.5 evidence path

tyson-solara6:
Unit 3 of the subagent-return-verbosity plan. Migrates `qa-engineer.md` (the corpus's largest contract) to the Shape-2 pointer return and rewires the Phase 8.5 evidence pipeline that consumed its inline Screenshot Evidence JSON.

## What changed

- `qa-engineer.md` 907 -> 832 lines. Six near-duplicate per-method report-row templates collapsed to one canonical row plus a 7-row method-extras table. What remains per-method is distinct verification procedure (axe install gate, SB6 URL conversion, CDP motion emulation, pixelmatch baselines) and is irreducible.
- Return is now a bounded pointer object: `result`, `criteria[]`, `blocking_count`, `blocking_issues[]`, `server_status`/`auth` enums, `screenshot_evidence_json_path`, `report_path`, capped `notes`.
- Non-blocking Observations, the Regression Spot-check narrative, and "Likely area" fold to ADVISORY (report file only).
- Phase 8.5 now reads the screenshot evidence JSON from `screenshot_evidence_json_path` instead of parsing an inline block, with an `[]` fallback preserved for absent path / missing file / malformed JSON.
- Write-capable agent count 10 -> 11 across all five pinned sites.

## Critical fixed in review

The first round wrote the report and evidence JSON to a **relative** `.agentic/qa-reports/` path from inside a mandatory `isolation: "worktree"` spawn. The conductor resolved that path against its own checkout, always hit the `[]` fallback, and silently dropped every PR's screenshot evidence and regression drafts with nothing going red. `content/references/qa-gate.md:87` already documented the mechanism.

Fixed by writing to `/tmp/qa-reports/` (host-level, the same property `/tmp/qa_*.png` already relies on) and returning the absolute path. Verified empirically, not by code read: report written from inside a worktree, worktree removed, file read successfully from the primary checkout. No cross-run collision - the path is `${TICKET_ID}-<ts>-$$` and every read and both `rm`s use the exact returned path, with no glob anywhere.

## Accepted debt (recorded, not deferred silently)

Shipping with these open under the round-budget policy being introduced in a sibling PR. None affects current correctness; all are protection for a fix already verified.

1. **No regression pin on the `/tmp` write path.** Reverting every `/tmp/qa-reports` back to `.agentic/qa-reports` leaves all 1251 `bin/tests` green. A ~10-line pin asserting an absolute write path for any `isolation: "worktree"`-mandated agent would catch it. Folds naturally with item 2.
2. **No gate pins the `screenshot_evidence_json_path` producer/consumer coupling** between `qa-engineer.md` and Phase 8.5. A rename on either side yields dead evidence links with nothing red.
3. **Partial `/tmp/qa-reports` leak across non-final FAIL iterations** - the cleanup block runs at loop exit while the path vars are overwritten each iteration. Fails safe (leak, not premature delete).
4. **The cleanup ordering constraint is stated three times** (~1,050 B) in the file carrying a dedicated byte ratchet.

Items 1 and 2 are batched into the follow-up that clears the remaining 25 snapshot violations, rather than each earning its own engineer + Skeptic cycle.

## North Star alignment

`docs/overview/vision.md` Goal 1: **"Guard operator attention. Surface decisions and work-stoppages, not status. A change that adds capability but increases what the operator must read, watch, or babysit is a regression, not a feature."**

This unit retires 194 lines of near-duplicate report scaffolding and moves QA narrative out of the conductor's context into a report file, while keeping every work-stoppage mechanically present in the return - failing criteria identity and reason in `criteria[].note`, blocking issues in `blocking_issues[]`, environment failures as `server_status`/`auth` enums. That distinction is the point: compressing status is the win, and relocating a work-stoppage would have been a regression wearing the same shape.

Goal 2 (verifiable outcomes) is served by the Phase 8.5 rewire being proven with an executed before/after dry-run rather than a code read, per the plan's own requirement for its highest-risk edit.

No pillar regresses. `docs/overview/` is not modified.


tyson-solara6:
Superseded by #689 - ported to a fresh branch cut from current `main` to resolve generated-artifact conflicts (no source change; verified numstat-identical against this branch's signed-off diff). Closing.

PR #695, opened 2026-08-12
fix(codex-skills): drop the repo-wide inventory hash that conflicted RESOURCE-MAP.json on every base move

tyson-solara6:
## Problem

`scripts/codex-skills.py` computed `inventory_hash = sha256(canonical_json(load_compatibility(repo)))` - a **whole-repo** digest over every `documents(repo)` entry - and embedded that single value into all four `.codex/skills/*/RESOURCE-MAP.json` files as one `canonical_json` line.

Any `content/` edit rewrote the same line in all four files, so two branches touching unrelated prose conflicted on every base move. A `brief`-skill map changed because someone edited `content/sections/04-risk-classification.md`.

Measured cost: one PR needed four ports and burned four PR numbers (#680, #682, #690, merged as #691). An independent session hit the identical wall - #689 is titled "port to current main to resolve generated-artifact conflicts". Force-push is blanket-denied here, so the only remedy each time was a full re-port.

## Change

Removed the field. Nothing reads it: the sweep found only the write site plus the four generated files; `validate_resources()` and `bin/ds-codex-dispatch` read only `payload["resources"][logical]`; `preamble()` never mentions it. (The similarly-named `source_inventory_sha256` is a different key in a `.agentic/` journal and is untouched.)

Drift detection is unaffected - `check()` calls `load_compatibility(repo)` itself before invoking `render_tree`, so the guarantee never depended on the value being written into the committed maps.

## Verification

- **Conflict, before/after:** two branches each adding a tracked token to a different `content/commands/*.md`, through the real regen workflow. Unfixed base: `CONFLICT (content)` in all four RESOURCE-MAP files. Fixed base: zero files touched, merge-tree rc=0.
- **Drift still caught:** mutating an `occurrence_hash` in `.codex/skill-compatibility.yml` still fails `codex-skills.py check` with `compatibility inventory drift: 1 unmapped/new and 1 stale occurrences`.
- `build-all.sh` (11 adapters) then `git diff --exit-code` over the verbatim 16-path adapter-sync pathspec: rc=0. `.codex/skill-compatibility.yml` 936,262 B. `scripts/test/test_codex_skills.py`: 107 tests OK. `pytest bin/tests/`: 1289 passed.

## Review rigor

- Brief / Plan path: none - scoped from an investigation that traced the mechanism with citations.
- Skeptic rounds (tier): 1 round, Tier 3, under an explicit round cap of 1.
- Findings summary: 0 Critical, 0 Major, 3 Minor, sign-off granted. The Skeptic's own first conflict reproduction was vacuous (it appended a non-token comment, so the inventory never changed and the merge was clean); it caught that itself and redid the repro through the real workflow. Three Minors ship as accepted debt: an uncommented side-effect `load_compatibility` call that reads as dead code, no test pinning `resource_map()` output as inventory-independent, and `schema_version` staying 1 across a field removal with the rationale only in the commit message.

## North Star alignment

- **Shorten wall-clock time:** removes a defect that cost four ports on a single PR and hit at least one other session independently. Any long-lived branch stops paying it.
- **Guard operator attention:** repeated mechanical re-ports are pure overhead with no review value.
- No pillar regressed; drift detection is provably unchanged.


tyson-solara6:
Superseded by the -v2 port - identical signed-off content re-cut onto current `main` after the generated-artifact conflict recurred (the very conflict this PR removes). Equivalence verified: identical 6-file set and byte-identical numstats.

PR #715, opened 2026-08-12
fix(agents): single-source shared check wording via content-layer transclusion

tyson-solara6:
## Summary

Makes the wording shared between `content/agents/engineer.md`'s pre-submit block and `content/agents/skeptic.md`'s numbered steps have exactly **one** source, enforced mechanically.

- **New canonical file** `content/fragments/pre-submit-check-kernels.md` - five hand-edited fragments (`test-file-glob-list`, `identifier-rename-trigger`, `identifier-type-list`, `rename-exemption-clause`, `async-primitive-list`).
- **New `scripts/stamp-agent-fragments.sh`** stamps each fragment into every `<!-- shared:<id> -->...<!-- /shared -->` span in `content/agents/*.md`. Idempotent; exits nonzero on a `shared:` marker with no matching `FRAGMENT:` definition rather than silently skipping.
- **Wired as the first step of `scripts/build-all.sh`**, before the adapter loop, so every adapter build sees already-stamped sources and `build-all.sh` remains the single local entrypoint.
- **New CI job** re-stamps and `git diff --exit-code`s the two agent files, so a future hand-edit that skips the script goes red instead of silently diverging.
- **New `bin/tests/test_engineer_skeptic_check_kernels.py`** covers what transclusion structurally cannot: that `engineer.md`'s "step 4.5 / 4.6 / 11.5 / 7" citations still resolve to correctly-titled steps in `skeptic.md`, and that `architect.md` still carries its "Per-consumer impact table" heading.

Both agent files remain complete, readable prose at every commit - the markers wrap live text, never a blank template, because these files are loaded verbatim into agent spawns.

## Why

The duplication was not hypothetical. After the original write (#702), a full Skeptic review, **and** a targeted three-finding fix round, one word still differed between the two copies: `skeptic.md` read "renames **that** nothing outside the function can reference" while `engineer.md` had dropped the "that". Three passes by three different agents failed to catch a one-word divergence. That is the argument for mechanical stamping over careful re-reading, made by the artifact itself. `skeptic.md`'s wording is canonical wherever the two historically disagreed, so this PR also fixes that discrepancy.

## Alternatives rejected

- **Runtime pointer form** (engineer reads `skeptic.md` on demand): non-portable. A bare `content/agents/skeptic.md` reference resolves only when the reading agent's cwd is the DinoStack checkout - false for a consumer project's engineer spawn - and the `~/.claude/agents/` form is Claude-Code-specific and wrong on the other 10 adapters. Fails the universality pillar.
- **Adapter-output-layer transclusion**: blocked by the adapter mix. `.codex` generates TOML, `.gemini` frontmatter-md, `.kimi` symlinks the directory, `.claude` hardlinks. Stamping at the `content/` layer before any adapter runs sidesteps all of it.
- **Sync gate only, no transclusion**: detects divergence but does not satisfy "one source". Its two orthogonal address-accuracy tests were kept, since transclusion cannot solve a renumbered or retitled step.

## Verification status - please read

The implementing engineer was terminated by a session limit immediately after pushing, **before** completing its final verification pass. Its gate results are therefore unproven and are not claimed here. CI on this PR is the first real adjudication, and a Skeptic review is running in parallel. Do not merge on the strength of the description above.

The mechanism itself was inspected directly: the fragments file, the `build-all.sh` wiring (stamp runs first, before the `ADAPTERS` loop), and the restored "that" in the exemption clause are all confirmed present in the pushed commit.

## North Star alignment

**Advances Pillar 7 (prevent defects at the producing step):** this is that pillar applied to the methodology's own source. A drift class that three review passes missed is converted into a build step plus a red CI check, so it cannot recur rather than being caught later.

**Advances Pillar 2 (produce verifiable outcomes autonomously):** the sync gate is mechanical and self-enforcing; correctness stops depending on a reviewer noticing a one-word difference.

**Pillar 5 (guard agent context):** essentially neutral. The stamped text is the same text that was already inline; the only addition to the resident set is the marker comments themselves, which are a few bytes per span.

**Pillar 4 (works for everyone):** actively protected - the content-layer approach was chosen precisely because the pointer alternative would have worked only inside the DinoStack checkout on one harness.

**Named trade-off:** the `<!-- shared:... -->` markers are permanently visible to any agent reading its own spawn file raw. Judged acceptable against silent unenforced drift.

## Doc-sync

`Doc-sync: content/agents/** changed. The stamped text is byte-identical to what was already published apart from the restored "that" in the exemption clause, so no doc states anything newly false. Pending confirmation of the docs-currency grep over docs/slides/*.md, which the engineer did not reach before termination - the Skeptic has been asked to perform it independently.`


tyson-solara6:
Superseded by the v2 branch: rebasing onto current main rewrote every SHA on this branch, and force-push is blanket-denied in this repo, so per the supersede-don't-force convention this PR is closed in favour of the rebased branch carrying all three Skeptic Majors closed.

PR #723, opened 2026-08-12
fix(telemetry): resolve harness config dir instead of hardcoded ~/.claude for subagent transcripts

tyson-solara6:
## Summary

`.agentic/events.jsonl` token telemetry has been unconditionally zero since inception. The cause was not a harness limitation - it was a hardcoded path.

`bin/ds-parse-subagent-usage` resolved `~/.claude/projects/...` unconditionally and ignored `CLAUDE_CONFIG_DIR`. On any machine that redirects the harness config dir (this one points at `~/.claude-moment8`), the tool looked in an empty directory and found nothing. The on-disk layout was the standard one all along; only the config dir differed.

- New shared config-dir resolver in `bin/_lib.py` and `hooks/lib/config-dir.js`, using the same precedence already established by `ds-identity`'s `PROFILE_CONFIG_DIR_ENV`: `AGENTIC_CONFIG_DIR` > `CLAUDE_CONFIG_DIR` > `CODEX_HOME` > `PI_CODING_AGENT_DIR`, falling back to `~/.claude`.
- `bin/ds-parse-subagent-usage` resolves both its primary path and its glob fallback through it.
- `hooks/subagent-stop-spawn-emit.js` resolves and reads the subagent's own transcript and populates real `data.tokens`.
- `hooks/stop-context.js` and `bin/ds-cost` surface those tokens in the ad-hoc aggregation path instead of hardcoding `None`.
- Four stale "harness ceiling" claims corrected across code comments and docs.

## The claim this corrects

`hooks/subagent-stop-spawn-emit.js` asserted `tokens_note: "unavailable (harness)"` unconditionally, and `content/references/events-log.md`, `docs/events-log-telemetry.md`, and a `hooks/stop-context.js` comment repeated it. That claim was false, and because it lived in committed code it read as authoritative - an investigation into this gap initially concluded "not viable at the hook layer, harness ceiling" by citing the very comment that was the origin of the belief. Two sources agreeing is not corroboration when one is downstream of the other.

Measured directly against live transcripts before any code was written: agent `a05445309a49bb49b` -> 18 usage blocks (output 5,130, cache_creation 257,315, cache_read 826,434); agent `ae09d206d8c384e59` -> 47 blocks (output 14,251, cache_creation 347,621, cache_read 3,725,144).

## Never-fabricate discipline

`data.tokens` is **absent, never zero-filled**, when unresolvable, and `tokens_note` carries a descriptive reason instead. The two are mutually exclusive. A transcript at or above 20 MiB is skipped entirely rather than partial-summed - the same never-fabricate principle as the existing `MAX_SANE_WALL_SECONDS` null-not-clamp behaviour. A zero that reads as a measurement would be worse than the honest zeros this replaces.

All new logic sits inside the existing try/catch with `process.exit(0)`; a transcript-read failure can never block a subagent's stop. The single `appendFileSync` is unchanged, so the documented TOCTOU race is not widened.

## What was deliberately NOT changed

`scanSessionAggregate()`'s double-count guard is untouched. Measurement showed `wall_seconds` is already working (732 of 756 post-DS-160 records nonzero) and the "ticketed sessions still get zeros" premise was stale - the conductor-emitted `spawn_complete` path went extinct on 2026-07-10, a month before the DS-160 cutoff, so `hasConductorSpawnComplete` is never true in current data. Redesigning it would have risked reopening the exact Critical that `test-stop-context-telemetry.js` Test 9 exists to prevent, for zero measured benefit.

`hooks/lib/capture-gap.js`'s "harness ceiling" comment is about `findings_count`/`signed_off` and was left alone. `hooks/pre-tool-use-spawn-emit.js`'s equivalent claim for `spawn_start` remains accurate - tokens genuinely do not exist yet at spawn-launch time.

## Verification

Every functional fix carries a break/restore mutation proof:

| Mutation | Result |
|---|---|
| Restore hardcoded `~/.claude` in `ds-parse-subagent-usage` | config-dir regression test FAILED, restored -> GREEN |
| Revert token wiring to the old unconditional `tokens_note` literal | 4 of 22 new tests FAILED, restored -> 22/22 GREEN |
| Revert `ds-cost` enrichment to `tokens=None` | regression test FAILED (n/a instead of real figures), restored -> GREEN |

End-to-end smoke, not simulated: a real subprocess invocation of `hooks/subagent-stop-spawn-emit.js` against a planted transcript under `CLAUDE_CONFIG_DIR` emitted `spawn_complete` carrying `data.tokens` and no `tokens_note`. The inverse - `ds-parse-subagent-usage` with no config-dir vars set - returned `transcript_not_found` at rc=0.

The `CLAUDE_CONFIG_DIR` test plants a decoy under `HOME` so it fails if the resolver falls back rather than honouring the env var.

`pytest bin/tests/` 1429 passed, 25 subtests. Hooks JS 48 files, Python 16, sh 1 - zero failures. Adapter drift over the 16-entry pathspec: only the expected `events-log.md` and `.hermes/SKILL.md` regeneration.

## North Star alignment

**Advances Pillar 4 (works for everyone), which is the heart of this fix.** The bug was invisible on a stock `~/.claude` setup and only manifested for operators who redirect their harness config dir - a supported configuration this repo's own `ds-identity` already handles. Telemetry is now accurate regardless of local harness layout rather than only on the default path.

**Advances Pillar 2 (produce verifiable outcomes autonomously):** per-task token cost had no data source; `/ds-cost` could only report n/a. It can now report measured figures where a transcript resolves, and still says n/a honestly where one does not.

**Pillar 5 and 6:** a bounded synchronous transcript read is added per subagent completion, capped at 20 MiB with a 200-entry glob bound, inside a 5s hook timeout. Measured transcripts were far under the cap.

**No pillar regressed.** The failure mode is strictly toward honesty: unresolvable stays absent-with-a-reason, never a fabricated zero.

## Doc-sync

`Doc-sync: content/references/events-log.md changed -> updated docs/events-log-telemetry.md in the same PR and regenerated the adapter copies via scripts/build-all.sh. Not a content/sections/** change, so scripts/.methodology-baseline.sha256 is unchanged (verified). No README/CONTRIBUTING/content/SKILL.md counts affected.`


tyson-solara6:
**Automated review (informational only - self-authored PR, no formal verdict per review policy)**

Ran an adversarial correctness/security read (skeptic) plus a North Star alignment check on this PR.

- **Correctness/security:** No Critical/Major issues. This PR is well-hardened - the round-2/round-3 fix-doc comments indicate prior review already caught the Python/JS resolver divergence and fabricated-zero token issues, and the new `bin/tests/test_config_dir_env_sync.py` drift-gate mechanically pins the env-var precedence lists (`AGENTIC_CONFIG_DIR > CLAUDE_CONFIG_DIR > CODEX_HOME > PI_CODING_AGENT_DIR`, fallback `~/.claude`) across `bin/_lib.py`, `hooks/lib/config-dir.js`, and `ds-identity` so they can't silently re-diverge. Verified this directly rather than trusting the commit narrative. Never-fabricate discipline for tokens (absent-with-note vs. zero) is correctly implemented.
- **Minor (informational, not blocking):** `hooks/subagent-stop-spawn-emit.js`'s project-dir fallback scan is bounded at 1000 entries while the Python-side glob in `bin/ds-parse-subagent-usage` is unbounded - a documented, deliberate tradeoff, but a real cross-language asymmetry worth tracking if project-dir counts grow. Also, `hooks/lib/config-dir.js`'s `absolutize()` doesn't expand `~user/...` the way Python's `os.path.expanduser` does - documented as an intentional fail-safe gap (resolves to a nonexistent path rather than a wrong one), not expected in practice for these env vars.
- **North Star:** ALIGNED - advances Pillar 2 (verifiable outcomes) by making telemetry honestly report absent data instead of fabricated zeros, without adding operator attention burden.

No action needed; this is a record for the review log.

tyson-solara6:
Automated PR review (no formal verdict - self-authored PR).

No trivial mechanical issues found - no direct edits made.

**Pre-existing, out-of-scope findings** (not this PR's responsibility - noting here since automated Jira filing is currently unavailable in this environment, so these aren't tracked elsewhere yet):
1. `hooks/subagent-stop-spawn-emit.js` has an explicitly-documented, pre-existing lack of serialization between concurrent `SubagentStop` invocations (a prior Skeptic "Minor" finding) - two subagents finishing close together could race on `appendFileSync`. This PR substantially expands that file (~280 lines, including a synchronous transcript read per invocation), which marginally widens the window. Worth a tracked follow-up ticket to add file locking or formally accept the risk via a decision record, since telemetry accuracy is the point of this whole PR line.
2. Cross-language `~user` home-directory expansion divergence between `bin/_lib.py::resolve_claude_config_dir` (uses `os.path.expanduser`, resolves `~otheruser/...`) and `hooks/lib/config-dir.js::absolutize` (leaves `~otheruser` untouched). Low priority per the PR's own comments, but worth a lightweight follow-up ticket so it isn't "rediscovered" later.

No other issues found. All CI checks pass; North Star alignment check: ALIGNED (fixes broken telemetry so `/ds-cost` reports verifiable data instead of silent zeros).

PR #732, opened 2026-08-13
feat(hooks): add enforce-worktree-write.py, write-side worktree isolation guard

tyson-solara6:
## Summary

Adds `hooks/enforce-worktree-write.py`, the write-side companion to `enforce-worktree-read.py` (DS-150). It denies a worktree-isolated subagent's `Write`/`Edit`/`MultiEdit` when the target resolves inside the primary checkout instead of the agent's own worktree.

## The gap this closes, with real instances

`enforce-shippable-edit.py` cannot catch this. Its deny axis is conductor-vs-subagent: it denies only when `agent_id` is **absent**. A subagent that has silently fallen back to the primary checkout still carries a present `agent_id` and sails through.

`AGENTS.md:61` already documents the mechanism that produces the fallback: across sequential spawns in one task, the isolation worktree is cleaned up between spawns and later Workers revert to the main tree, indistinguishable from a legitimate isolated engineer in the hook payload.

Three commits in this repo's own history landed shippable edits directly in the primary checkout by exactly that route: `1577c984` and `3142a803` (edits to `content/commands/ds-implement-ticket.md`, `.claude/commands/*.md`, plus adapter regenerations) and `530ad687` (`scripts/test/*.py`). This is a demonstrated failure, not a hypothetical one.

## Why this is not "gating on inferred session capability"

The guard reads two structural facts straight from the payload and compares them. `primary_root` comes from `CLAUDE_PROJECT_DIR`; `caller_root` comes from the payload's own `cwd` field. Those are complementary, not interchangeable - taking both from `CLAUDE_PROJECT_DIR` would compare the primary root against itself and the guard would never fire. Nothing is inferred about what the session *can* do.

Deny requires all of: `agent_id` present and non-blank (subagent call); `caller_root` is a proper realpath-normalized subdirectory of `primary_root` (genuine isolation); the resolved target is inside `primary_root` but outside `caller_root`; and no configured exemption matches.

## Design choices

**Separate hook, not merged into `enforce-shippable-edit.py`.** Different deny axes, and that hook is fail-open-CRITICAL - a false deny there blocks every conductor `Write`/`Edit`/`MultiEdit` for the whole session. Its design must not be perturbed.

**Separate config key** `worktree_write_guard_exemptions`, deliberately distinct from the read guard's `worktree_read_guard_exemptions` - writes carry a different risk profile and should not inherit read exemptions.

**Kill-switch** `AE_WORKTREE_WRITE_GUARD_DISABLE=1`. Registered with the guarded `test -f <path> && python3 <path> || exit 0` form, because a bare `python3 <missing>.py` exits 2, which is BLOCKING on PreToolUse - a registration outliving its script would deny every guarded call in every session.

Fail-open on every other path, matching the read guard's failure-mode table.

## Verification

**Mutation-proven, not asserted.** Neutering the `caller_root == primary_root` isolation check to `if True:` dropped 14 of 38 test cases to RED; restoring returned all 38 to GREEN. The suite covers the full matrix: cross-boundary write denied, own-worktree write allowed, main-session payload (`agent_id` absent) allowed unconditionally, `CLAUDE_PROJECT_DIR` unset fails open, missing payload `cwd` fails open, exemption match allowed, kill-switch allowed - across all three of `Write`, `Edit`, and `MultiEdit`.

Direct smoke test of the hook confirmed a cross-boundary write denies with the correct diagnostic reason string, and an own-worktree write produces no output.

`pytest bin/tests/` 1451 passed. Hooks Python 17 files, JS 49 files plus both wrap-lock suites, sh 1 non-quarantined file - all green. Adapter drift over the 16-entry pathspec: exit 0. Methodology-baseline drift: OK, regenerated after the `content/sections/04-risk-classification.md` edit.

## Count-sync

The new config key moves the behavioral-toggle count from twenty-two to twenty-three, which this repo has repeatedly gotten wrong because the count appears in four surface forms and a single-form grep returns clean. Every site was swept and bumped: `content/references/risk-config-and-tiers.md`, `content/references/conventions-detail.md`, `content/sections/04-risk-classification.md`, `README.md` (both prose sites), `docs/components.md`, `docs/configuration-reference.md` (the only numeral-form site). Both `.agentic/config.json` seed sources were updated - the DS-150 failure was shipping the key to one seed and not the other, producing a 21-key seed against 22-toggle doc claims.

`_ENFORCER_SUBCOUNT_SITES` and `TOGGLE_COUNT_FILES` in `bin/tests/test_tracker_writeback_ranking_spec.py` were updated in the same commit, so the gate that polices these counts moves with them.

## North Star alignment

**Advances Pillar 7 (prevent defects at the producing step):** the worktree-isolation invariant is currently prose that must be remembered across sequential spawns. This makes it mechanical, so the silent fallback becomes a caught error instead of a contamination incident discovered later in a reflog.

**Advances Pillar 2 (produce verifiable outcomes autonomously):** an agent's sandbox boundary stops depending on the harness happening to preserve a worktree between spawns.

**Pillar 1 (guard operator attention):** the deny message names the concrete diagnosis and prints `caller_root`, so the operator gets an actionable cause rather than a generic denial.

**Pillar 4 (works for everyone):** exemptions are per-project config, the kill-switch is env-based, and the guard fails open on every unexpected condition - a machine with an unusual layout degrades to current behavior rather than breaking.

**Named trade-off:** this is near-duplicate code with `enforce-worktree-read.py` by deliberate instruction. The two guards must stay independently editable, and the read guard's fail-open discipline should not be coupled to the write guard's. No shared module was extracted.

## Doc-sync

`Doc-sync: triggered - new hook, new config key, and toggle-count bump required updates across hooks/AGENTS.md, content/references/{delegation-detail,events-log,risk-config-and-tiers,conventions-detail}.md, content/sections/04-risk-classification.md, README.md, docs/{components,configuration-reference,index.html,safe-configuration}.md, and both .agentic/config.json seed sources. All adapter copies regenerated via scripts/build-all.sh and verified against the check-adapter-sync pathspec; scripts/.methodology-baseline.sha256 regenerated.`


tyson-solara6:
**Automated review (informational only - self-authored PR, no formal verdict per review policy)**

Ran a security-auditor + skeptic pass on this write-side worktree isolation guard.

- **Security correctness:** `hooks/enforce-worktree-write.py` resolves all paths with `realpath()` and does segment-aware containment checks (not string-prefix matching), so `../` traversal and symlink escapes are correctly handled. The conductor/subagent/worktree-isolated decision matrix looks correct in all three cases.
- **Design tradeoff (not a bug):** the hook is deliberately fail-open on malformed/ambiguous input, mirroring the existing `enforce-worktree-read.py` sibling - a conscious choice to avoid a fail-closed PreToolUse hook blocking all conductor writes session-wide on any edge case. Flagging as an accepted risk for visibility, not requesting a change.
- **Tests:** `hooks/tests/test-enforce-worktree-write.py` (38+ cases) covers symlink escapes, relative traversal, non-canonical symlinked roots, and the ordinary-subdirectory false-positive case. PR body's mutation-testing claim (14/38 cases red when the core check was neutered) checks out as solid verification discipline.
- **Consistency:** spot-checked `.claude/` and `.codex/` adapters - the new config key, toggle count, and kill-switch env var are propagated consistently with the canonical `content/` source.
- **North Star alignment:** ALIGNED - makes a previously prose-only isolation invariant mechanical, advancing verifiable-outcomes without adding operator overhead.

**Blockers:** `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY` against `main` - needs a rebase. Also, at last check only the Vercel preview check had run; the actual test-suite/CI workflow didn't appear to have fired yet - worth confirming it runs green post-rebase before merging.

tyson-solara6:
Superseded by the v3 branch. This PR went CONFLICTING after concurrent hook PRs (#729, #733) merged into the same enforcer-count prose sites and regenerated adapter artifacts; force-push is blanket-denied here, so per the supersede-don't-force convention it is replaced by a branch cut from current main with all counts recomputed and all generated artifacts regenerated rather than hand-merged. Same Skeptic-signed-off substance, verified file-set and per-file-numstat identical.

PR #752, opened 2026-08-15
fix(adapter-hooks): anchor .agentic writes in Cursor, Copilot, and OpenCode hooks

tyson-solara6:
Closes the last remaining path to the phantom-`.agentic/` bug.

## Why this exists

PR #745 anchored every `.agentic/` write to the resolved repo root, and #750 added the tool that cleaned up the damage (39 phantom trees, 430 recovered telemetry records in one consumer repo). Neither covered the adapter hook copies: they are hand-authored and `build.sh` does not regenerate them, so they inherited nothing.

Concretely, Claude Code users were protected and Cursor, Copilot, and OpenCode users were not. The same stray `cd` still produced phantom trees under those harnesses.

## What changed - 11 sites across 6 files

**JS (3 files)** - `.cursor/hooks/stop-context-cursor.js`, `.copilot/hooks/stop-context-copilot.js`, `.github/hooks/stop-context-copilot.js` now `require('../../hooks/lib/repo-root.js')` and skip the write when no `.git` ancestor is found. These files already required `stdin-guard.js` from the same shared lib dir, so this follows the established pattern.

**Bash (2 files)** - `.copilot/hooks/session-start-copilot.sh` and `.github/hooks/session-start-copilot.sh` now source `hooks/lib/repo-root.sh`. Their `CWD="$(pwd)"` fallback was itself a second instance of the bug, not merely the join beneath it - the same shape as `enforcement_log.py`'s `os.getcwd()` fallback that needed its own fix in #745. Neither falls back to `pwd` any more.

**TypeScript (6 sites in 1 file)** - `.opencode/plugins/session-context.ts`. The drift-injection point was `directory || process.cwd()` at the `session.idle` and `command.executed` handlers, feeding all six write sites downstream.

`.copilot/` and `.github/` are hand-duplicated byte-identical copies, not generated. Both fixed; verified identical before and after each edit. Not deduplicated here - that is a separate decision.

## OpenCode: inline, not shared

`session-context.ts` is a standalone Bun plugin loaded from `~/.config/opencode/plugins/` with no access to the repo's `hooks/` tree, so it carries an inline `resolveAgenticRoot` rather than requiring the shared module. Semantics mirror `hooks/lib/repo-root.js` exactly: realpath-pinned, walk up to the nearest `.git`, **existence-only** check (a linked worktree's `.git` is a FILE - an `isDirectory()` test fails in the most common execution environment here), MAX_DEPTH 64, never throws.

Written with Node built-ins only rather than Bun APIs specifically so it can be executed under test instead of only regex-matched. `hooks/tests/test-opencode-agentic-root.js` extracts and runs the real function, including the linked-worktree case.

## The durable half

The coverage gate that prevents a new unanchored writer from being added silently did not scan adapter dirs. That omission is precisely why these 11 sites survived #745. `SCAN_DIRS` now includes `.cursor/hooks`, `.copilot/hooks`, `.github/hooks`, `.gemini/hooks`, `.kimi/hooks`, `.opencode/plugins`, and `.ts` was added to the candidate extensions.

Scoped to each adapter's hook/plugin subdirectory, never the adapter's full tree - those hold a generated verbatim `content/**` copy including two literal `templates/.agentic/` scaffold dirs, which would flood the scan with prose matches. Each of the six directories was verified to hold only executable hook code today.

Inventory regenerated mechanically via the scanner's own `_find_candidates()`: exactly 12 new lines, nothing else. `.gemini` and `.kimi` hooks correctly produced zero candidates (HOME-scoped or unrelated).

**Reddening mutation proven:** appended an uninventoried `.agentic` construction to `.kimi/hooks/session-start.sh`, confirmed the gate failed naming the exact line, removed it, confirmed green.

## Tests

Each fixed site has a regression test firing the write from a drifted cwd inside a fixture repo, asserting `.agentic` lands at the fixture root. Each was confirmed failing pre-fix by mutating the fix back to its original shape. Fixtures create `<fixture>/.git` explicitly - without it the resolver fail-opens and a pre-fix assertion fails for the wrong reason, reading as a false reproduction.

`hooks/tests/test-session-start-copilot.js` is new: that script had zero prior coverage. The shared `mkFixtureDir` helper in the cursor test file gained `.git` creation, because every pre-existing test there would otherwise have silently no-opped after this change.

## North Star alignment

**Pillar: mechanical enforcement over remembered rules.** The gate extension matters more than the 11 fixes. Those sites were missed because they sat outside the scanner's scope, not because anyone forgot them - and a hand-maintained list of write sites has now been wrong four separate times in this work.

**Pillar: works for everyone.** The methodology ships six harness adapters; fixing only the one the maintainer happens to use leaves every other user exposed to a bug we have already diagnosed.

**Cuts against: single-sourcing.** OpenCode gets an inline resolver duplicating the shared contract, which can drift. Accepted because the plugin genuinely cannot reach the repo's `hooks/` tree; mitigated by testing the real function behaviorally rather than trusting the copy.

## Gates

`pytest bin/tests/` 1606 passed + 25 subtests; `hooks/tests/test-*.js` loop 54 files all pass; `hooks/tests/test-*.sh` loop passes; `test-repo-root.py` 45 passed; em-dash check clean; `check-resident-budget.sh` OK.

Doc-sync: predicate not triggered - no reality-asserting change to README, CONTRIBUTING, or `content/SKILL.md`; adapter hook internals are not documented there.

## Review rigor

**Brief / Plan path:** none on disk; the site inventory is the committed `hooks/tests/fixtures/agentic-write-sites.txt`.

**Skeptic rounds (tier):** see review thread on this PR.

**Findings summary:** this is the fourth PR in a sequence where every hand-built inventory of `.agentic` write sites proved incomplete (4 sites, then 39, then 47, then 57, then these 11). The gate extension is the response to that pattern rather than a fifth list.

### DIFF
--- a/.copilot/hooks/session-start-copilot.sh
+++ b/.copilot/hooks/session-start-copilot.sh
@@ -8,13 +8,26 @@
 # Output: JSON on stdout with hookSpecificOutput.additionalContext (or empty object).
 #
 # Reference: VS Code Copilot Extensibility - hooks reference (SessionStart output format).
+#
+# The .agentic/ read is anchored at the nearest .git ancestor of the payload
+# cwd via hooks/lib/repo-root.sh's resolve_agentic_root, not the payload cwd
+# verbatim, and NEVER falls back to `pwd` - a stray cd or a drifted
+# harness-supplied cwd could otherwise point this hook at a phantom
+# .agentic/ tree wherever the process happens to be. On resolution failure
+# (payload has no cwd, or no .git ancestor is found), the hook SKIPS the
+# lookup entirely and emits {}.
 
 set -euo pipefail
 
-# Resolve workspace root from stdin JSON (cwd field), fall back to pwd
-CWD=""
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# shellcheck source=../../hooks/lib/repo-root.sh
+source "$SCRIPT_DIR/../../hooks/lib/repo-root.sh"
+
+# Resolve workspace root from stdin JSON (cwd field). Never falls back to
+# `pwd` - see the header comment above.
+PAYLOAD_CWD=""
 if command -v python3 >/dev/null 2>&1; then
-  CWD="$(python3 -c "
+  PAYLOAD_CWD="$(python3 -c "
 import json, sys
 try:
     data = json.load(sys.stdin)
@@ -24,8 +37,14 @@ except Exception:
 " 2>/dev/null || true)"
 fi
 
+CWD=""
+if [[ -n "$PAYLOAD_CWD" ]]; then
+  CWD="$(resolve_agentic_root "$PAYLOAD_CWD")"
+fi
+
 if [[ -z "$CWD" ]]; then
-  CWD="$(pwd)"
+  printf '{}\n'
+  exit 0
 fi
 
 CONTEXT_FILE="$CWD/.agentic/context.md"
--- a/.copilot/hooks/stop-context-copilot.js
+++ b/.copilot/hooks/stop-context-copilot.js
@@ -24,6 +24,13 @@
  * reader with a first-byte timeout and a re-armed inactivity timeout) instead
  * of a blocking fs.readFileSync('/dev/stdin'), so this hook cannot hang
  * Copilot's shutdown path when the spawning process never closes stdin.
+ *
+ * The .agentic/ write is anchored at the nearest .git ancestor of the
+ * payload cwd via hooks/lib/repo-root.js's resolveAgenticCwdWithDiagnostics,
+ * not the payload cwd verbatim - a stray cd or a drifted harness-supplied
+ * cwd could otherwise write a phantom .agentic/ tree wherever the process
+ * happens to be. On resolution failure (no .git ancestor found), the write
+ * is SKIPPED entirely rather than falling back to the unresolved cwd.
  */
 
 'use strict';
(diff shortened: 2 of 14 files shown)


tyson-solara6:
Automated review pass (own PR, no formal verdict per self-review policy).

**Skeptic review (adversarial correctness pass, ran the new tests locally + a live mutation check on the Cursor hook to confirm regression assertions actually redden):**
- Verified the `resolveAgenticRoot` duplication in `.opencode/plugins/session-context.ts` is behaviorally equivalent to the canonical `hooks/lib/repo-root.js` (existence-only `.git` check, `MAX_DEPTH`/`AGENTIC_ROOT_MAX_DEPTH` match, linked-worktree `.git`-as-file handled in both).
- Verified the "skip write entirely on no `.git` ancestor" behavior change is safe everywhere it's applied - consistent with the pre-existing pattern in `session-start-wrap.sh`/`enforce-skeptic-round-cap.py`, and is precisely the fix this PR series exists to make.
- Confirmed the new tests are real behavioral coverage, not cosmetic (mutation-tested one to verify it actually reddens on regression).
- One **Major** finding: `hooks/lib/repo-root.js`'s "Downstream consumers" manifest comment was stale - missing the three new JS requirers this PR adds (`.copilot/hooks/stop-context-copilot.js`, `.github/hooks/stop-context-copilot.js`, `.cursor/hooks/stop-context-cursor.js`). This is a mechanical doc-only fix, so I applied it directly in an ephemeral clone (commit `23a80105`, pushed to this branch) rather than leaving it as a dangling comment.

**North Star alignment:** ALIGNED - closes a real phantom-`.agentic/`-tree bug in the remaining adapter hooks and adds a coverage-gate test so the bug class can't silently recur, improving verifiable outcomes without adding operator ceremony.

No out-of-scope findings to file as follow-up tickets this pass.

tyson-solara6:
Superseded by a clean branch. An unsigned commit (23a80105, authored tyhummel@gmail.com / committed admin@fullmetalblanket.com) landed on this branch from a concurrent session and fails the DCO check; force-push is blanket-denied here, so the branch cannot be repaired in place.

Its content - a 3-line manifest addition to hooks/lib/repo-root.js - is a strict subset of the round-2 rework and is preserved in full on the replacement branch, which does the same update more completely. Nothing is lost.

Replacement: fix/adapter-hooks-agentic-root-v2, both commits DCO-signed.

PR #755, opened 2026-08-16
fix(content): ticket creation captures intent, not a plan - gate on first spawn of any kind

tyson-solara6:
## Problem

Ticket creation was structurally forced to over-enrich. The Ticket-offer gate fired before the "first implementer (architect, engineer, or orchestration-planner)" but `investigator` was not in that list, while investigator-before-architect is mandatory for unfamiliar territory. So for net-new work in an unfamiliar area - exactly the work most likely to need a ticket - the required sequence ran the investigator *before* the ticket existed, leaving its terrain map in conductor context at the moment `TICKET_BODY` was composed.

Nothing pushed back. No prose anywhere in `content/` said a ticket body carries intent rather than a plan; the only shape control was a ~15-line soft cap, which limits verbosity, not content type - a compressed file-by-file plan passed it. Meanwhile the Create Helper contract hedged the one obligation that should be hardest: "Problem + Acceptance Criteria **when known**". The rule set was backwards on both halves at once.

Three costs, in increasing severity: the creating session pays investigation for an offshoot unrelated to its in-flight work; the resulting plan is stale by pickup; and a plan in a ticket body anchors the implementing session, converting Phase 3's independent derivation into a grading exercise and narrowing its search to a closed list it never chose and cannot see.

## Change

One governing principle - **recoverability** - with four consequences rather than four rules. Operator intent (Problem, why it matters, Acceptance Criteria) is irrecoverable once the moment passes, so capture it completely. Everything else is recoverable at pickup and strictly better re-derived there.

- **Gate fires on the first subagent spawn of any kind** (`content/sections/02-delegation.md`), replacing an inclusion list with an exemption list. An inclusion list fails open - a new agent file is silently ungated. An exemption list fails safe. Net **-54 B** on the embedded kernel surface.
- **Content rule** in `content/references/conventions-detail.md` §Ticket descriptions: Problem and AC mandatory with a reachable `not yet defined - <blocker>` escape; evidence is intent-bearing and may run long; a proposed approach is excluded regardless of cost.
- **Consuming-side guard**: the architect treats ticket-body specifics as `[per ticket-body, unverified]` and re-derives, grading its plan against Problem and AC only.
- **All three authoring paths** covered - Create Helper direct, `/ds-brief`, `/ds-feedback-triage`. The last was building `TICKET_BODY` from `item.suggested_body`, a proposed fix, which is precisely the excluded content.
- **Mechanical pin**: `bin/tests/test_ticket_offer_gate_exemption_spec.py` asserts bidirectional set equality between the prose classification and the disk-derived agent set, so the lists cannot go stale by hand.
- **7 stale restatements** of the old trigger fixed across `content/references/`, `content/commands/`, `README.md`, and 3 docs pages.

## Verification

All gates re-run independently by the reviewer, not accepted from the implementer.

| Gate | Result |
|---|---|
| `check-resident-budget.sh` | 391 B / 600 |
| `check-skill-embed-budget.sh` | 138,826 B (**-54 B**), floor 100,000 / ceiling 139,160 |
| `check-command-file-budget.sh` | 370,964 B (**-1 B**) / 371,000, headroom 36 B |
| `check-methodology-drift.sh` | OK, 12 files |
| `pytest bin/tests/ -q` | 1642 passed, 25 subtests |
| New spec | 4 passed; mutation proven both directions |
| hooks JS suite | 54 files, 1 pre-existing env failure (see below) |
| `build-all.sh` + adapter-drift pathspec | zero diff, zero untracked |
| `check-symlinks-relative.sh` | exit 0 |

**No threshold was raised.** Both byte budgets were breached during design and closed by relocating payload to uncapped reference files, not by moving a limit.

**Mutation proof, both directions:** adding `content/agents/zzz-probe.md` fails 3 tests; renaming `skeptic.md` fails 3 tests; restoring passes 4. Run with `-p no:cacheprovider`.

**Adapter regeneration audited:** rebuilding from scratch at the head commit produces zero diff against the verbatim `adapter-sync.yml` pathspec, so all 48 regenerated files are derived by construction. `.codex/skill-compatibility.yml` head matches a fresh inventory byte-for-byte; occurrence delta is exactly +1 `conventions-detail.md`, +1 `delegation-detail.md`, zero removals.

**Pre-existing, not from this change:** `hooks/tests/test-stdin-guard.js` fails with `Cannot find module 'espree'`. Reproduces identically on `origin/main` - missing `node_modules`, not a regression.

Doc-sync: triggered -> updated `docs/index.html`, `docs/components.md`, `docs/configuration-reference.md`, `README.md`, plus all 11 adapter dirs via `build-all.sh`. `docs/slides/*.md` grepped by hand for the affected phrasing - zero hits, no deck rebuild needed.

## North Star alignment

**Pillar 1 (guard operator attention)** - served. The mandatory `not yet defined` marker surfaces an honest gap instead of a silently incomplete ticket, and the creating session stops spending an investigation on an offshoot that does not serve the work in hand.

**Pillar 2 (verifiable outcomes autonomously)** - served, and this is the load-bearing one. A ticket carrying a plan collapses the implementing session's independent derivation into agreement with a stale guess, and lets a Skeptic mistake two descendants of one guess for corroboration. Keeping design out of the ticket preserves the independence the review loop depends on.

**Pillar 7 (prevent defects at the producing step)** - served. The gate stops the myopia defect being produced rather than catching it downstream, and the set-equality pin moves classification drift from review-time to CI-time.

**Pillar 5 (guard agent context)** - genuine tension, resolved rather than asserted away. The first design added +798 B to the always-loaded kernel against 280 B of headroom. Rather than raise the ceiling, the content principle moved out of the embedded surface entirely; the kernel now shrinks by 54 B while carrying the one fact that must be resident (a ticket exists before any spawn).

## Review rigor

Four review rounds: three on the plan, one on the implementation diff. Worth recording because the failure pattern was consistent and is itself the subject of the change.

Round 1 found the plan's "full apparatus, derived from `content/agents/`" enumeration accounted for 9 of 18 agents - the fix reproduced the exact closed-list defect it existed to repair. Round 2 found the byte fix had *relocated* rather than repaired: the payload moved off the skill-embed ceiling straight onto `check-command-file-budget`, which had 35 B of headroom, and found 7 untouched restatements the plan's own docs-currency grep provably could not detect (pattern run, exit 1). Round 3 found the Edit 6 anchor text occurs twice in the file, where a both-sites apply would have netted +49 B and broken the budget.

Three separate instances of one class - a hand-maintained closed list nobody re-derived - which is why the shipped change replaces enumeration with an exemption set plus a mechanical bidirectional pin rather than adding another list. Rounds 1 and 2 each recorded one convergence failure where a claimed fix was a reinterpretation of a false premise rather than a repair.

One implementer claim did not survive verification: the `.codex/skill-compatibility.yml` regeneration was reported as unrelated pre-existing drift. Measured, `origin/main` regenerates byte-identically and the delta traces entirely to this change's two new cross-references. The artifact is correct; the stated rationale was not.


tyson-solara6:
Automated review pass (self-authored PR, no formal verdict per review policy - GitHub blocks self-review anyway).

**North Star alignment:** ALIGNED - strengthens verifiable outcomes (pins the ticket-gate to actual operator intent, preserving the Skeptic review loop's independence) and guards operator attention (stops unfamiliar-territory investigation from silently drifting into a de facto plan before intent is captured), with no added attention tax since the gate is mechanical.

**Adversarial check (no material issues found):**
- Exemption list in `content/references/delegation-detail.md` / `content/sections/02-delegation.md` correctly partitions all 18 files under `content/agents/` plus the kernel-only `general-purpose` role - no gaps, no overlap.
- `bin/tests/test_ticket_offer_gate_exemption_spec.py` genuinely asserts bidirectional set equality (`disk_roles - classified` and `classified - disk_roles` both empty), with mutation tests exercising both failure directions independently.
- Spot-checked adapter regeneration (`.copilot/`, `.gemini/`, `.hermes/`) against `content/` source - byte-identical propagation, no drift found in the samples checked.
- The "not yet defined - `<blocker>`" escape hatch is prose-only with no mechanical enforcement, but that's consistent with the rest of the methodology's other prose-only rules, not a regression this PR introduces.

Minor non-blocking nit: "an inclusion list fails open... an exemption list fails safe" appears verbatim in both `content/sections/02-delegation.md`'s propagated wording and `delegation-detail.md` - a little repetitive for a reader who hits both, not worth a fix.

tyson-solara6:
Automated review pass (own PR, no formal verdict per self-review policy).

**Scope of this pass:** the round-3 rework commit (`70fd7b9f`), landed after our prior automated comment which covered round-2 (`32aec972`). The PR body's own "Review rigor" section already documents an extensive independent-verification history across all 3 rounds, so this pass focused on whether round-3's own claims hold up.

**Skeptic re-run:** blocked on missing formal spawn context (architect-plan path, qa_criteria, per-consumer impact table, etc. — the Skeptic protocol requires these for a from-scratch review and this is an ad-hoc automated pass, not a Phase-4 spawn). Not re-attempted; the PR's self-documented 3-round review history (with two recorded convergence failures and one falsified rationale, per the "Review rigor" section) already demonstrates a real adversarial process rather than rubber-stamping, so no gap here.

**North Star alignment: ALIGNED.** Forces ticket bodies to capture only irrecoverable operator intent (Problem/AC, with a mandatory "not yet defined" escape) and strips proposed plans out, which serves Pillar 2 (verifiable outcomes) by preserving the implementing agent's independent derivation instead of anchoring it to a stale guess, and serves Pillar 1 by replacing silent gaps with an honest marker — no new operator-facing step, net -54 B on the always-loaded kernel, so attention/friction pillars aren't regressed.

One nit surfaced during the alignment pass: the PR's own "North Star alignment" section cites "Pillar 5" and "Pillar 7," which don't exist in the current three-pillar `docs/overview/vision.md` — the underlying reasoning still maps cleanly onto real Pillars 1 and 2, so this is a labeling mismatch only, not a substantive alignment problem. Worth a quick fix to the PR body/commit message wording if this lands, but not blocking.

**Mergeability:** `mergeStateStatus: BLOCKED` / `mergeable: MERGEABLE` — all CI checks green. BLOCKED typically means branch-protection review requirements aren't satisfied yet (expected for a self-authored PR that can't self-approve), not a technical conflict.

No out-of-scope pre-existing issues spotted worth a Jira follow-up this pass (and note: Jira MCP tools are not available in this environment this run, so follow-up filing is skipped entirely for this run — see run summary).

PR #759, opened 2026-08-17
fix(bin): retire bin/ds-reap-all now superseded by ds-cleanup-worktrees --multi-repo

tyson-solara6:
## Summary

Retires `bin/ds-reap-all`, `bin/agentic-reap-all`, and `bin/tests/test_ds_reap_all.py` (about 1124 lines), whose capability moved in-process to `bin/ds-cleanup-worktrees --multi-repo` in PR #758. This completes a three-unit sequence: #757 renamed the binary to match the slash command, #758 folded in the cross-repo sweep and added a ranked `--report`, and this unit retires the now-redundant wrapper.

## North Star alignment

- Pillar 2 (Produce verifiable outcomes autonomously): the round-2 gap this PR closes was exactly a claim without verification - a coverage table said "MATCH on level" for the runtime failure path when nothing tested it. The fix is verified two ways before merge: mutation-testing (`if rc != 0` -> `if False:` on `_run_multi_repo:2534-2536` turns the new test RED) and marker-file instrumentation confirming the branch actually fires - the same two techniques the Skeptic used to disprove the round-2 claim.
- Pillar 7 (prevent defects at the producing step): the marker-file reachability audit is a mechanical check (does every branch in `_run_multi_repo`/`_run_report` get hit by the suite) that this round runs on itself before submitting, not just after a Skeptic finding - the same check is now available to catch the next retirement-style PR before it reaches review.
- Pillar 1 (guard operator attention): retiring the wrapper and its now-superseded subprocess-per-repo test suite removes ~1124 lines of code an operator no longer needs to read or maintain, with no coverage loss.

## Review rigor

Three Skeptic rounds on this unit.

- Round 1 found 2 Majors: two tests marked "ported" from the retired `test_ds_reap_all.py` were ported at a LOWER LEVEL than the originals - exit codes, summary lines, and stderr diagnostics stopped being checked while the test names still matched. Both were mutation-proven to ship green with the underlying behavior broken.
- Round 2 found the same shape a third time, hidden behind an "out of scope" carve-out in my own re-audit table: the runtime per-repo failure path (`_run_repo` returning nonzero mid-sweep, aggregated by `_run_multi_repo`'s `if rc != 0: repos_errored += 1`) had zero coverage, proven by both mutation testing and marker-file instrumentation showing no test ever enters that branch.
- Round 3 (this round) closes it: a new test monkeypatches `_worktree_list` (the first real git call inside `_run_repo`, reached only after discovery already succeeded) to raise for one named repo mid-sweep, driving the exact `except (RuntimeError, ValueError): return 1` path the aggregation logic exists to handle. Mutation-confirmed RED against the same `if False:` mutation; marker-file-confirmed the branch now fires. The same marker-file technique was then run against every remaining branch in `_run_multi_repo` and `_run_report` - all reachable, no further gaps found. Also fixed a Minor: the manifest's network-call arithmetic overstated per-entry cost (it said 3 network calls per entry; measured, `gh auth status` runs once per REPO via `_gh_available`, not per entry - the real shape is 2 per entry plus 1 per repo).

Durable process point: when retiring a test file, a mapping table of names is not evidence of coverage. Port each assertion at the LEVEL it was written at, and verify reachability by instrumenting the branch rather than by reading the replacement test. An "out of scope" note on a coverage question is a finding, not a scoping decision.

## Post-merge

After merge, every existing install has a dangling `~/.local/bin/ds-reap-all` symlink pointing at the now-deleted binary. This is repaired generically by `ds-doctor --fix` (verified by execution against a fabricated dangling symlink in a scratch HOME) - no dedicated migration step needed.

Doc-sync: predicate not triggered (no reality-asserting change outside bin/ and its own tests; the docstring/manifest fix is a same-file accuracy correction, not a new claim requiring cross-doc sync).


tyson-solara6:
## Automated review pass (informational - self-authored PR, no formal verdict per self-review policy)

Adversarial (skeptic-style) verification of the PR body's claims, checked against the actual diff/repo rather than taken on faith:

- **Mutation-tested runtime-failure test claim: checks out.** `test_runtime_repo_failure_mid_sweep_continues_and_errors` in `bin/tests/test_cleanup_worktrees_multi_repo.py` monkeypatches `_worktree_list` to raise for one repo mid-sweep, then asserts the sweep continued (`calls == ["good-a", "bad-b", "good-c"]`), `rc == 1`, and `errored=1` - genuinely exercising `_run_repo`'s `except (RuntimeError, ValueError): return 1` and `_run_multi_repo`'s `if rc != 0: repos_errored += 1` aggregation.
- **Full deletion / no dangling references: checks out.** `bin/ds-reap-all`, `bin/agentic-reap-all`, and `bin/tests/test_ds_reap_all.py` are all cleanly deleted. All remaining "reap-all" mentions in the diff are retrospective prose, not live code/path dependencies.
- **`ds-doctor --fix` auto-repairs the dangling `~/.local/bin/ds-reap-all` symlink post-merge: checks out.** Verified `check_local_bin_stale` / `_stale_repo_bin_link_target` in `bin/ds-doctor` at the PR's head SHA - this is generic, pre-existing dangling-symlink repair logic that already covers exactly this case, not something new added by this PR.
- The docstring correction around the old `--timeout` flag's "parity" claim is an honest walk-back (narrower parity, not full parity) rather than a new problem.

No leftover dead references or unfinished migration steps found. North Star alignment: **ALIGNED** - removes ~1124 lines of redundant code (attention-tax reduction), closes a previously-unverified coverage gap via mutation testing (verifiability pillar), and adds no new setup/ceremony.

Nothing blocking spotted. Self-authored, so no approve/request-changes verdict is posted.

PR #760, opened 2026-08-18
fix(bin): close five deferred discovery-guard defects, including a fail-closed regression (DS-177)

tyson-solara6:
## Problem

Five defects in `bin/tests/test_discovery_zero_match_guard_spec.py`, found by the DS-176 implementation review and recorded as accepted debt at that merge. Deferring was the wrong call, and one of the five was not debt at all: it was a fail-closed regression DS-176's own final round introduced.

All five failed in the loud direction, so nothing was unprotected. But two of them **rejected correctly-written guards**, meaning the file enforced a narrower standard than `content/references/code-standards-detail.md` §Discovery-Based Check Discipline documents. A contributor writing a conforming guard would have hit an opaque red and had to reverse-engineer the real rule.

Ticket: DS-177.

## What shipped

The original five:

1. **`_check_form_c` broke on the first offset match** (regression from DS-176). With the phrase in both an assert's condition and its message on one line, only the leftmost hit was tested.
2. **Byte-vs-character offset mismatch.** `msg.col_offset` is a UTF-8 byte offset; `_line_start_offsets` counted characters.
3. **Overlap where the docstring said containment.** The stricter relation now ships.
4. **Hand-typed cardinals removed.** The `40+` worktree claim, then its `35-36` replacement, then two more found by an all-forms re-sweep (word form, not just numeral - the documented count-sync blind spot).
5. **Manifest `Public API:`** completed, then **pinned** - see below.

Plus four found during review:

6. **The `splitlines()` class**, which splits on `\x0c`, `\x0b`, U+2028/2029, U+0085 where the CPython tokenizer and `text.count("\n")` do not. Fixed at all three source-line-indexing sites (`_site_label_python`, `_site_label`'s yaml branch, `_conforms_to_mandated_form`); `:132` parses git stdout and is legitimately out of class.
7. **Three checker neuterings that survived green** - Form A's `>&2` check, Form B's stderr-direction check, and the `only_on_disk` set direction.
8. **A dangling cross-reference** and an **overstated docstring equivalence** (a lone `\r` is also a tokenizer boundary; unreachable in practice because `read_text` applies universal-newline translation).
9. **AC-5 pinned mechanically.** The `Public API:` list had been hand-reconciled twice and regressed once. It is now set-equality-pinned against an AST-derived walk of every `_`-helper any `test_*` function calls, bidirectionally, so over-declaration reddens as well as omission.

## Verification

Every fix ships a regression test proven to fail against the pre-fix implementation by function swap, not by assertion. The two label-site fixes are pinned **independently** - reverting either site reddens only its own test, so neither can revert silently behind the other.

**Gate neither narrowed nor widened.** All 7 live guard sites conform (4 Form A, 2 Form B, 1 Form C) with correct resolved labels; `disk_live == LIVE_GUARD_SITES` with both difference directions empty. The six-mutation real-file battery still reddens, with shasum-verified byte-identical restores.

**Vacuity sweep clean.** Every test carries a named, executed mutation that reddens it, with each mutation diff-confirmed as landed before trusting the result - a reviewer's own battery was silently vacuous once from heredoc quoting.

All three budget gates unmoved (test-only diff). Zero adapter drift against the verbatim `adapter-sync.yml` pathspec. Rebased onto current `main`; all commits DCO-valid with no merge commits.

## North Star alignment

**Pillar 7 (prevent defects at the producing step)** - served. Two defects silently rejected valid work; a contributor would have learned the real rule only by hitting a red gate.

**Pillar 2 (verifiable outcomes autonomously)** - served, and this is the substantive one. Each fix is pinned by a test proven to fail against the exact prior implementation, so the fix's shape is encoded rather than its symptom. AC-5 moved from hand-maintained to mechanically enforced.

**Pillar 1 (guard operator attention)** - served indirectly. These were surfaced as a debt list rather than fixed; that is the cost this closes.

**Pillar 5 (guard agent context)** - neutral. Test-only, no budgeted surface touched.

## Review rigor

Five implementation review rounds, on top of DS-176's seven passes on the same file. Every round found something, and the pattern is worth recording because it is the ticket's own subject:

- Round 1: the `splitlines()` class, fixed at two sites and declined at a third on "not part of the named fix" - the declined site demonstrably misclassified a conforming guard.
- Round 2: all three sites fixed, but only one pinned. Reverting either label site left the whole suite green.
- Round 3: both sites pinned - and the new test called a symbol the manifest did not list, reverting AC-5 which had passed one commit earlier.

Three consecutive rounds of the same shape: a named class repaired at the sites someone pointed at, while the sibling, the pin, or the manifest went unexamined. That is why AC-5 ended up mechanically pinned rather than hand-corrected a third time - a derived bidirectional assertion is the only form that has held anywhere in this repo.

AC-4's original wording ("no hand-typed cardinal remains anywhere") proved unachievable and was formally amended on the ticket to the rule actually enforced: no such cardinal unless it is pinned by a test that reddens on drift, or explicitly snapshot-framed with a re-derive-at-use-time disclaimer.

Two Minors carried deliberately: two `currently 2` echoes that quote the pinned prose rather than asserting an independent count, and nothing else.

## Final rounds

Two hardening rounds after sign-off, both closing findings in the newly-added manifest pin rather than in the original defects.

The pin initially sliced the whole `Public API:` docstring field, so a `_name` appearing in the field's rationale prose counted as declared - demonstrated blind by deleting a name from the list while mentioning it in prose. It now parses only the enumerated, comma/and-joined run of backtick names, and asserts loudly rather than degrading silently if that shape is absent. Separately, `ast.walk` was including nested function defs where the docstring claimed module-level, so a purely local `_helper` inside a test would have falsely reddened the pin and pressured the author into declaring a non-API name; that is now `tree.body`.

Both fixes are pinned by permanent regression tests proven RED against the pre-fix implementation, not just verified once by hand. That mattered: the unpinned-fix pattern - change lands, pin does not, next round finds it - produced a Major earlier in this same PR.

**Declined, deliberately:** `_PUBLIC_API_LIST_RE` does not accept a no-Oxford-comma two-name list, and a line-wrap split in a first or middle name reddens as "missing from manifest" rather than the locate assertion. Both fail safe in every probed position - they redden, never pass silently - so the only cost is a less precise error message on an edit nobody has made. The reviewer and the implementer independently reached the same conclusion.

Final: 1709 tests pass, 28 in this file. All six pin behaviours verified live with byte-identical restores: name deleted, phantom name, undeclared new helper, helper renamed in def and call sites, list-scope-vs-field-scope, and nested-local-does-not-trip. Seven live guard sites still conform, 4 Form A / 2 Form B / 1 Form C, with set equality empty in both directions.

### DIFF
--- a/bin/tests/test_discovery_zero_match_guard_spec.py
+++ b/bin/tests/test_discovery_zero_match_guard_spec.py
@@ -19,16 +19,17 @@ Public API: pytest test functions only. `_normalized_scan`, `_site_label`,
          hand-typed - DS-177 Fix 5).
 Upstream deps: `git ls-files` (repo-relative, tracked-only discovery -
          this checkout's `.claude/worktrees/agent-*` count fluctuates
-         session-to-session as isolation worktrees are created and reaped
-         (measured 35-36 across two checks in one review), each a full
-         copy of every guard site; an unfiltered `rglob()` from the repo
-         root would
+         session-to-session as isolation worktrees are created and reaped,
+         each a full copy of every guard site; an unfiltered `rglob()`
+         from the repo root would
          multiply every LIVE_GUARD_SITES entry by (1 + worktree count) and
          either mask the problem under bare-basename keys or redden this
-         gate locally while CI, a clean checkout, stays green - see
-         `_tracked_relative_paths()` in
-         bin/tests/test_ticket_offer_gate_trigger_wording_spec.py:110-133
-         for the identical precedent this reuses). `LIVE_GUARD_SITES` keys
+         gate locally while CI, a clean checkout, stays green - see the
+         `_tracked_relative_paths()` function in
+         bin/tests/test_ticket_offer_gate_trigger_wording_spec.py (grep
+         for that function name - a raw line-number citation here has
+         already drifted twice) for the identical precedent this reuses).
+         `LIVE_GUARD_SITES` keys
          are `(repo_relative_path, derived_label)` pairs; the label is
          content-derived (AST scope resolution for .py, nearest enclosing
          `jobs:` job name for .yml), never hand-typed, so a 3-guards-to-2
@@ -86,10 +87,11 @@ GUARD_PHRASE = "discovery is broken, not clean"
 
 SEARCH_SUFFIXES = {".py", ".yml", ".yaml"}
 
-# Content-derived, never hand-typed (DS-176 Round-2 Critical fix). The 6
-# pre-existing entries plus this file's own vacuous-pass guard (7th entry,
-# added because this file's own discovery guard carries the same literal
-# phrase and is therefore itself discoverable by the mechanism it tests).
+# Content-derived, never hand-typed (DS-176 Round-2 Critical fix). All but
+# the last entry below pre-date this file; the last entry is this file's
+# own vacuous-pass guard, added because this file's own discovery guard
+# carries the same literal phrase and is therefore itself discoverable by
+# the mechanism it tests.
 LIVE_GUARD_SITES = frozenset(
     {
         (".github/workflows/bin-tests.yml", "hooks-python-tests"),
@@ -112,8 +114,8 @@ def _tracked_relative_paths() -> list[pathlib.Path]:
     unfiltered directory walk) matters because this checkout carries a
     fluctuating number of `.claude/worktrees/agent-*` copies of the whole
     tree - created and reaped session-to-session as isolation worktrees
-    spin up and down (measured 35-36 across two checks in one review; see
-    the module docstring's Upstream deps note) - none of which are part of
+    spin up and down (fluctuating count, see the module docstring's
+    Upstream deps note) - none of which are part of
     the primary worktree's git index. DS-177 Fix 4: no cardinal is
     asserted here, since one would go stale independently of this
     function's own behavior."""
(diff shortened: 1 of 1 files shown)


tyson-solara6:
Automated review pass (own-PR, no formal verdict since self-review is blocked). No trivial/mechanical fixes found worth auto-applying - every change in `bin/tests/test_discovery_zero_match_guard_spec.py` is substantive (doc/comment rewording tied to a specific DS-177 fix, or an actual logic change), nothing is formatting-only.

Non-blocking concerns for your own follow-up, not auto-fixed:
- `test_bidirectional_comparison_reddens_on_unpinned_disk_site` (around line 233-263) directly calls `test_live_guard_sites_bidirectional_set_equality()` as a plain function from inside another test. That's a naming/refactor coupling - renaming or restructuring the called function would silently break this test's monkeypatch target and control flow. Consider extracting the shared assertion logic into a helper instead.
- Line ~115 docstring: "DS-177 Fix 4: no cardinal is asserted here" - worth a self-confirmation that removing the magic-number assertion doesn't weaken a regression signal elsewhere.

The byte/char offset conversion and overlap/containment logic look correct given the documented AST byte-offset behavior.

No out-of-scope pre-existing issues found in this diff, so no Jira follow-up tickets filed for this PR.

tyson-solara6:
**Automated review pass (informational only - self-authored PR, no formal verdict per review policy)**

Ran an adversarial correctness read (skeptic) plus a North Star alignment check. Diff touches one file: `bin/tests/test_discovery_zero_match_guard_spec.py`.

**Correctness:** Critical: 0, Major: 0. Verified by actual execution, not just reading:
- Each of the six claimed fixes (first-offset-only break in `_check_form_c`, byte-vs-char offset via `_byte_col_to_char_col`, overlap->containment relation, the four `splitlines()`-class fixes, and the AC-5 mechanical pin restricted to `tree.body`) was confirmed red-before/green-after by hand-reverting each site independently and re-running the suite - none masks another, each has its own dedicated regression test.
- The guard's own mutation/vacuous-pass self-checks (weakened-to-log-only, guard-entirely-removed, stdout-only, straddling-boundary) still pass post-fix - no silent-pass ("fail-open") regression introduced by the fixes.
- Full non-repo-dependent run: 25/25 tests passed.

**Minor (2), non-blocking, no trivial fix needed:**
1. The self-referential `DOCSTRING_REFERENCE_COUNT`/"currently 2" docstring pin - already flagged as accepted debt in the PR body.
2. `_PUBLIC_API_LIST_RE` anchors on the literal `"are\b"` following the name list; if that docstring phrasing is ever reworded, `_manifest_public_api_names` raises rather than degrading gracefully. Acceptable (fails loud, not silent) but worth knowing as a coupling point.

No trivial/mechanical issues found worth auto-applying.

**North Star alignment:** ALIGNED - advances Pillar 2 ("Produce verifiable outcomes autonomously") by mechanically pinning previously hand-maintained assertions and encoding each bugfix's shape with a proven red-before/green-after test rather than an unverified assertion. CI: two checks still pending (`hooks-js-tests`, `python-bin-tests`, `bin-sh-tests`, `check-codex-skill-sync`) at time of review; nothing failing.

No out-of-scope pre-existing issues spotted worth filing a Jira ticket for.

PR #763, opened 2026-08-18
DS-179: make methodology obsolescence-pruning an actual process

tyson-solara6:
## Summary

DS-179: wires the floor-vs-dial obsolescence judgment into `/ds-prune-harness` as a mandatory pre-filter, replaces hand-typed "does this hook ever fire" reasoning with a signal derived off `.agentic/.enforcement-fires.jsonl`, and moves disposition tracking out of gitignored `docs/planning/` into a tracked ledger + audit-state file whose lifecycle is closed by the same PRs that action candidates.

- `content/references/obsolescence-signal.md` - the floor-vs-dial test (sourced verbatim from root `MEMORY.md`'s 2026-07-02 decision) plus two worked examples from this repo. Trigger-loaded (Goal 4 partition test), not resident.
- `scripts/lib/enforcer_facts.py` - the existing AST-derivation logic extracted verbatim from `bin/tests/test_tracker_writeback_ranking_spec.py`'s round-5 `_derive_enforcer_facts()`; that test now imports it (mirroring `test_stamp_agent_fragments.py`'s import pattern) instead of defining it locally.
- `bin/ds-hook-fire-report` (+ `bin/agentic-hook-fire-report` compat symlink) - joins per-hook posture (ACTION_ONLY vs EVERY_VERDICT) against fire-log counts over a window, classifying each hook into `UNMEASURED` / `ZERO_INVOCATIONS` / `ZERO_ACTION_IN_WINDOW` / `ACTIVE`. The analyst Worker has no Bash in its `tool_scope`, so the conductor runs this and passes the JSON into the spawn prompt (new Signal 8).
- `docs/pruning-ledger.md` + `docs/pruning-audit-state.json` - tracked, path-resolved per repo (prefer a tracked `.agentic/pruning-ledger.md` when the consuming repo tracks `.agentic/`, else `docs/pruning-ledger.md`; determined mechanically via `git check-ignore -q`, never assumed). Resolves to `docs/` in this repo, as expected. Fed by exactly two commands (`/ds-prune-harness`, `/ds-representation-audit`) - scoped out `/ds-evaluate` and `/ds-failure-audit` explicitly, per conductor resolution, since those produce whole-report scorecards, not standing per-item candidates.
- `content/commands/ds-prune-harness.md`, `ds-representation-audit.md` - new Step 0.5 (ledger path resolution, carry-over read, audit-state model/commit-count advisory nudge - prune-harness only), Step 2.5 (record ledger entries via a delegated Trivial engineer spawn), and a same-PR ledger-closure requirement wired into Step 4's `/ds-update-agentic-engineering` spawn brief - the Skeptic reviewing the deletion/rewrite diff necessarily reviews the ledger update in the same pass, closing the write-only-graveyard risk.

## North Star alignment

**Advances:**
- **Pillar 4 (universality)** - the ledger/audit-state path resolution is explicitly per-repo, never hardcoded to `docs/`; DinoStack's own resolution to `docs/` is stated as the expected local outcome, not baked into the resolution logic itself.
- **Pillar 2 (verifiable outcomes)** - disposition closure (`RAISED` -> `ACTIONED`) can only happen inside a PR diff a Skeptic already gated, closing a write-only-graveyard risk the prior `docs/planning/`-only design had no mechanism against.
- **Pillar 5 (guard agent context)** - the floor-vs-dial test and its worked examples live in a trigger-loaded reference doc, reached only from the prune-harness command, not the resident skill embed.

**Cuts against, accepted as a deliberate trade-off:**
- **Pillar 3 (low friction)** - this adds process weight to `/ds-prune-harness` and `/ds-representation-audit` (Step 0.5, Step 2.5, a new tracked ledger file, a delegated engineer spawn per run). Accepted because both commands are quarterly/on-upgrade utilities, not part of the daily loop, and the added ceremony is exactly what closes the audit-trail gap DS-179 was filed to fix - the prior state was zero enforcement of the floor-vs-dial judgment and a purely ephemeral proposal trail.

## Known limitations (stated per plan, not hidden)

- The AC4 model-change nudge's primary signal (self-reported model string) is inherently unreliable by this repo's own precedent (`never trust an agent-reported SHA`); the commit-count signal (`git log --since=<last_run.date> -- content/... hooks`) is the load-bearing half by design, and the command prose labels it as such.
- The ledger's disposition-closure only closes the loop for candidates that get actioned or rejected. A candidate neither approved nor explicitly rejected stays `RAISED` and depends on the next run's carry-over. If `/ds-prune-harness` is never run again, a `RAISED` entry can go stale indefinitely - this is a real residual gap inherent to "the process only runs when invoked," and this ticket does not build a forcing function for it (a periodic-cadence `enforce-*.py` hook was considered and rejected, see below).

## Rejected alternative (named, not built)

A `SessionStart`/`Stop` `enforce-*.py` hook firing the model/commit-change nudge automatically every session. Rejected because (a) it cannot know the current model version any more reliably than the conductor's self-report, so it buys no rigor; (b) to be useful it would gate on inferred session capability, which `hooks/AGENTS.md` flatly prohibits; (c) it adds a 13th hook plus `MANAGED_HOOK_BASENAMES`/subcount-site updates for a nudge that fires at most quarterly.

## Deferred (stated in plan, reversible)

- Ledger schema-validator test - deferred until the ledger has real entries (a header-only-file assertion would be vacuous).
- `bin/ds-hook-fire-report --days` default of 14 - revisit once real audit-cadence data exists.

## Doc-sync

`Doc-sync: predicate triggered (content/commands, content/references, and bin/ changes) -> updated docs/index.html (reference-doc-count bump + fixed two stale command descriptions for /ds-prune-harness and /ds-representation-audit that predated this PR), docs/slides/contributing-slides.md AND rebuilt .html (via `bash scripts/build-slides.sh` in round 2 - round 1 bumped the .md source but never ran the rebuild, leaving the .html stale at "39 reference docs"; `check-slides-sync` is now green), README.md, CONTRIBUTING.md, ADAPTERS.md, docs/components.md (all reference-doc count 39->40), and bin/AGENTS.md (added the `ds-hook-fire-report` row to the curated tool table, round 2).`

## Verification

- `python3 -m pytest bin/tests/ -q` - 1702 passed, 25 subtests passed (full suite, watched to completion).
- 54/54 `hooks/tests/test-*.js` files (excluding the 2 quarantined into `wrap-lock-tests`) - watched to completion, zero FAIL lines.
- `bash scripts/build-all.sh` - all 11 adapters built; `.claude/skills/dinostack/{agents,commands,references,rules}` symlinks re-checked relative (no absolutization) before staging.
- `bash scripts/check-resident-budget.sh`, `check-skill-embed-budget.sh`, `check-command-file-budget.sh` - all OK.
- Each new test in `bin/tests/test_ds_hook_fire_report.py` was proven to redden against a real code mutation (not just a paired before/after test) - see the engineer return summary for the exact mutation executed and reverted.

## Round 2 (Skeptic rework - 5 Major, 6 Minor, all resolved)

- **M1 (slides drift)** - `docs/slides/contributing-slides.md` was bumped 39->40 in round 1 but the `.html` was never rebuilt (`check-slides-sync` was RED). Ran `bash scripts/build-slides.sh`; only `contributing-slides.html` changed, now correctly reads "40 reference docs" in both spots. Doc-sync line above corrected to stop claiming a rebuild that didn't happen.
- **M2 (`ds-hook-fire-report` loaded its dependency from `--repo`, not its own install dir)** - `_load_enforcer_facts()` now resolves `scripts/lib/enforcer_facts.py` against `_REPO_ROOT_DEFAULT` (the tool's own resolved install dir), never against the `--repo` target. Module manifest updated to describe this. New regression test `test_repo_without_its_own_enforcer_facts_still_succeeds` (a `--repo` target with `hooks/` but no `scripts/lib/`) - confirmed it fails with an uncaught `FileNotFoundError` against the pre-fix code, passes against the fix.
- **M3 (Signal 8 structurally inert)** - two changes: (a) Step 1's conductor invocation now passes `--days 90` so a quiet hook can actually reach the confidence tier; (b) confidence now keys on a new MEASURED field `meta.log_coverage_days` (derived from the log's own earliest parseable timestamp), never on the requested `--days` value (`meta.requested_window_days`). New tests prove the two are distinct: `test_log_coverage_days_is_measured_not_the_requested_window` (a 3-day-old log + `--days 90` measures ~3 days of coverage, not 90) and `test_log_coverage_days_is_none_when_log_absent`.
- **M4 (present-but-unparseable log read as a real zero)** - `build_report` now tracks `log_parsed_lines` distinct from `log_total_lines`/`log_malformed_lines`; when a present log has zero parsed lines (empty or wholly garbage), every hook now reports `UNMEASURED` via a new `log_effectively_empty` meta flag, never `ZERO_INVOCATIONS`/`ZERO_ACTION_IN_WINDOW`. A partially-malformed log (some lines parse, some don't) is treated as a real measurement over the lines that did parse, with `log_malformed_lines` surfaced so the caller can see it wasn't clean. New tests: `test_wholly_malformed_log_reports_unmeasured_not_zero`, `test_empty_present_log_reports_unmeasured`, `test_partially_malformed_log_is_a_real_measurement` - each proven to redden against the pre-fix `if not log_present:` gate.
- **M5 (ledger carry-over contradicted its own command prose, and half was a graveyard)** - `/ds-representation-audit` now has its own Step 0.5 carry-over read (it had none before). Both commands' carry-over reads are now scoped by `Source` field (`ds-prune-harness` vs `ds-representation-audit`), not just `File(s)` overlap, so a rewrite candidate can never be suppressed as an "already-known" pruning candidate or vice versa. The Relationship section in `ds-representation-audit.md` ("this command only writes to it") is corrected to state it now also reads its own carry-over.
- **Minor: no legend in report output** - added `meta.legend` (JSON mode) and a printed legend block before the table (table mode); `test_legend_present_in_json_and_table` covers both.
- **Minor: `docs/pruning-ledger.md`'s `.agentic/` gitignore citation was wrong** - it cited a hooks-gating decision that has nothing to do with `.agentic/` trackability. Now cites the actual decision ("DinoStack does not commit its own `.agentic/` runtime files").
- **Minor: `git check-ignore -q` prose treated exit 128 as trackable** - `ds-prune-harness.md` Step 0.5 and the two other restatements (`docs/pruning-ledger.md`, `ds-representation-audit.md`) now distinguish exit 1 (trackable) / exit 0 (ignored) / exit 128 (git failure, neither - falls back to `docs/` with a warning).
- **Minor: records with absent/unparseable `ts` silently inflated all-time vs window with no explanation** - new per-hook `fire_count_unparsed_ts` field makes this visible; `test_unparsed_ts_record_visible_via_unparsed_ts_count` covers it.
- **Minor: unused `import os`** in `bin/tests/test_ds_hook_fire_report.py` - removed.
- **Minor: `bin/ds-hook-fire-report` absent from `bin/AGENTS.md`'s tool table** - added (the table is a curated subset per its own header, but this tool is now a documented `/ds-prune-harness` dependency analogous to already-listed peers like `ds-doctor`).

All new/changed tests re-run to completion: `python3 -m pytest bin/tests/test_ds_hook_fire_report.py -q` - 16 passed. Each new assertion's named mutation was applied and confirmed to redden the test, then reverted and confirmed green again (see engineer return summary for the exact mutation diffs).

`bin/ds-hook-fire-report`'s `--json` output shape changed from a bare array to `{"meta": {...}, "hooks": [...]}` in round 2 (required to carry `log_coverage_days`, `log_effectively_empty`, and `legend`) - `content/commands/ds-prune-harness.md` Signal 8 and Step 1 updated to match; no other consumer exists yet (Signal 8 is the sole downstream reader, per the module manifest).


tyson-solara6:
## Automated review pass (informational only — self-authored PR, no formal verdict per self-review policy)

Ran a parallel adversarial pass (skeptic, security-auditor, qa-engineer) plus a North Star alignment check against the diff.

### Skeptic findings
- **Major** — `content/commands/ds-prune-harness.md` Step 1 tells the conductor to run `bin/ds-hook-fire-report --json` with no `--days` flag, so it silently uses the script's 14-day default. But Signal 8's confidence rule requires `window_days >= 30` for MEDIUM confidence. As literally documented, every real invocation only ever produces LOW confidence — the MEDIUM branch is unreachable unless someone deviates from the written instructions. This isn't called out in "Known limitations" (which only flags the *default value* will be revisited, not that the doc's own tiering is currently dead).
- **Major** — `build_report()` sets `"window_days"` to the requested CLI arg, not derived from actual log coverage (`earliest_ts`/`latest_ts`). If the fire-log only has 3 days of real data but someone runs `--days 30`, the report claims `window_days: 30` while real coverage is 3 days — Signal 8 could be scored MEDIUM confidence off a near-empty window. No test asserts window_days against measured log age.
- **Minor** — `docs/pruning-ledger.md`'s `RAISED` status conflates "approved, PR pending" and "explicitly deferred" into one indistinguishable state; carry-over logic treats both the same, losing information a later reader might want.
- **Minor** — Ledger entry IDs (`## PL-YYYYMMDD-N`) are assigned per-branch by isolated engineer spawns for `/ds-prune-harness` and `/ds-representation-audit`. If both run the same day (the doc's own "Relationship" section suggests running them back-to-back), each could independently pick the same `N` since neither sees the other's branch — collision/merge-conflict path isn't described anywhere.
- Verified clean: the `enforcer_facts.py` extraction from `test_tracker_writeback_ranking_spec.py` is verbatim (logic, docstrings, `_FIRE_LOG_DECISIONS` unchanged), and ledger/audit-state file shapes match the Step 0.5/2.5 prose.

### Security audit
No Critical/High/Medium findings. No command injection, unsafe deserialization, path traversal, or symlink escape in `bin/ds-hook-fire-report` or `scripts/lib/enforcer_facts.py` (the latter uses `ast.parse`/`ast.walk` only, deliberately avoiding exec of hook code — good). One Info-level note: `ds-hook-fire-report` dynamically loads `scripts/lib/enforcer_facts.py` via `importlib` based on the `--repo` arg — inherent to the tool's purpose and only reachable via local CLI invocation, not a real risk, but worth a comment noting the trust boundary.

### QA / test-coverage gaps
`test_ds_hook_fire_report.py` covers the core status-enum contract well, but gaps remain:
1. No test that an `EVERY_VERDICT` hook also flips to `UNMEASURED` (not `ZERO_INVOCATIONS`) when the fires log is entirely absent (only `ACTION_ONLY` is tested for that case).
2. No test with a mixed valid+malformed-JSONL-line log to prove one bad line doesn't drop/corrupt the rest of the parse.
3. 0-byte fires-log file path isn't exercised (only "file exists, no rows for this hook" is).
4. Days-window boundary test only covers `EVERY_VERDICT`; no equivalent for `ACTION_ONLY` (expected `ZERO_ACTION_IN_WINDOW` vs `ACTIVE`).
5. `main()`'s exit-1 branch (missing `hooks/` dir under `--repo`) is never invoked in tests.
6. `earliest_ts`/`latest_ts` fields are computed but never asserted anywhere.
7. `.py`-suffixed hook names in the log (vs. bare names) aren't exercised by any test.
8. The ledger/audit-state path-resolution branch (`git check-ignore` → prefer `.agentic/` vs `docs/`) and the audit-state bootstrap case (`last_run: null`) are prose-only conductor instructions with zero automated coverage — flagged since the PR's own verification section otherwise reads as "fully tested."

### North Star alignment
**ALIGNED** — converts hand-typed "does this hook fire" guesswork into a data-derived signal and gates ledger disposition-closure behind a Skeptic-reviewed PR diff, strengthening Pillar 2 (verifiable outcomes); the added friction is confined to two quarterly commands, not the daily loop.

### CI status
- `check-slides-sync` is currently **failing** — `docs/slides/contributing-slides.html` is out of sync with the reference-doc-count bump (39→40) baked into this diff. Mechanical fix: `bash scripts/build-slides.sh` and commit the regenerated HTML. (Attempted to apply this directly per the "minor fixes on our own PRs" flow in an ephemeral clone, but repeated filesystem/git operations were denied by this run's sandbox permissions, so leaving it as a comment instead rather than risking a half-applied fix.)
- `mergeStateStatus` is `BEHIND` (branch behind `main`) though currently `MERGEABLE` — worth rebasing before merge.

No out-of-scope/pre-existing issues were spotted (all findings above are in code this PR itself introduces), so no Jira follow-up tickets were filed this run.

tyson-solara6:
## Automated review pass (informational only - self-authored PR, no formal verdict per self-review policy)

New commits since the last automated pass: `47a29f5f`, `8b654b6` (round 2 - resolves 5 Major + 6 Minor Skeptic findings). Ran a parallel adversarial pass (skeptic, security-auditor) plus a North Star alignment check against this delta.

### Skeptic findings
No Critical or Major issues. The core round-2 fixes hold up:
- `_load_enforcer_facts()` now correctly loads from the tool's own install dir, never `--repo` - verified `test_repo_without_its_own_enforcer_facts_still_succeeds` genuinely reddens against the pre-fix code.
- `log_effectively_empty` correctly forces UNMEASURED globally without conflating with a per-hook real zero; partially-malformed logs correctly fall through to normal status logic.
- `log_coverage_days` is correctly decoupled from `requested_window_days` (measured off `min(all_parsed_timestamps)`, returns `None` when nothing parsed).
- `fire_count_unparsed_ts` correctly isolates records with a bad `ts` from whole-line JSON parse failures - no double-counting.
- Minor, non-blocking: `log_coverage_days`'s scan skips any record lacking a `hook` key before collecting timestamps, so it's technically "earliest timestamp among records with a recognized hook field," slightly narrower than the docstring's "the log's own earliest parseable timestamp." Not worth blocking on; a one-line docstring caveat would close the gap if it ever matters.

### Security-auditor findings
No Critical/High/Medium findings. Confirms the `_load_enforcer_facts()` fix is a *real* security improvement, not just cosmetic: previously a caller-controlled `--repo` could have planted a malicious `scripts/lib/enforcer_facts.py` there and gotten arbitrary code executed via `exec_module` (CWE-829); now `--repo` only feeds path-joined reads, never code loading. No exec/subprocess/network/writes anywhere in the new code. A couple of informational-only notes (unbounded `ast.parse`/`read_text` if `--repo` or the fires log were ever attacker-writable) - not applicable under this tool's actual trust model (conductor-supplied `--repo`, own-repo log).

### North Star alignment
**ALIGNED** - improves verifiability (fixes false-confidence bugs in hook-fire measurement) and prevents silently-forgotten pruning candidates via the tracked ledger, for a proportional and accepted attention-tax increase on two already-infrequent commands.

### CI note
`python-bin-tests` shows as `fail` on the current head SHA, but the actual job conclusion is **cancelled**, not a real test failure - the `install zsh` step hung for ~20 minutes before the run was cancelled, and the actual `pytest` step never ran (skipped). Looks like a runner/infra flake (likely the `apt-get install zsh` step stalling), not a code regression from this diff. Worth a manual re-run before merge; if it hangs again on the same step, that'd point to something evironmental rather than this PR's changes.

No trivial/mechanical nits found worth auto-applying.

PR #774, opened 2026-08-20
DS-185: trigger-load the Kimi methodology instead of embedding it in AGENTS.md

tyson-solara6:
**DRAFT - not ready to merge. The round-4 reshape has never been reviewed.**

Moves the methodology body out of `.kimi/AGENTS.md`, loaded unconditionally in every Kimi session, into `.kimi/skills/dinostack/SKILL.md` as a trigger-loaded Agent Skill.

**Measured:** `.kimi/AGENTS.md` 133,474 B -> 1,462 B; `SKILL.md` 139,277 B.

Worth noting about the prior state: `.kimi/skills/dinostack/SKILL.md` existed on `main` only as a symlink to the lean `content/SKILL.md`, carrying no frontmatter, so the skill could not description-match at all. `.kimi/hooks/session-start.sh` already told the model to read `~/.kimi/skills/dinostack/SKILL.md` - a path that resolved to nothing useful. This branch makes that instruction true. That hook is byte-identical to `origin/main` and was deliberately not touched; it is the only mechanism reaching consumer projects, where `.kimi/AGENTS.md` does not exist.

## Why this is a draft

Rounds 1-3 were reviewed. Round 4 was not, and round 4 is the substantive one.

The same defect relocated at constant severity across three rounds, all in `install.sh`'s decision about when to fall back to embedding the body in the resident file:

- Round 1: a hand-picked 20,000 B floor accepted a body-free 40,495 B skill.
- Round 2: the floor derived from the gate, but any nonzero gate exit re-embedded ~132 KB into the tracked, always-resident `AGENTS.md` on a healthy body.
- Round 3: classification switched to matching the gate's diagnostic text, which declared any unrecognised failure "complete and healthy" - so removing both `content/rules/*.md` files silently dropped 33 KB, and a crashed gate produced a false "the skill body itself is complete and healthy".

Round 4 changed the form rather than writing a fourth pattern-match: a fail-safe allowlist where only exact matches on `SKILL.md ABOVE CEILING.`, `AGENTS.md ABOVE STUB CEILING.` and `is likely intentional.` warn without appending, and everything else - missing gate, crashed gate, dropped source file, unrecognised text - degrades by construction.

That inversion is the right shape, and it is exactly the kind of change that should not merge unreviewed.

## Verification claimed by the engineer, not independently confirmed

All three fail-open reproductions run pre- and post-fix through real `HOME=<scratch> bash .kimi/install.sh`; a new `bin/tests/test_kimi_install_link_health_classification.sh` with 5 real-installer scenarios that scores 2/7 against round-3 code and 7/7 against the fix; exact-string pins added for all three diagnostic markers, each mutation-reddened; `build-all.sh` plus adapter drift clean; `bin/tests` 1747 passed and 57 shell suites green; `check-symlinks-relative.sh` clean.

## Also in the branch, reviewed in rounds 1-3

The stub and skill build, a link-health floor derived from the gate rather than hand-picked, `hooks/pre-commit` staging both new generated artifacts (without which the required `check-adapter-sync` would go red on every future `content/**` PR), `ADAPTERS.md:70`, the `DiscoverySpec` in `bin/tests/test_frontmatter_validity.py`, `scripts/check-kimi-skill-embed-budget.sh` with its own suite, a `scripts/lib/kimi-rules.sh` extraction shared by `build.sh` and `install.sh`, and deletion of a false "mirrors `.claude/install.sh`" provenance claim rather than a narrowed rewrite of it.

## Accepted debt

`content/SKILL.md:12-13`'s "`.codex` and `.kimi` embed this file (including this comment) largely unchanged" is false for Kimi. Deliberately left: DS-186 (#772) deletes that per-adapter enumeration outright, which resolves it for every harness at once. Editing it here would conflict for no gain.

## Not done, and not doable by any agent

Two operator-only questions: whether the injection arrives intact in a live Kimi session, and whether `.kimi/hooks/session-start.sh`'s stdout actually reaches Kimi's agent context. The latter is flagged as unverified in the hook itself and in the gate header. Nothing in this branch claims either settled.

## Review rigor

Skeptic rounds (tier): 3 on the diff. Round 1 found 6 Majors, round 2 found 1, round 3 found 5 - severity did not fall, which is what triggered the round-4 reshape rather than another patch. Conductor-side contributions to the cost, recorded in `.agentic/learnings.md` KNW-20260819-004: the round-1 brief named a specific suspected defect instead of a threat category and the reviewer flagged the steer, and the `qa_criteria` block was misstated across all four sibling briefs. Round 3 also accepted an engineer's claim that the harness blocks `HOME` override; the round-3 reviewer refuted it by running the installer nine times and by pointing at `bin/tests/test_kimi_install_symlink.sh:115`, an existing CI-wired test that does exactly that.

Doc-sync: triggered -> `ADAPTERS.md`, `.kimi/README.md`, `.kimi/install.sh` summary. Unverified for round 4.

## North Star alignment

Advances Pillar 5 (guard agent context): removes ~132 KB from the always-resident set of every Kimi session, and improves consumer-project reach rather than regressing it, since the path `session-start.sh` already advertised now resolves to a real body. Named trade-off against Pillar 2 (verifiable outcomes): the body reaches the model only on trigger, with `install.sh`'s degrade path as the compensating control - which is the code round 4 reshaped and which therefore needs review before merge.


tyson-solara6:
Superseded by the v6 branch, which carries rounds 5 and 6 (the fail-safe classification reshape and the test-hygiene fix). Reopening as a fresh PR against the current tip.

PR #776, opened 2026-08-20
DS-183: trigger-load the Codex methodology instead of embedding it in AGENTS.md

tyson-solara6:
Stops Codex double-shipping the methodology. `.codex/AGENTS.md` is symlinked to `~/.codex/AGENTS.md` and loaded unconditionally in every session, while the body was already generated into `.codex/skills/dinostack/METHODOLOGY.md` as a native Agent Skill. This leaves a stub and relies on the skill.

**Measured:** `.codex/AGENTS.md` 140,910 B -> ~6,000 B. `METHODOLOGY.md` unchanged at 102,902 B.

## Also in this PR

A degrade path for when the skill is unreachable, a reachability-based link-health gate replacing an install-provenance (`-L`) test, restoration of Codex-specific `--profile-dir` identity guidance with a `assert_literal_rules_reachable()` guard that now actually fails on its own mutation, `budget_check_embedded_set` extracted into `scripts/lib/budget-gate.sh`, Codex un-gated from `hooks/skill-auto-load-check.sh`, a new `scripts/check-codex-skill-budget.sh` with its own suite registered in `bin/ds-evaluate`, and docs-currency across `ADAPTERS.md`, `CONTRIBUTING.md`, `.codex/README.md` and `AGENTS.md`.

## Notable history

Round 1 shipped a **Critical**: the degrade path wrote `~/.codex/AGENTS.md` as a regular file, which `bin/ds-codex-dispatch:203` and the stub's own runtime-binding preamble both reject, so the safety net produced a session that refuses the methodology it just embedded. Fixed by writing to a companion file and symlinking at it.

That companion was first placed in the gitignored checkout, where `git clean -xfd` left the symlink dangling and no methodology loaded at all. Moving it to `$CODEX_CONFIG_DIR` fixed that but made it user-owned, so uninstall was destroying pre-existing user files. Now marker-based, matching the sibling DS-184 unit: `<!-- dinostack:codex-degrade-generated -->` on the first line, checked before any overwrite, auto-heal deletion or uninstall removal, with the installer's own backup restored on removal.

A bash defect found along the way: `if ! { ...; } > file; then` does not propagate a redirection failure through the negation, so a failed write silently took the success branch. Fixed with an explicit `$?` check and a regression test.

## Verification

Real `env HOME=<scratch>` install, uninstall and `runtime-bindings` runs across the companion state space. `test_codex_skills.py` 110 passed plus `--clean-clone`; `check-codex-skill-sync.sh` OK at 4 skills; `bin/tests` green; build, drift and symlinks clean; all five budget gates exit 0 with `check-skill-embed-budget` unchanged at 5,716 B headroom.

## Accepted debt

- The engineer time-boxed and did not individually mutation-verify all nine Minors from the round-5 review, relying on code inspection plus shared verification for the two most load-bearing (dispatch, install guards).
- Any findings from the round-6 review still open at merge time are carried; none is a Critical.

## Not doable by any agent

Injection verified intact in a fresh Codex session. Operator-only; runbook in `docs/planning/ds-183-186-harness-trigger-load.md`.

## Review rigor

Skeptic rounds (tier): 4 on the diff across 6 engineer rounds. Round 1 found 1 Critical and 11 Majors. The round cap fired and `decision: "escalate"` was recorded each time with cost-to-date; in hindsight several of those should have been ship-with-debt, which is the bar now applied.

Doc-sync: triggered -> `ADAPTERS.md`, `CONTRIBUTING.md`, `.codex/README.md`, `AGENTS.md`, `.codex/install.sh` summary.

## North Star alignment

Advances Pillar 5 (guard agent context): removes ~135 KB from the always-resident set of every Codex session and puts the artifact under a budget gate. Named trade-off against Pillar 2 (verifiable outcomes): the body reaches the model only on trigger, with the degrade path as the compensating control - which is why that path took the most review attention in this unit.


tyson-solara6:
Superseded by the rebased branch onto current main.

PR #777, opened 2026-08-20
DS-184: trigger-load the Gemini methodology via a project-local skill

tyson-solara6:
Moves the methodology body out of `.gemini/GEMINI.md`, loaded unconditionally in every Gemini session, into `.gemini/skills/dinostack/SKILL.md` as a trigger-loaded Agent Skill.

**Measured:** `GEMINI.md` 133,014 B -> 2,119 B; `SKILL.md` 133,457 B.

## Also in this PR

A marker-based degrade path for when the skill is unreachable, a single destination classifier (`_ae_classify_gemini_md_dst`) that both install branches call so they cannot disagree about ownership, `scripts/check-gemini-skill-budget.sh` with its own suite and an advisory workflow, `hooks/pre-commit` staging the new generated artifact, and docs-currency across `ADAPTERS.md`, `README.md`, `docs/index.html`, `docs/components.md` and `.gemini/README.md`.

`hooks/skill-auto-load-check.sh` previously skipped Gemini unconditionally, on the premise that GEMINI.md always-loads the methodology. This change makes that premise false, so Gemini now gets the same nudge Claude does.

## A shipped limitation, documented not worked around

Gemini denies `activate_skill` outright in headless runs by default policy (`write.toml`'s non-interactive rule), so a plain non-interactive session gets neither the resident body nor the skill. The two documented overrides are `--approval-mode yolo` and `settings.tools.allowed`. **`install.sh` deliberately sets neither** - weakening a security confirmation on the user's machine is not ours to take. The limitation and both override routes are documented in `.gemini/README.md`.

## Notable history

The symlink-ownership logic took several rounds. Round 3 fixed "don't delete the user's foreign symlink" by skipping every symlink, which made the degrade path refuse to deliver the body. Round 4 fixed that on one branch only. Round 5 added a resolved-path comparison. Round 6 extracted the shared classifier but dropped the both-sides canonicalization, so on any aliased checkout path (macOS `/tmp`, a symlinked `~/dev`) the installer's own symlink classified as foreign and the body was never written. Round 7 restores it and adds a scenario that manufactures an aliased path with `ln -s` so CI actually exercises it.

## Verification

Full destination-state matrix (12 states x both branches) through real `env HOME=<scratch> bash .gemini/install.sh`, at both aliased and non-aliased checkout paths. No cell loses user data; no cell leaves the user with neither stub nor body. The class-merge mutation reddens 7 assertions; the aliased-path revert reddens exactly scenario 8. `bin/tests` 1748 passed, all shell suites green, build, drift and symlinks clean.

## Accepted debt

- `bin/tests/test_frontmatter_validity.py` conflicts with the merged DS-185 change; resolution is to keep this branch's docstring-parsing derivation with DS-185's Kimi `DiscoverySpec` layered on and the count bumped.
- Any findings from the round-7 review still open at merge time are carried; none is a Critical.

## Not doable by any agent

Injection verified intact in a live Gemini session. Operator-only; runbook in `docs/planning/ds-183-186-harness-trigger-load.md`.

## Review rigor

Skeptic rounds (tier): 6 on the diff across 7 engineer rounds. The round cap fired repeatedly and `decision: "escalate"` was recorded each time; several should have been ship-with-debt. Two conductor-side defects cost rounds here: a misstated `qa_criteria` block in all four sibling briefs, and priming a reviewer with a specific suspected defect. Both recorded in `.agentic/learnings.md` KNW-20260819-004.

Doc-sync: triggered -> `ADAPTERS.md`, `README.md`, `docs/index.html`, `docs/components.md`, `.gemini/README.md`, both module manifests.

## North Star alignment

Advances Pillar 5 (guard agent context): removes ~131 KB from the always-resident set of every Gemini session. Named trade-off against Pillar 2 (verifiable outcomes): in headless runs the body does not load at all, disclosed above and in the shipped README rather than papered over by silently weakening the user's approval posture.


tyson-solara6:
Superseded by the rebased branch onto current main.

PR #794, opened 2026-08-22
DS-190: advisory nested-worktree spawn guard (once per session per worktree)

tyson-solara6:
## Summary
- New advisory-only PreToolUse hook `hooks/enforce-nested-worktree-spawn.py`: when a MAIN session (no agent_id) issues an Agent/Task spawn with its cwd inside a git worktree, the model gets a once-per-session-per-worktree advisory (worktrees are subagent-reserved; conductors drive from the primary checkout). Never a deny - no deny branch exists in the file.
- Session-scoped suppression via `<cwd>/.agentic/.nested-worktree-spawn-<session_id>.json` (parent dir created; every I/O failure fails open toward emitting, never toward blocking). Kill-switch: `AE_NESTED_WORKTREE_GUARD_DISABLE=1`.
- Full registration: guarded install.sh entry under both Task and Agent matchers, ds-doctor basename, all 13 count-sync sites + the separately-pinned allow_advisory enumeration (three -> four hooks), kill-switch added to both CI-pinned doc surfaces.
- 17-assertion test file with a real `git worktree add` fixture; mutation-verified by review (mkdir removal, agent_id-gate inversion, fail-open branch, per-worktree dedup all redden).

## Jira
Closes [DS-190](https://solara6.atlassian.net/browse/DS-190)

## Test plan
- [x] `hooks/tests/test-enforce-nested-worktree-spawn.py`: 17/17 (incl. absent-.agentic precondition asserted, .agentic-as-file fail-open, same-session different-worktree re-fire)
- [x] `pytest bin/tests/test_tracker_writeback_ranking_spec.py -k "enforcer_subcount or decision_enumeration"`: green at 13 hooks
- [x] `pytest bin/tests/test_docs_currency_sync.py`: green with the new kill-switch in both surfaces
- [x] Full bin/tests (1825 passed), 19-file hooks-python sweep, 55-file hooks-JS sweep, adapter regen byte-identical to a fresh build

## North Star alignment
Advisory-only by design: the field that would scope a hard block to worktree-creating spawns (`tool_input.isolation`) is unverified in the live payload, and blocking every spawn from a worktree cwd would break the legitimate reopened-feature-worktree session - so the guard surfaces the session shape without gating capability (per the no-gating-on-inferred-capability prohibition). The advisory is bounded to once per session per worktree, guarding operator attention. Retirement condition: retire the hook if Claude Code is confirmed to always resolve worktree creation against CLAUDE_PROJECT_DIR regardless of session cwd.

## Review rigor notes
- Plan: 3 Skeptic rounds (count-sync enumeration corrected to 13 sites; separately-pinned decision enumeration caught; two CI-pinned kill-switch doc surfaces added; mechanism claim replaced with a verifiable session-shape anchor; suppression added after an unbounded-noise finding; a false citation and a missing mkdir caught in round 2).
- Code: sign-off round 1 with 6 Minors, 5 fixed in round 2 (mutation-confirmed). Accepted debt (Minor): the advisory-text `{cwd}` interpolation and its fallback have no reddening test assertion - a one-line `assert <worktree path> in reason` in a future touch of the test file closes it.
- Dormant until an installer re-syncs the hooks snapshot; run `.claude/install.sh` or `/ds-update` after merge.

Developer: tyson-solara6
Model: claude-fable-5

[DS-190]: https://solara6.atlassian.net/browse/DS-190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tyson-solara6:
**Automated review (self-authored PR, no formal verdict per review policy)**

Ran skeptic, security-auditor, qa-engineer, engineer (design), and product-discovery (North Star alignment) against this PR. Synthesis:

**Security - Medium: path traversal via unsanitized `session_id`** (`hooks/enforce-nested-worktree-spawn.py`, ~line 550-587)
`session_id` comes from the untrusted PreToolUse JSON payload and is only checked for "non-empty string" before being interpolated into `state_dir / f".nested-worktree-spawn-{session_id}.json"`. A `session_id` containing `../` could make `state_path` escape `.agentic/`. Exploitation is constrained today (only `state_dir` itself is `mkdir`'d, and the flow only creates new files, never overwrites existing ones), but there's no sanitization. Recommend restricting `session_id` to a safe charset (e.g. `^[A-Za-z0-9_-]+$`) before building the path, and rejecting/dropping otherwise (fail open to "no advisory" is fine given this hook's advisory-only design).
Also flagged (Low/informational): predictable state-file path could be pre-seeded with a broken symlink by a co-resident local actor, redirecting the `write_text()` target - same trust tier as the running session, so limited impact, but worth the same input sanitization fix.

**Skeptic - 2 Minor, no Critical/Major, sign-off granted**
- Check-then-write race on the suppression state file: two spawns issued close together in the same session could both observe `exists() == False` and both fire the advisory before either write lands. Bounded to a duplicate advisory (never a duplicate/incorrect deny, since there is no deny branch) - acceptable given the "best-effort" framing already in the docstring, but worth a one-line callout in the Failure modes section.
- `is_git_worktree()` (shared, unchanged) only recognizes a worktree when `cwd` is exactly the worktree root. If the reported `cwd` for a main session is ever a subdirectory of a worktree rather than the root, the advisory silently never fires for that session. Real coverage gap for the "once per session per worktree" goal, not called out in the PR's Failure modes docstring.

**QA - test coverage solid**
11-12 scenarios covered with real `git worktree add` fixtures: first-fire, suppression, non-worktree silence, subagent silence, non-Task/Agent passthrough, malformed/empty stdin, kill-switch, missing cwd/session_id, mkdir-fails-open, per-worktree dedup keying. No deny-path test exists, which is correct (advisory-only). Minor gaps: no test for a genuinely missing `.agentic/` parent path, and no direct assertion on state-file path safety (relevant now given the security finding above - worth adding a test once the sanitization fix lands).

**Design - clean, no blocking issues**
Mirrors the sibling `enforce-worktree-read.py` dynamic-import pattern, correct `hookSpecificOutput`/`permissionDecision` JSON shape, always exits 0. Doc/adapter sync verified complete across `.copilot`, `.cursor`, `.gemini`, `.hermes`, `content/`, `docs/`, `hooks/AGENTS.md`, `bin/ds-doctor`, and `install.sh` registration (guarded `test -f ... && python3 ... || exit 0` form, matching sibling hooks). One nit: a first-commit bug (`ADVISORY_TEMPLATE.format(cwd=cwd)` against a template with no `{cwd}` placeholder) was already caught and fixed in the second commit - no action needed.

**North Star alignment - ALIGNED**
Model-facing, once-per-session advisory that is never a deny; nudges correct conductor/subagent worktree usage with negligible attention tax while reinforcing verifiable delegation structure.

**Bottom line:** fix the `session_id` sanitization (Medium security) before merging; the two Skeptic Minors and QA gaps are worth a follow-up but not blocking. Everything else is solid.

PR #799, opened 2026-08-23
DS-170: extend the knowledge-commit candidate set to AGENTS.md and .agentic/tracking.md

tyson-solara6:
## Summary
- Extends the knowledge-commit candidate set at both sites (Phase 11e and `/ds-wrap` Part G) from 3 files to 5, adding `AGENTS.md` and `.agentic/tracking.md`.
- Adds a `knowledge_commit_exclude` config toggle (default `[]`, exact-string match against the five candidate entries) so a repo can opt a file out without a code branch.
- Adds a per-file, pre-staging revert-risk gate: a candidate whose copy deletes lines relative to the branch tip is skipped alone, before it is ever staged, and the pre-existing aggregate fail-closed guard is left byte-frozen alongside it.
- Resets a skipped file's temp-index entry to the tip blob, because `git update-index -q --refresh` silently restages differing working-tree content (measured, git 2.55.0) and would otherwise defeat the gate.
- Bumps the toggle count twenty-three to twenty-four across all 8 count-sync entries (6 paths, 4 grammatical forms) and updates every candidate-set enumeration, including two slide decks and four `docs/index.html` sites.

## Jira
Closes [DS-170](https://solara6.atlassian.net/browse/DS-170)

## North Star alignment
**Works for everyone (universality).** The opt-out is a config toggle rather than a code branch, so consumer projects get the mechanical guarantee with zero migration while a repo with a different convention can diverge locally. **Verifiable outcomes.** A session that edits `AGENTS.md` or `.agentic/tracking.md` can no longer end with those edits stranded outside its own PR. **Prevent at the producing step.** The mechanism replaces "someone remembers to stage it" with a gate at the point of commit. No pillar regression identified; the two operator decisions below are the alignment-relevant risks and are recorded explicitly.

## Operator decisions recorded

**1. `knowledge_commit_exclude` exists so DinoStack itself stays excluded.** Adding `AGENTS.md` would otherwise make knowledge-commit live in this repo for the first time (it is the only candidate here that is tracked and un-ignored), contradicting the standing 2026-08-11 decision that session learnings are never written to `AGENTS.md` here. That decision stands. This repo sets `knowledge_commit_exclude: ["AGENTS.md"]` in its own gitignored `.agentic/config.json`; that edit is local and deliberately not carried by this PR.

**2. `THRESHOLD_BYTES` raised, deliberately, against the standing "never raise it" convention.** `scripts/check-command-file-budget.sh` goes 373,000 to 395,000, derived at ~2% over a measured 387,197 B per that gate's own stated convention. This is an operator-approved override, not drift. Note an earlier revision of this branch used a ~10% derivation borrowed from `DELTA_LIMIT_BYTES`; that was wrong and was corrected.

**Behavioral change beyond the ticket's literal scope:** the per-file revert-risk gate applies to the three original candidates too, and is not gated on `auto_merge_on_ci_green`. Without that, adding a tracked file to a set whose guards were auto-merge-only would have let Phase 11e commit the conductor's stale `AGENTS.md` over an engineer's newer one on a green PR. A legitimate operator compaction of `MEMORY.md` or `decisions.md` therefore no longer auto-commits and routes to `/ds-wrap` Part G, which is what the existing warning already directs.

## Test plan
- [x] `bin/tests/` full suite: 1858 passed, 25 subtests
- [x] `hooks-js-tests` suite (55 files, wrap-lock pair run separately)
- [x] `check-command-file-budget` / `check-skill-embed-budget` / `check-resident-budget` all OK
- [x] `build-all.sh` then `git diff --exit-code` over the verbatim 17-path adapter-drift pathspec: clean
- [x] `check-methodology-drift.sh` OK; `build-slides.sh` regenerated both edited decks
- [x] Mutation-verified: reverting the gate to post-staging `git rm --cached` reddens the tree-content assertion; removing the `KC_REVERT_SKIP_COUNT` branch reddens the all-skipped status test; the pre-fix branch ordering reddens the reset-precedence test
- [x] Pre-fix execution confirmed the shipped-default revert really occurred (`status=committed` with the stale blob in the pushed tree) and that the fix closes it

## Accepted debt
- The aggregate `awk -v kc_branch` warning has no behavioral test coverage; it can be gutted with CI green. Reachable via the suite's existing `install_git_stub`, just not attempted here.
- `git_fixture.py`'s shape count is asserted "derived" in prose with no mechanical pin. Correct today (6), unpinned.
- In the `KC_N == 0` plus reset-failed path, `deleted_lines` emits `0` rather than the per-file total. Consistent with the sibling commit-failed path; the status names the more severe condition.
- `SHELLS` resolves `bash` via PATH, so bash 3.2 is never exercised in CI. Pre-existing harness property; the new tests were re-run under a 3.2 shim and pass.

Doc-sync: triggered -> updated `docs/index.html` (4 sites), `docs/slides/context-management-slides.md` and `getting-started-slides.md` (plus rebuilt `.html`), `README.md`, `docs/components.md`, `docs/configuration-reference.md`, `content/rules/conventions.md`, `content/references/conventions-detail.md`, `content/references/risk-config-and-tiers.md`, `content/commands/ds-init-project.md`, and `AGENTS.md`.

[DS-170]: https://solara6.atlassian.net/browse/DS-170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tyson-solara6:
Automated review check-in (self-authored PR, no formal verdict): well-documented change with explicit operator decisions and test plan recorded. Currently showing merge conflicts (mergeable: CONFLICTING, mergeStateStatus: DIRTY) against `main` — worth rebasing before merge. No further automated action taken per self-review restriction.

tyson-solara6:
Superseded by #800. This branch went DIRTY/CONFLICTING when DS-194 (#798) landed on main mid-review; force-push is blanket-denied here, so the work was ported to a fresh cut of current main rather than rebased. #800 carries the identical reviewed payload (per-file numstat verified identical across all 70 files) with the generated adapter artifacts regenerated on the new base.

PR #804, opened 2026-08-24
DS-198: detect foreign PreToolUse Agent/Task hooks in ds-doctor

tyson-solara6:
## Summary

Unit 2 of DS-198. Unit 1 already removed the context-mode plugin's `PreToolUse` hook entry (matcher `"Agent"`) from this local install. This unit adds `ds-doctor` detection so a plugin version bump - which reinstalls a fresh cache dir and discards that local edit - does not silently reinstate the hazard.

`check_foreign_agent_hooks` enumerates enabled plugins from `enabledPlugins` in a settings.json resolved as `_plugins_dir()`'s sibling (honors `CLAUDE_CONFIG_DIR`, normally `~/.claude/settings.json`), resolves each `installPath` via `installed_plugins.json`, and inspects the plugin's own packaged `hooks/hooks.json` for a `PreToolUse` entry whose matcher covers `Agent` or `Task`. It reads settings.json only to enumerate `enabledPlugins`, never as a source of hook definitions - a plugin's own `hooks.json` is the sole hook source. FAIL, read-only, no `--fix` repair - mirrors `check_hook_scripts_exist`'s convention exactly (ds-doctor must never touch a third-party plugin's files).

## Round 2 fixes (this push)

**Critical, fixed:** `check_foreign_agent_hooks` read `hooks_data.get("PreToolUse")` at the top level, but a real plugin `hooks.json` nests event arrays under a top-level `"hooks"` key (verified against context-mode's and vercel's actual packaged files - both use the nested shape, none use flat). The shipped check was structurally blind to every real plugin on the machine, including the exact `hooks.json.bak-ds198` file that motivated DS-198. Fixed by accepting both `{"hooks": {"PreToolUse": [...]}}` and a flat `{"PreToolUse": [...]}` form.

**Major, fixed - fixture schema:** the regression test for the "real defect" (T21c) now uses the real nested shape copied from an actual plugin file; flat-shape coverage is retained separately (T21b/T21d).

**Major, fixed - module manifest wording:** corrected to say the check reads settings.json only to enumerate `enabledPlugins`, never as a hook-definition source (it does read it - the prior wording said "never `~/.claude/settings.json`" outright, which was false).

**Major, fixed - `CLAUDE_CONFIG_DIR`:** `_plugins_dir()` now genuinely honors a `CLAUDE_CONFIG_DIR` override (checked directly, before falling back to `_settings_path()`'s sibling), without changing `_settings_path()`'s behavior for any other check in this file. `check_foreign_agent_hooks` derives its own settings_path as a sibling of `plugins_dir` so `enabledPlugins` and `installed_plugins.json` always resolve from the same config root - reading one from `CLAUDE_CONFIG_DIR` and the other from the hardcoded `~/.claude` would silently mismatch. New test (T21i) confirms the override is honored.

**Major, resolved - event-scope conflict:** grepped the plugin tree directly. `hooks/core/routing.mjs` (imported by `sessionstart.mjs` and 5 other platform `sessionstart.mjs` files) DOES import and call `createRoutingBlock` on its Agent branch (lines 324-338) - confirming the earlier investigator's citation, not the Skeptic's "sessionstart.mjs is the only importer" claim. However the LIVE `hooks.json` currently has no `PreToolUse` `Agent` matcher wiring that branch up (unit 1 removed it), so `routing.mjs`'s Agent branch is presently dead on this machine. The `<context_window_protection>` block actually observed in this session's spawn prompt traces to `sessionstart.mjs`'s `SessionStart` hook instead (empty matcher = unconditional injection on every session, including subagent spawns). Documented as an explicit KNOWN LIMITATION (docstring + `docs/updating.md`) rather than widened: `SessionStart`/`UserPromptSubmit` have no matcher-based signal to test against (they don't gate on tool name), and nearly every legitimate context-injection plugin registers a `SessionStart` hook for benign reasons - a presence-based check there would be almost entirely false positives.

**Major, deferred (documented):** project-scope plugin enablement (`installed_plugins.json` records with `"scope": "project"`) is not consulted - only the global settings.json's `enabledPlugins`. Deferred because ds-doctor is a global, cwd-independent tool with no single "the project" to resolve here; documented as a known limitation.

**Minor, fixed:** empty matcher and the `"*"` wildcard are both treated as match-all (new tests T21e/T21f). Exact-token matching false-positive guard now has a fixture (T21g, `AgentX|TaskCreate`). An unparseable settings.json no longer falls through to a false OK (T21h). A non-list `installed_plugins.json` entry now WARNs instead of a silent `continue` (T21j).

Every new/changed assertion (T21c-T21j) was mutation-tested individually: each targeted mutation reddens exactly the intended test and no others.

## Real smoke evidence (post-fix, against actual plugin files)

Prior smoke claim ("reports OK as expected post-unit-1-fix") was vacuous - the schema bug made the check blind, so it would have reported the same OK before the fix too. Re-run against the real resolution chain:

```
=== Real positive case (pre-unit-1 hooks.json.bak-ds198, real nested schema, matcher:Agent) ===
FAIL foreign_agent_hook: plugin 'context-mode@marketplace' registers a PreToolUse hook
(matcher 'Agent') at .../hooks/hooks.json - this can silently rewrite subagent (Task/Agent)
spawn input before Claude Code launches the subagent (DS-198). Disable or uninstall the
plugin, or confirm its hook cannot mutate tool_input, before spawning subagents.

=== Real negative case (current live hooks.json, no Agent matcher post-unit-1-fix) ===
OK foreign_agent_hook: no enabled plugin registers a PreToolUse hook on the Agent/Task tools
```

This is genuine evidence the schema fix works: FAIL against the real historical defect file, OK against the real current file.

## North Star alignment

- **Pillar 2 (Produce verifiable outcomes autonomously) and Pillar 7 (Prevent defects at the producing step)** are the primary pillars this change serves: it turns a silent, re-injectable prompt-mutation hazard into a mechanically detectable `ds-doctor` finding, so recurrence after a plugin update is caught at the next doctor run instead of discovered by an agent noticing its own prompt has been rewritten. Round 2 additionally closes the gap where the check itself was silently non-functional against real plugin files - a check that cannot detect the real shape of the hazard is worse than no check, since it manufactures false confidence.
- **Pillar 1 (Guard operator attention)**: the check produces a single actionable FAIL line naming the plugin, matcher, and file path - it does not add ongoing operator babysitting, since it is read-only and only fires when the hazard is actually present.
- Cuts against nothing directly, but is deliberately narrow (exact-token matcher check on `PreToolUse` only, not general static analysis of arbitrary hook JS) - `SessionStart`/`UserPromptSubmit` injection and project-scope enablement are documented, known false-negative surfaces, not silently unhandled ones.

Doc-sync: triggered - `docs/updating.md`'s `foreign_agent_hook` bullet updated with the event-scope and project-scope limitations. Verified no other doc surface (`content/SKILL.md`, `CONTRIBUTING.md`, `docs/index.html`, `README.md`, `docs/slides/*.md`) enumerates ds-doctor's individual check categories - only general usage is documented there, so no further changes needed.

## Test plan

- [x] `bin/tests/test_agentic_doctor.sh`: T21a-T21j (10 fixtures) - 71 passed, 0 failed (up from the original 3 T21 fixtures + 61 pre-existing = 64).
- [x] Every new/changed T21 assertion mutation-tested individually: each targeted mutation reddens exactly the intended test, all others stay green; reverted after confirming.
- [x] Full `bin/tests/*.sh` sweep: all pass.
- [x] Full `bin/tests` pytest suite (`python3 -m pytest -q -p no:cacheprovider`): 1861 passed, 25 subtests passed.
- [x] Runtime smoke test: real positive case (FAIL) and real negative case (OK) both reproduced against actual plugin files through the real `installed_plugins.json` -> `installPath` -> `hooks/hooks.json` resolution chain (see evidence above).

## Round 3 fixes (this push)

Skeptic round 2 review found three Majors and five Minors on top of round 2's fixes:

**Major, fixed - matcher semantics:** `_matcher_covers_spawn_tool` did exact-token-set comparison, but Claude Code hook matchers are regex PATTERNS matched via `re.search` against the tool name - context-mode's own `PostToolUse` matcher (`"mcp__"`, a prefix, not an exact token) is direct evidence real plugins rely on this. The prior implementation reported OK for `".*"`, `"(Agent|Task)"`, and `"^Agent$"`, all of which must FAIL. Rewritten to `re.search(matcher, "Agent") or re.search(matcher, "Task")`, with a conservative token-match fallback when a matcher fails to compile as a regex (a plugin author's typo) rather than crashing the whole check. New fixtures T21k-T21o cover `.*`, `(Agent|Task)`, `^Agent$`, the `mcp__` prefix (must stay OK), and an uncompilable matcher (must fall back, not crash).

**Major, fixed - test isolation leak:** the round-2 `CLAUDE_CONFIG_DIR` unset was applied only to `t21_invoke`; `invoke_doctor`, the three T20 subshells, and `test_agentic_doctor_ds_prefix.sh`'s `invoke_doctor` never unset it, so a developer machine with a real `CLAUDE_CONFIG_DIR` pointing at an enabled hazardous plugin would fail 5-7 unrelated tests. Fixed in all four sites plus `test_ds_doctor_precommit_fallback_safety.sh`'s single-line invocation. Swept the module for sibling env-var leaks: `CLAUDE_CONFIG_DIR` is the only env var this check reads (the module's other `os.environ` use is the already-scrubbed git-env-override set for git subprocess calls, unrelated to this check). Re-ran the full suite with `CLAUDE_CONFIG_DIR` pointed at a config root containing an enabled hazardous plugin - stays green.

**Major, fixed - manifest/docstring contradiction:** the module docstring and `check_foreign_agent_hooks`'s own docstring both asserted OK for a missing/unparseable `installed_plugins.json`, but the shipped code WARNs (correctly - the check cannot establish which plugins are installed). Corrected both docstrings; re-read the full docstring for further contradictions (found none beyond the ones already fixed here).

**Minor, fixed - "hooks" key masking:** a hooks.json whose top-level `"hooks"` key is present but not an object (`null`, a list, a string) was silently treated as "no PreToolUse hooks here" by `.get("hooks", hooks_data)`'s missing-key-only default, masking a real sibling top-level `"PreToolUse"` key with a false positive OK. Now explicitly checked and WARNed (T21s). Fixing this exposed a related contradiction: a per-plugin WARN-and-skip (malformed plugins-map entry, unparseable hooks.json, or this malformed "hooks" value) no longer lets the final summary print a contradicting "scanned N, all clean" OK on the same run - it now WARNs instead when every enabled plugin hit a skip branch (T21j, T21s).

**Minor, fixed - CLAUDE_CONFIG_DIR expansion:** `_plugins_dir()` now `expanduser()`s a literal `~`-prefixed `CLAUDE_CONFIG_DIR` value (T21u); empty-string and relative values are handled sanely (empty falls back to the `~/.claude` default, relative resolves against cwd).

**Minor, fixed - ambiguous OK message:** the "no plugins enabled at all" and "scanned N plugins, none register a hazard" OK messages are now distinguishable strings (T21v) - previously identical, so an operator couldn't tell settings.json-absent from a completed clean scan.

**Minor, fixed - stale limitation text:** the enablement-scope KNOWN LIMITATION text (module docstring and `docs/updating.md`) said "only `~/.claude/settings.json`" is consulted, contradicting `_plugins_dir()`'s own `CLAUDE_CONFIG_DIR` handling. Both corrected to name the CLAUDE_CONFIG_DIR-aware sibling path.

**Minor, fixed - un-reddenable mutations:** added regression fixtures for four behaviors the round-2 suite could not distinguish from their mutated form: the `if not found_any` OK guard (T21c2), the missing-`matcher`-key default of `""` (T21p), the `enabledPlugins` `val is True` filter rejecting non-boolean-true values (T21q), and the `installPath` dedupe preventing a double-scan of the same resolved path (T21r). All four were confirmed to redden against a hand-applied mutation and restore clean.

Every new/changed T21 assertion (T21c2, T21k-T21v) was individually mutation-tested: each targeted mutation reddens exactly the intended test and stays green everywhere else, then the mutation was reverted and the pre-fix state confirmed byte-identical.

## Round 3 test plan

- [x] `bin/tests/test_agentic_doctor.sh`: 84 passed, 0 failed (T21a-T21v, up from round 2's 71).
- [x] `bin/tests/test_agentic_doctor_ds_prefix.sh`: 21 passed, 0 failed.
- [x] `bin/tests/test_ds_doctor_precommit_fallback_safety.sh`: 7 passed, 0 failed.
- [x] Full `bin/tests/*.sh` sweep: 0 failures across the suite.
- [x] Full `bin/tests` pytest suite (`python3 -m pytest -q -p no:cacheprovider`): 1861 passed, 25 subtests passed.
- [x] Every new/changed T21 assertion (T21c2, T21k-T21v) mutation-tested individually against a hand-edited copy of `bin/ds-doctor`, confirmed reddened, then restored and diffed byte-identical to the pre-mutation file.
- [x] Full suite re-run with `CLAUDE_CONFIG_DIR` pointed at a config root containing an enabled hazardous plugin (matcher `"Agent"`) - all three doctor-invoking test files stay green, confirming the Major 2 fix.

Doc-sync: triggered - `docs/updating.md`'s enablement-scope sentence updated to name the `CLAUDE_CONFIG_DIR`-aware settings.json path, matching `_plugins_dir()`'s existing behavior. No other doc surface enumerates this check's matcher semantics or per-plugin scan behavior.

## Round 4 fixes (this push)

Skeptic round 3 review found two Majors and eight Minors on this same `check_foreign_agent_hooks` defect family.

**Major, fixed - uncompilable-matcher masking:** `_matcher_covers_spawn_tool`'s round-3 "|"-split token-comparison fallback silently returned `False` (a confirmed-clean signal, no WARN, no `scan_incomplete`) for any uncompilable matcher that did not happen to contain a bare `"Agent"`/`"Task"` pipe-token - measured fixtures `"(Agent"`, `"[Agent"`, `"Agent)$["`, and `"(?P<"` all returned `False` under the old fallback, contradicting both this module's own "never report OK when a check could not run" invariant and its own docstring's claim of "erring toward detection". Fixed by letting `re.error` propagate out of the predicate entirely; the caller now treats an uncompilable matcher as UNCONFIRMABLE (WARN + `scan_incomplete`), the same treatment as every other malformed-schema shape this check encounters, rather than guessing True/False for a value it cannot actually inspect. New fixture T21o2 covers the exact non-Agent/Task-containing uncompilable shapes the old fallback missed; T21o updated to assert the new WARN behavior (its own fixture, `"Agent|(unterminated"`, happens to contain a bare `"Agent"` token so it FAILed under the old fallback - it now correctly WARNs like every other uncompilable matcher).

**Major, fixed - manifest omission:** the module docstring's "Upstream deps" enumeration omitted `re` (added at round 3) and `shutil` (pre-existing, never listed). Re-verified the full enumeration against the actual top-level `import` statements rather than assuming only `re` was missing.

**Minor, fixed - `matcher: null` vs missing-key asymmetry:** an explicit `"matcher": null` entry was silently skipped with no WARN, while a MISSING `matcher` key already defaulted to `""` (match-all, FAIL) - the semantically closest case, treated oppositely. Both now default to match-all (T21p2). Any OTHER non-string, non-null matcher type (a list, a number, ...) is a genuinely malformed schema, not an implicit match-all - it now WARNs and marks the scan incomplete (T21p3), unified with the uncompilable-regex treatment above.

**Minor, fixed - six previously silent inspection-skipping paths:** verified the Skeptic's enumeration against the code and confirmed all six. Each now WARNs + sets `scan_incomplete` instead of a bare `continue`: an enabled plugin with no entry at all in `installed_plugins.json`'s plugins map (T21t2), a non-object entry in that map's list (T21t3), a missing/invalid `installPath` (T21t4), a PRESENT-but-wrong-type `PreToolUse` key (T21t5 - explicitly distinguished from an ABSENT `PreToolUse` key, which stays a normal silent `continue` since that's the overwhelmingly common "this plugin only registers other event types" case; T21t5b pins this common case stays a clean, un-warned OK), and a non-object entry in the `PreToolUse` list (T21t6).

**Minor, fixed - unpinned plugin count:** the "scanned N enabled plugin(s)..." OK message's count was previously asserted only to differ from the no-plugins-enabled message's text, never pinned to the actual count - a mutation adding a constant offset to the count passed all 84 pre-existing tests. New fixture T21w pins the exact count with two enabled, cleanly-scanned plugins ("scanned 2 enabled plugin(s), none register...").

**Minor, fixed - `scan_incomplete` docstring:** corrected "if EVERY enabled plugin hits one of those WARN-and-skip branches, the final summary is ALSO a WARN" to "if ANY... hits" - the code sets `scan_incomplete = True` unconditionally on any single occurrence and never clears it, so even a 1-warn/1-clean run WARNs for the whole run, not only when every plugin hits a skip branch.

**Minor, fixed - `re.search` overclaim:** downgraded the docstring's unqualified claim that Claude Code matches hook matchers via `re.search` specifically (as opposed to `re.match`/`re.fullmatch`) - the cited context-mode evidence (a bare `"mcp__"` prefix matcher on a tool name it doesn't equal) only rules out exact-string equality, it does not distinguish among the three regex call semantics. The docstring now states this is unconfirmed and that `re.search` is the conservative (most permissive) choice among them.

**Minor, fixed - uncaught RecursionError:** a pathologically nested, uncompilable-by-recursion-limit matcher (measured: 5000-deep nested regex groups) raises `RecursionError` from `re.compile`, not `re.error` - this escaped the check entirely and aborted the whole `ds-doctor` run with a traceback, contradicting "rather than crashing the whole check". Now caught alongside `re.error` at the same call site. New fixture T21x.

**Minor, deferred (documented) - T21 fixture helper extraction:** the 765-line T21 block's near-identical `installed_plugins.json`/`hooks.json` heredocs would benefit from a `_t21_fixture <key> <matcher>` helper, matching the file's existing style. This round added several more fixture variants (T21o2, T21p2, T21p3, T21t2-T21t6, T21w, T21x) under the same time budget as the fixes above; deferred rather than risk a botched extraction weakening an existing assertion mid-round, per the finding's own "genuinely optional... only if it does not risk weakening any assertion" framing.

Every new/changed assertion (T21o, T21o2, T21p2, T21p3, T21t2-T21t6, T21w, T21x) was individually mutation-tested against a hand-edited copy of `bin/ds-doctor`: each targeted mutation reddens exactly the intended test and no others, confirmed, then restored and diffed sha256-byte-identical to the pre-mutation file.

## Round 4 test plan

- [x] `bin/tests/test_agentic_doctor.sh`: 98 passed, 0 failed (up from round 3's 84).
- [x] `bin/tests/test_agentic_doctor_ds_prefix.sh`: 21 passed, 0 failed.
- [x] `bin/tests/test_ds_doctor_precommit_fallback_safety.sh`: 7 passed, 0 failed.
- [x] `bin/tests/test_ds_doctor_precommit_source_parity.sh`: 9 passed, 0 failed.
- [x] Full `bin/tests/*.sh` sweep: 0 failed scripts.
- [x] Full `bin/tests` pytest suite (`python3 -m pytest -q -p no:cacheprovider`): 1861 passed, 25 subtests passed.
- [x] Every new/changed T21 assertion (T21o, T21o2, T21p2, T21p3, T21t2-T21t6, T21w, T21x) mutation-tested individually, confirmed reddened, then restored and diffed byte-identical to the pre-mutation file.
- [x] Hazard-env re-run: `CLAUDE_CONFIG_DIR` pointed at a config root with an enabled plugin registering `matcher: "Agent"` - all three doctor-invoking suites stay green; a direct `ds-doctor` invocation against the same hazard fixture correctly FAILs naming the plugin:
  ```
  FAIL foreign_agent_hook: plugin 'hazardous@mkt' registers a PreToolUse hook (matcher 'Agent')
  at .../plugin-installs/hazardous/hooks/hooks.json - this can silently rewrite subagent
  (Task/Agent) spawn input before Claude Code launches the subagent (DS-198). Disable or
  uninstall the plugin, or confirm its hook cannot mutate tool_input, before spawning subagents.
  ```

Doc-sync: predicate not triggered - `docs/updating.md`'s `foreign_agent_hook` bullet describes the check's event/enablement scope only, not its per-branch WARN/FAIL granularity or matcher-compilation semantics; none of round 4's fixes change what that bullet asserts. No other doc surface (`content/SKILL.md`, `CONTRIBUTING.md`, `docs/index.html`, `README.md`, `docs/slides/*.md`) enumerates this check's internal behavior.

## Round 5 fixes (this push)

Skeptic round 4 review found two Majors and eight Minors; both Majors were the same shape (a defect CLASS addressed by enumerating known instances rather than closing the class).

**Major, fixed - wrong-typed settings.json yields a false positive OK:** a settings.json that parses but has the wrong shape (`enabledPlugins` as a list or string, or a non-dict top-level document) silently produced `enabled_plugins=[]` and printed `OK ... no plugins enabled`, exit 0, while a hazardous, enabled plugin was present in `installed_plugins.json`. Reproduced directly against three wrong-type fixtures (list, string, non-dict top-level) before fixing - all three printed the false OK; a correct-shape control FAILed as expected. Fixed by closing the class: settings.json's parse now WARNs whenever a value it reads is present but not the expected JSON type (non-dict top-level, or an `enabledPlugins` key present but not itself an object), rather than silently coercing to an empty/falsy default indistinguishable from a legitimately empty or absent `enabledPlugins`. New fixtures T21y (list-typed `enabledPlugins`) and T21y2 (non-dict top-level settings.json).

**Major, fixed - `OverflowError` escapes the exception tuple and aborts the whole run:** `re.compile` raises `OverflowError` ("the repetition number is too large") on a matcher like `Agent{4294967296}` - not in the `(re.error, RecursionError)` tuple the caller checked, so it escaped `check_foreign_agent_hooks` and aborted the entire `ds-doctor` run with a traceback (confirmed: under `--json` this produced a traceback on stderr and zero JSON bytes on stdout, exit 1 - a consumer parsing the JSON gets nothing and cannot distinguish a crash from findings). Reproduced directly before fixing. Fixed by catching by CLASS (`except Exception`) at the call site rather than maintaining a ban-list of named exception types for untrusted third-party regex input - the docstring now states explicitly that this predicate's exception surface for malformed/pathological patterns is not a fixed, enumerable set. New fixtures T21z (plain mode) and T21z2 (`--json` mode still emits well-formed JSON).

Both fixes close the underlying class ("a value/exception this check cannot fully characterize must never be silently treated as a known-safe case") rather than adding the two newly-named instances to an existing enumeration - the module docstring and `check_foreign_agent_hooks`'s own docstring were rewritten to state this as one general invariant instead of an enumerated list, which also resolves round-5 Minor 5 (WARN-shape enumerations under-listing 6/11 and 3/11 branches).

**Minor, reconsidered and KEPT (reasoned disagreement) - `matcher: null`:** the Skeptic recommended reversing round 4's decision (treat `matcher: null` identically to a missing key, i.e. match-all/FAIL) to the UNCONFIRMABLE/WARN path, since vercel 0.45.1 was observed shipping `matcher: null` only on `SessionEnd` (an event with no tool-name matching at all) - so that observation says nothing about `PreToolUse` semantics specifically. Weighed and kept match-all: this same function's docstring already resolves an identically-unverified choice (`re.search` vs. `match`/`fullmatch`) in favor of the conservative, least-likely-to-under-detect reading, and a JS-hosted hook-matching implementation is far more likely to test a matcher with a bare falsy check (under which `null` and an omitted key are indistinguishable) than to special-case `null` into a narrower semantics unsupported by the missing-key case. Reversing this would be the one place in the function where an ambiguous match-all-shaped value is resolved away from the conservative reading, in a security check whose whole purpose is to not under-detect. Documented the full reasoning inline; T21p2 continues to pin the current (match-all/FAIL) behavior.

**Minor, fixed - "Upstream deps" scope mismatch:** the header claimed "Python 3 stdlib only" but omitted function-local `importlib.machinery`/`importlib.util`/`glob` (all stdlib, used by `--cross-harness` support code) and the optional third-party `yaml` import (guarded by `try/except ImportError`, `--cross-harness`-only). Corrected to enumerate both categories and note `yaml`'s absence degrades to a WARN, not an ImportError.

**Minor, fixed - untrusted matcher echoed unbounded:** T21x's own 10,001-character pathological matcher printed in full in the WARN message. Added `_truncate_for_message()` (200-char limit, explicit `...<+N more chars, truncated>` elision marker) and applied it to both the compile-failure WARN and the non-string-matcher WARN's `repr()` echo.

**Minor, fixed - missing empty-string `installPath` fixture:** T21t4 covered only the missing-`installPath`-key case; the empty-string sub-clause (`not install_path_raw`) had no fixture of its own. Added T21t4b.

**Minor, fixed - stale `_run_doctor` line-number citation:** `bin/ds-update`'s `_run_doctor` docstring cited `bin/ds-doctor:947` for the `_os.getcwd()` role-models.yml probe; the real line had already drifted to 1851 pre-round-5 and moved again to 1970 after this round's edits. Replaced the line-number citation with a function/behavior description plus a grep pointer, since a line-number citation in this docstring has now gone stale twice from unrelated edits to the target file.

**Minor, re-deferred (unchanged) - T21 fixture duplication:** the Skeptic re-scanned the now-larger T21 block for copy-paste divergence and found none, accepting the prior deferral. Re-deferred again on the same reasoning: the block is now ~1,250 lines of near-identical heredocs; a `_t21_fixture` helper extraction remains genuinely optional and risks weakening an existing assertion mid-round if rushed.

**Minor, updated (this section) - stale PR-body smoke evidence:** the "Real smoke evidence" block above dated to round 2 output and had not reflected any fixture actually exercised since round 3. See the round 5 smoke evidence below, which replaces it as the current record.

Every new/changed test (T21y, T21y2, T21z, T21z2, T21t4b) was individually mutation-tested against a hand-edited copy of `bin/ds-doctor`: each targeted mutation reddens exactly the intended test and no others, confirmed, then restored and diffed sha256-byte-identical to the pre-mutation file.

## Round 5 smoke evidence (supersedes the round-2 block above, which was stale since round 3)

Real machine, current live config (`~/.claude-spacedinosaurs`, one enabled plugin):

```
OK foreign_agent_hook: scanned 1 enabled plugin(s), none register a PreToolUse hook on the Agent/Task tools
```

Major 1 pre-fix repro (list-typed `enabledPlugins`, hazardous plugin present and enabled) - false OK reproduced against the unfixed code, then FAIL against the fixed code with a correct-shape control:

```
=== pre-fix: list-typed enabledPlugins, hazardous plugin enabled ===
OK foreign_agent_hook: no plugins enabled (no enabledPlugins entries in <settings.json>) - nothing to scan

=== post-fix: same fixture ===
WARN foreign_agent_hook: <settings.json>: 'enabledPlugins' is present but not an object - could not determine which plugins are enabled

=== control: correct dict-shape enabledPlugins, same hazardous plugin ===
FAIL foreign_agent_hook: plugin 'evil@mkt' registers a PreToolUse hook (matcher 'Agent') at <hooks.json> - this can silently rewrite subagent (Task/Agent) spawn input before Claude Code launches the subagent (DS-198). Disable or uninstall the plugin, or confirm its hook cannot mutate tool_input, before spawning subagents.
```

Major 2 pre-fix repro (`Agent{4294967296}` matcher, `OverflowError`) - uncaught traceback reproduced against the unfixed code, then WARN against the fixed code with well-formed `--json` output:

```
=== pre-fix ===
Traceback (most recent call last):
  ...
  File ".../bin/ds-doctor", line 267, in _matcher_covers_spawn_tool
    pattern = re.compile(matcher)
  ...
OverflowError: the repetition number is too large

=== post-fix ===
WARN foreign_agent_hook: <hooks.json>: plugin 'overflow@mkt' PreToolUse matcher 'Agent{4294967296}' could not be evaluated as a regex (the repetition number is too large) - could not confirm this hook does not cover Agent/Task, skipping
```

## Round 5 test plan

- [x] `bin/tests/test_agentic_doctor.sh`: 103 passed, 0 failed (up from round 4's 98; +5: T21t4b, T21y, T21y2, T21z, T21z2).
- [x] `bin/tests/test_agentic_doctor_ds_prefix.sh`: pass.
- [x] `bin/tests/test_ds_doctor_precommit_fallback_safety.sh`: pass.
- [x] `bin/tests/test_ds_doctor_precommit_source_parity.sh`: pass.
- [x] Full `bin/tests/*.sh` sweep: 0 failed scripts.
- [x] Full `bin/tests` pytest suite (`python3 -m pytest -q -p no:cacheprovider`): pass.
- [x] Both Majors reproduced directly against the unfixed code before fixing (see smoke evidence above), then re-verified fixed.
- [x] Every new/changed test (T21y, T21y2, T21z, T21z2, T21t4b) mutation-tested individually, confirmed reddened, then restored and diffed sha256-byte-identical to the pre-mutation file.
- [x] Hazard-env re-run across all doctor-invoking suites with `CLAUDE_CONFIG_DIR` pointed at a config root with an enabled plugin registering a matching matcher: green.
- [x] `--json` mode verified well-formed on the new OverflowError WARN path (T21z2) and on the wrong-typed-settings.json WARN path (manual check).

## Round 6 fixes (this push)

**Major, fixed:** a truthy non-boolean `enabledPlugins` ENTRY VALUE (e.g. `1`, `"true"`, a nested object) coerced to a false-clean OK via `val is True`'s strict identity check - the same false-clean-OK class round 5 closed for the CONTAINER (`settings_data`, `enabledPlugins`), left open one level deeper for VALUES inside it. Under JS `if (enabled)` truthiness semantics, `1`/`"true"` mean ENABLED, so this under-detected an actually-enabled hazardous plugin as "nothing to scan." Fixed by treating any non-boolean value as unconfirmable (WARN, abort the whole check) rather than silently coercing to disabled. Full sweep of every remaining wrong-typed-value site at every nesting depth (settings.json, installed_plugins.json, each plugin's hooks.json) found no other coercion-to-default site - every other JSON value this check reads is already followed by an `isinstance` guard with a WARN fallback.

**Major, fixed:** the module manifest's failure-mode paragraph asserted a single exceptionless rule ("MISSING or unconfirmable -> WARN, never OK and never FAIL") that is false on three shipped, intentional paths: a missing `settings.json` is a positive OK (known-clean), a plugin's missing `hooks.json` is a silent `continue` that still feeds the "scanned N" OK claim (most plugins register no hooks at all - normal, not a hazard), and a missing `matcher` key is treated as match-all by design and can FAIL. The function-level docstring was already accurate (it scopes to "could not confirm" with correct MISSING examples); only the module-manifest summary overclaimed. Fixed by splitting the invariant: MISSING varies deliberately by site (pointer to the function docstring for exact per-site treatment, no attempt to re-enumerate every shape here), while PRESENT-but-wrong-typed stays the one exceptionless rule (verified true against every read site in this round's sweep, not just the three named ones).

**Minor, fixed:** `_truncate_for_message`'s docstring claimed it guards both `doc.warn()` and `doc.fail()` echoes, but the `doc.fail()` matcher echo was untruncated - a hazardous `"(?:Agent|" + "z"*10000 + ")"` matcher would print a ~10 KB FAIL line. Applied `_truncate_for_message` at that call site too.

**Minor, fixed:** no test reddened `_truncate_for_message` itself - neutering it to return its input unchanged left the whole suite green. Added T21z3 (a pathologically long hazardous matcher hitting the now-truncated FAIL path), which serves both as the regression test for the Minor 3 fix above and as the first test that actually exercises `_truncate_for_message`'s truncation behavior.

**Minor, fixed:** an inline rationale comment for keeping `"matcher": null` as match-all cited a specific observation ("vercel 0.45.1 was observed shipping `\"matcher\": null` only on SessionEnd") that does not exist in that plugin's actual `hooks.json` - checked directly (`~/.claude/plugins/cache/claude-plugins-official/vercel/0.45.1/hooks/hooks.json`): its `SessionEnd` entry has no `matcher` key at all (missing, not null), and no `"matcher": null` appears anywhere in the file. This was a fabricated factual attestation. Replaced with an honest "unobserved" statement plus a description of the concrete probe that would settle it (install a throwaway plugin with an explicit `"matcher": null` PreToolUse entry, run any tool, observe whether the hook fires) - that probe was NOT run this round, per the brief's instruction. The underlying decision (keep match-all/FAIL as the conservative reading) is unchanged; only the false citation was removed.

**Pre-existing test corrected:** T21q previously pinned the OLD (now-incorrect) behavior - it asserted that `enabledPlugins: {"disabled@mkt": "true"}` (a truthy string) is silently treated as disabled and reports a clean OK. That is exactly the under-detection class Major 1 closes. T21q's fixture was changed to JSON `false` (the one value that is unambiguously "disabled" under any type-checking discipline); the truthy-string/truthy-int case is now covered by the new T21y3, which asserts the corrected WARN behavior.

## Round 6 smoke evidence

Major 1 pre-fix repro (truthy non-boolean `enabledPlugins` value `1`, hazardous plugin present and enabled) - false OK reproduced against the unfixed code, then WARN against the fixed code:

```
=== pre-fix: enabledPlugins: {"evil@mkt": 1} ===
OK foreign_agent_hook: no plugins enabled (no enabledPlugins entries in <settings.json>) - nothing to scan

=== post-fix: same fixture ===
WARN foreign_agent_hook: <settings.json>: 'enabledPlugins' has non-boolean value(s) for evil@mkt - could not confirm which plugins are enabled
```

Wrong-typed-value sweep (post-fix), confirming the WARN/OK split is correct for every JSON-typed value in `enabledPlugins`:

```
val=1            -> WARN (non-boolean)
val="true"       -> WARN (non-boolean)
val=true         -> proceeds to scan (correctly ENABLED)
val=false        -> OK "no plugins enabled" (correctly DISABLED)
val={"enabled":true} -> WARN (non-boolean)
val=[1,2]        -> WARN (non-boolean)
```

Minor 3 fix, `--json` well-formed with the truncation marker present on the FAIL path:

```
=== --json output for a 10,000-char hazardous FAIL matcher ===
well-formed JSON: yes
"more chars, truncated" present in FAIL message: yes
```

Real machine, current live config (`~/.claude-spacedinosaurs`, one enabled plugin) - unchanged from round 5, confirming no regression:

```
OK foreign_agent_hook: scanned 1 enabled plugin(s), none register a PreToolUse hook on the Agent/Task tools
```

## Round 6 test plan

- [x] Major 1 reproduced directly against the unfixed code before fixing (see smoke evidence above), then re-verified fixed.
- [x] Full wrong-typed-value sweep across every nesting depth this check reads (settings.json/enabledPlugins entries, installed_plugins.json/plugins map/per-plugin entries/installPath, each plugin's hooks.json/hooks-wrapper/PreToolUse list/PreToolUse entries/matcher) - every site already guarded by an `isinstance` check with a WARN fallback; the only unguarded coercion-to-default site was the one Major 1 fixed.
- [x] `bin/tests/test_agentic_doctor.sh`: 105 passed, 0 failed (up from round 5's 103; +2 new: T21y3, T21z3; T21q's fixture corrected in place, same test count contribution).
- [x] `bin/tests/test_agentic_doctor_ds_prefix.sh`: 21 passed, 0 failed.
- [x] `bin/tests/test_ds_doctor_precommit_fallback_safety.sh`: 7 passed, 0 failed.
- [x] `bin/tests/test_ds_doctor_precommit_source_parity.sh`: 9 passed, 0 failed.
- [x] Full `bin/tests/*.sh` sweep: 0 failed scripts.
- [x] Full `bin/tests` pytest suite (`python3 -m pytest -q -p no:cacheprovider`): 1861 passed, 25 subtests passed, 0 failed.
- [x] T21y3 (Major 1 regression) mutation-tested: reverting the non-boolean guard reddens exactly T21y3 (103 passed, 1 failed), all other tests including T21q stay green; restored byte-identical (sha256-verified).
- [x] T21z3 (Minor 3/4 regression) mutation-tested twice: (a) reverting the FAIL-site `_truncate_for_message` call reddens exactly T21z3; (b) neutering `_truncate_for_message` itself to an identity function also reddens exactly T21z3, with T21x/T21z (the other matcher-echo tests) staying green. Both mutations restored byte-identical (sha256-verified).
- [x] Hazard-env re-run: real config roots (`~/.claude/settings.json` and `CLAUDE_CONFIG_DIR=~/.claude-spacedinosaurs`) both report the unchanged clean OK - no new noise from this round's fixes.
- [x] `--json` mode verified well-formed on both new paths this round: the Major 1 WARN (non-boolean enabledPlugins value) and the Minor 3 FAIL-truncation path (truncation marker present in the JSON message field).
- [x] Diff scope verified clean: 2 files changed (`bin/ds-doctor`, `bin/tests/test_agentic_doctor.sh`), no reverts, 0 added em dashes, no `.claude/skills/dinostack/{agents,commands,references,rules}` symlink drift.

Doc-sync: predicate not triggered - round 6's fixes are internal to `check_foreign_agent_hooks`'s type-checking granularity, its module-manifest wording, and test coverage; `docs/updating.md`'s `foreign_agent_hook` bullet describes event/enablement scope only and is unaffected by this round.

## Round 7 fixes (this push)

Skeptic round 6 review found two Majors and four Minors; six of six had a disposition this round.

**Major, fixed - round 6 regression, aborted the entire scan on the first non-boolean value:** round 6's fix for a truthy non-boolean `enabledPlugins` VALUE (e.g. `1`) treated it identically to a wrong-typed CONTAINER - WARN and abort the WHOLE check via `return`, even though `enabled_plugins` is computed from the very same dict one line below and already supports a partial-list scan. Reproduced directly before fixing with the reviewer's exact fixture (`{"haz@mkt": true, "other@mkt": 1}`, `haz@mkt` shipping a real `matcher: "Agent"` hazard):

```
=== pre-fix (c18afbd6) ===
WARN foreign_agent_hook: <settings.json>: 'enabledPlugins' has non-boolean value(s) for other@mkt - could not confirm which plugins are enabled
(no FAIL line; exit unaffected by this finding)

=== post-fix ===
WARN foreign_agent_hook: <settings.json>: 'enabledPlugins' has non-boolean value(s) for other@mkt - could not confirm whether they are enabled, excluding them from the scan
FAIL foreign_agent_hook: plugin 'haz@mkt' registers a PreToolUse hook (matcher 'Agent') at <hooks.json> - this can silently rewrite subagent (Task/Agent) spawn input before Claude Code launches the subagent (DS-198). Disable or uninstall the plugin, or confirm its hook cannot mutate tool_input, before spawning subagents.
```

Fixed by WARNing and EXCLUDING the non-boolean entries from `enabled_plugins`, then continuing to scan the boolean-`true` ones - the never-report-a-false-OK invariant is preserved via a new `enabled_incomplete` flag that forces a WARN (not OK) if every remaining entry ends up excluded/false, folded into the existing `scan_incomplete` mechanism for the summary line. Removed the inaccurate inline justification ("there is no partial-list concept to fall back to"). New fixture T21y4 pins exactly this fixture shape (a boolean-true hazardous sibling alongside a non-boolean entry).

**Major, fixed - function docstring's blanket MISSING claim was false on three measured paths:** the docstring said every MISSING or unconfirmable value WARNs, never OK/never FAIL. Measured false on three real call paths: a missing `settings.json`/`enabledPlugins` key -> `OK ... no plugins enabled`; a plugin's missing `hooks.json` -> `OK ... scanned N enabled plugin(s), none register...`; a missing `matcher` key -> `FAIL` (match-all by design). The module manifest's own MISSING enumeration already got all three right; only the function docstring it deferred to was wrong. Rewrote the function docstring to state the per-site MISSING treatment accurately (mirroring the module manifest's existing correct summary), keeping the exceptionless PRESENT-but-wrong-typed rule as the one universal WARN case. Verified all three corrected claims by running the actual cases, not by re-reading the code:

```
missing settings.json (plugins dir present)   -> OK foreign_agent_hook: no plugins enabled (no enabledPlugins entries in <settings.json>) - nothing to scan
settings.json present, no enabledPlugins key  -> OK foreign_agent_hook: no plugins enabled (no enabledPlugins entries in <settings.json>) - nothing to scan
enabled plugin, no hooks/hooks.json file      -> OK foreign_agent_hook: scanned 1 enabled plugin(s), none register a PreToolUse hook on the Agent/Task tools
enabled plugin, PreToolUse entry with no matcher key -> FAIL foreign_agent_hook: plugin 'nomatch@mkt' registers a PreToolUse hook (matcher '') at <hooks.json> ...
```

**Minor 3, fixed - `matcher: null` carve-out added to the "exceptionless" rule's prose:** the DECISION to treat `matcher: null` as match-all (endorsed, kept unchanged in round 5/6) is a real, narrow exception to the module manifest's "PRESENT but wrong JSON type -> WARN, never FAIL" rule - both docstrings previously implied it was truly exceptionless. Added an explicit one-sentence carve-out to the module manifest naming `matcher: null` as the sole exception, cross-referenced to the function docstring's existing (unchanged) rationale block. Re-verified behavior-unchanged: `matcher: null` still FAILs post-fix (confirmed by direct invocation, see smoke evidence below).

**Minor 4, fixed - untruncated echo at the new non-boolean-keys WARN site:** applied `_truncate_for_message` to the `', '.join(non_bool_keys)` echo, closing the same unbounded untrusted-echo class this same delta's Major 1 fix touches.

**Minor 5, fixed - understated verification claim:** the round-6 rationale said "no real hooks.json on file has been checked for a `matcher: null` entry on any event"; in fact two were checked (context-mode, vercel) and neither contains one. Re-verified both files directly this round (grepped both for `"matcher": null` and `null` generally - zero occurrences in either) and restated the claim as the stronger, accurate fact.

**Minor 6, accepted as debt - no test guards the two-docstring prose duplication:** the module manifest and the function docstring restate the same MISSING/PRESENT invariant in near-identical prose (how round 6's drift went undetected). No cheap mechanical test can assert semantic agreement between two free-text docstrings - a grep-based test would only catch a wording match, not a truthful one, which is exactly the failure mode this round closes. Documented explicitly rather than left silent.

Every new/changed assertion (T21y4) was mutation-tested: reverting the WARN-and-exclude fix to the round-6 `return` reddens exactly T21y4 (105 passed, 1 failed), all 105 other tests including T21y3 stay green; restored byte-identical (sha256-verified, `bin/ds-doctor` unchanged before/after the mutation-and-restore cycle).

## Round 7 smoke evidence

Major 1 pre-fix repro (Skeptic's exact fixture) reproduced against `c18afbd6`, then re-verified fixed - see the Major 1 write-up above for the full before/after transcript.

Major 2 docstring-accuracy verification (three cited MISSING paths, run against the fixed code) - see the Major 2 write-up above for the full transcript; all three match the corrected docstring text.

Minor 3 (`matcher: null` behavior-unchanged) re-verified post-fix:

```
enabled plugin, PreToolUse entry with matcher: null -> FAIL foreign_agent_hook: plugin 'nullmatch@mkt' registers a PreToolUse hook (matcher '') at <hooks.json> ...
```

`--json` well-formed on the new WARN+FAIL combination path (Major 1 fixture) and on the WARN-only path (all `enabledPlugins` entries excluded) - both parsed cleanly with `json.load`, both entries carry `"status": "WARN"`/`"FAIL"` as expected; exit code reflects FAIL (1) when a hazard is confirmed and is unaffected by the WARN lines' presence.

## Round 7 test plan

- [x] `bin/tests/test_agentic_doctor.sh`: 106 passed, 0 failed (up from round 6's 105; +1 new: T21y4).
- [x] `bin/tests/test_agentic_doctor_ds_prefix.sh`: 21 passed, 0 failed.
- [x] `bin/tests/test_ds_doctor_precommit_fallback_safety.sh`: 7 passed, 0 failed.
- [x] `bin/tests/test_ds_doctor_precommit_source_parity.sh`: 9 passed, 0 failed.
- [x] Full `bin/tests/*.sh` sweep (59 files): 0 failed scripts.
- [x] Full `bin/tests` pytest suite (`python3 -m pytest -q -p no:cacheprovider`, `__pycache__` cleared first): 1861 passed, 25 subtests passed, 0 failed.
- [x] Major 1 reproduced directly against the unfixed code (`c18afbd6`) before fixing, then re-verified fixed (see smoke evidence above).
- [x] Major 2's three corrected docstring claims each run directly against the fixed code, not re-derived from reading the source (see smoke evidence above).
- [x] Minor 3's `matcher: null` behavior re-verified unchanged post-fix (still FAILs).
- [x] T21y4 mutation-tested: reverting to round 6's abort-on-first-non-boolean `return` reddens exactly T21y4 (105 passed, 1 failed), all other tests (including T21y3) stay green; restored byte-identical (sha256-verified).
- [x] `--json` mode verified well-formed on both the new combined WARN+FAIL path and the all-excluded WARN-only path.
- [x] Diff scope verified clean: 2 files changed (`bin/ds-doctor`, `bin/tests/test_agentic_doctor.sh`), no reverts, 0 added em dashes (`git diff --cached -U0 | grep '^+.*—'`), no `.claude/skills/dinostack/{agents,commands,references,rules}` symlink drift (`git status --short` clean on those paths).

Doc-sync: predicate not triggered - round 7's fixes are internal to `check_foreign_agent_hooks`'s non-boolean-value handling and its own/module docstring accuracy; `docs/updating.md`'s `foreign_agent_hook` bullet describes event/enablement scope only and is unaffected by this round. Per the brief's fallback instruction, if the abort-on-first-non-boolean behavior were kept instead of fixed, it would need documenting in `docs/updating.md:63`'s known-limitations list - not needed here since the behavior was fixed rather than kept.


tyson-solara6:
**Automated review comment (self-authored PR — no formal verdict, self-review is blocked on GitHub).**

Ran a skeptic pass and a security-auditor pass on this diff. Both independently converged on the same real gap, which is worth fixing before merge:

**Major/High — false "OK" when an enabled plugin key is absent from `installed_plugins.json`.**
In `check_foreign_agent_hooks` (`bin/ds-doctor`), the loop over `enabled_plugins` does:
```python
entries = plugins_map.get(key)
if entries is not None and not isinstance(entries, list):
    doc.warn(...); scan_incomplete = True; continue
if not isinstance(entries, list):
    continue
```
If `key` is enabled in `settings.json` but simply **missing** from `installed_plugins.json`'s `plugins` map (as opposed to present-but-malformed, which is correctly WARNed), `entries` is `None` and the loop silently `continue`s — no WARN, no `scan_incomplete = True`. That plugin's `hooks.json` is never scanned, yet it's still counted in the final `"scanned N enabled plugin(s), none register..."` OK message. This reproduces exactly the "manufactured false confidence" failure mode this PR's own round-3 fix was built to close for the sibling malformed-entry case — just for the "entirely absent" case instead, which isn't covered by any of the T21a–T21v fixtures.

Suggested fix: treat `entries is None` for an enabled key the same as the non-list-entry branch — WARN + `scan_incomplete = True`.

**Minor (both reviewers, consistent nits):**
- `PreToolUse` not being a list, a hook entry not being a dict, and `matcher` not being a string all fall through to a bare `continue` with no WARN — inconsistent with the PR's own hardened handling of the sibling "hooks" key being malformed (T21s). Same "never silently mask an unparseable shape" principle should probably apply here too.
- The `FileNotFoundError` message for a missing `installed_plugins.json` ("could not parse {path}: {exc}") is misleading — the file doesn't exist at all, it's not a parse failure.
- `_matcher_covers_spawn_tool` compiles an arbitrary regex from a plugin-supplied matcher; low risk since it's only ever matched against short fixed strings ("Agent"/"Task"), so not exploitable as ReDoS here, but worth a one-line note if this pattern is reused elsewhere against untrusted longer input.
- Cross-key `seen_install_paths` dedupe is keyed only by resolved path, not `(key, path)` — if two enabled plugins share a resolved installPath, only the first-seen name gets attributed in output. Cosmetic, not a bypass.

Everything else (nested `hooks.json` schema, regex matcher semantics, `CLAUDE_CONFIG_DIR` test-isolation fix, OK/WARN masking fix) checked out correctly against the diff and tests. No injection/command-execution surface in this diff.

tyson-solara6:
**Automated review comment (self-authored PR — no formal verdict, self-review is blocked on GitHub).**

Ran skeptic (blocked on missing formal Global-context inputs — expected for an ad-hoc review, not re-run), a security-auditor pass, and a static test-coverage pass against the current head. The previously-flagged `entries is None` silent-continue bug is now fixed and has a regression test (T21t2) — confirmed.

**High — log/output injection via unescaped `matcher` in the FAIL message.**
`check_foreign_agent_hooks`'s `doc.fail(...)` call interpolates `matcher` raw (no `_truncate_for_message`, unlike every other echo site in this function, and truncation alone wouldn't help — it only bounds length, not control characters). A matcher like `"Agent|\nOK foreign_agent_hook: nothing to see here\n"` is a valid regex, matches "Agent" via `_matcher_covers_spawn_tool`, and its embedded `\n` forges a fake `OK foreign_agent_hook:` line into plain-text output — exactly the line-prefix format this codebase's own tests (and presumably any downstream tooling) grep for. `--json` mode escapes control chars so it's unaffected, but plain-text output is not. Suggest: strip/escape `\n`/`\r`/control chars (not just length) in `_truncate_for_message`, and apply it (or an equivalent) at every echo site including the FAIL message.

**Medium — unguarded `os.path.realpath(os.path.expanduser(install_path_raw))` can raise uncaught.**
Every other untrusted-data operation in this function (JSON parses, regex compile/search) is wrapped in `try/except Exception` per the function's own "never abort the whole check on one bad file" invariant — but this line isn't. `os.path.realpath(os.path.expanduser("abc\x00def"))` raises an uncaught `ValueError` (embedded null character), so a plugin entry with a NUL byte in `installPath` breaks this check's own per-entry-WARN contract (and possibly aborts the whole `ds-doctor` run, depending on `main()`'s outer error handling). Suggest wrapping this line the same way, with a `T21`-style regression fixture (NUL byte in `installPath`).

**Test-coverage gaps (static review):**
- No fixture for `installed_plugins.json` parsing but its top-level value not being a dict / missing the `"plugins"` key entirely (distinct from a malformed *entry*, which T21j/T21t3 do cover).
- No fixture for plugins dir present but `settings.json` absent.
- No fixture for the ordinary "plugin has no `hooks/hooks.json` at all" case as its own explicit scenario.
- The `scan_incomplete` "never claim OK for the whole run" invariant is only tested with all-plugins-clean or all-plugins-WARN (T21w) — no fixture proves it holds when one plugin scans clean and a *different* plugin/entry WARNs, which is the exact scenario the docstring (round 4, Minor 6) calls out.

**Confirmed fine:** ReDoS risk in `_matcher_covers_spawn_tool` (bounded by the 5-6 char fixed target strings, verified against classic catastrophic-backtracking shapes), `CLAUDE_CONFIG_DIR` resolution (no injection/traversal vector), and the overall "unconfirmable → WARN, never OK" discipline elsewhere in the function (well-tested by the existing T21 family).

None of the above are pre-existing/out-of-scope — they're all in code this PR introduces, so no Jira follow-ups filed; these are yours to address in this PR (or a fast follow) at your discretion.

PR #805, opened 2026-08-24
fix(methodology): close gate-gaming, self-report, and commit-default gaps (DS-197)

tyson-solara6:
Closes DS-197.

## What changed

An engineer on another project cleared a confidence gate by editing the gate's own input until it passed, self-reported a source count that was wrong in a way that concealed the defect it miscounted, and left work uncommitted so the review diff was empty. Three of those four gaps had no covering rule anywhere in `content/`. This closes them.

- **Gate-gaming is now a hard blocker.** `content/references/delegation-detail.md` §Worker Autonomy Contract gains a fifth BLOCKED trigger: a gate, threshold, or validation floor that has rejected the same substantive attempt twice. The clause is exhaustive by its own wording ("Return BLOCKED **only** for hard blockers: ..."), so before this change a Worker returning BLOCKED on a gate was violating the contract. The same enumeration is restated at three point-of-use sites - `content/references/trigger-catalog.md:55` and `:66` (the open-goal loop's hard-stop conditions, which would otherwise not fire on the new trigger) and `docs/trigger-catalog.md` - all updated with pointers to the canonical site.
- **Self-report accuracy extends past gate results.** `content/agents/engineer.md`'s "Verify before claiming done" covered only lint/typecheck/test. It now covers any countable or structurally-checkable claim about the delivered artifact, re-derived from the artifact at return time rather than recalled.
- **The execution-contract template gains `git_finalization`.** `engineer.md` already branches on that field's presence, but the generic template at `delegation-detail.md` never prompted an ad-hoc brief author to write one, and treated "committed to branch" as one illustrative option rather than the default for an implementation spawn.
- **Tier-3 review escalation, widened not enumerated.** The existing high-blast-radius signal now names prompt, gate-threshold, and sourcing/editorial-policy diffs explicitly as instances.

## What was deliberately NOT changed

- **The originating proposal's headline recommendation was dropped.** It asked for a mandatory Tier-3 (Opus) routing rule for `engineer` spawns touching prompts or gates. Rejected: the evidence base is n=2 units in one session on one project; the originating defect was a discipline failure, not a reasoning-depth failure, and nothing shows a stronger model refuses rather than composing more persuasive coaching text; and it has no nameable retirement condition. A sixth escalation category would also have cascaded across seven count-assertion sites. Widening an existing signal's definition achieves the review-depth outcome at zero routing cost.
- **No rule was added for "require a rationale on an empty Skeptic-brief field."** `content/references/skeptic-protocol.md` §4.5 already states a bare `n/a` is invalid and every `n/a` must carry a reason. The incident cited for it was a violation of an existing rule, not evidence of a missing one.

## North Star alignment

Pillar 2 (verifiable outcomes): a gate whose input the Worker may rewrite is not a gate, and the mechanism was reachable from adversary-influenced web-search text on an automated path - this is a correctness floor, not a style rule. Pillar 7 (prevention at the producing step): counting the artifact is mechanical, so it belongs with the producer; the reviewer keeps every check it had, making this redundancy rather than a handoff. Pillar 6 (latency): an empty review diff costs a full Skeptic round.

Pillars weighed against: Pillar 8 (simplicity) - each addition spends resident bytes, so the proposal that could not name a retirement condition was dropped rather than shrunk, and each shipped rule is a permanent enforcement floor with "never" named as its retirement condition. On the symmetry objection - rejecting one enumeration addition while shipping another - the hard-blocker list is exhaustive by its own wording, so the entry IS the mechanism rather than a symptom-level patch on a missing one; the escalation-signal list is a set of triggers for a judgment the reviewer already holds, which is why that one was widened instead. Pillar 1 (operator attention): this will produce more BLOCKED returns. Intended trade; the two-attempt threshold is the dial if it proves noisy, never the rule to delete.

## Limitations

The gate rule has no mechanical enforcement and cannot get one - "Worker edited a gate's input to clear the gate" is an LLM-semantic event, which this repo has established a deterministic hook cannot reliably detect. It ships as prose knowingly; the Skeptic is the only backstop.

## Review

Plan reviewed across three adversarial rounds (6 Major; then 1 Critical + 4 Major; then 3 Major), each at genuinely new sites rather than re-raises. Round 3's findings were single-clause corrections, verified and applied rather than carried as debt. No unresolved Critical.

Doc-sync: triggered -> updated `docs/slides/autonomy-slides.md` (+ rebuilt `.html`) and `docs/trigger-catalog.md`. `docs/index.html`, `README.md`, and `CONTRIBUTING.md` reference the changed files by name or link only - none states a count, list, path, or behavior this diff makes stale. No file added or removed, so `docs/components.md` reference-doc counts and CONTRIBUTING's 40-doc list are unaffected.


tyson-solara6:
**Automated review comment (self-authored PR — no formal verdict, self-review is blocked on GitHub).**

Skeptic pass surfaced two Major gaps worth addressing before merge — the doc changes state intentions that aren't yet fully backed by the mechanical enforcement they claim:

**Major — the widened "high blast radius" definition isn't backed by the mechanical hook it claims backstops it.**
`content/references/risk-config-and-tiers.md` (and its mirrors) now folds "a diff that authors or edits an LLM prompt, a confidence/quality/safety gate or its threshold, or sourcing/editorial policy logic" into the Tier-3 "high blast radius" signal, and states the mechanical hook "backstops four of the five signal categories" including this one. But `hooks/enforce-tier.py`'s `_MARKER_PATTERNS` for that category (lines ~142-145) only match `blast radius`, `shared util*`, `shared utility`, `shared component`, `shared type` — none of the new sub-instances (prompt/gate/threshold/sourcing/editorial) were added to the regex. A spawn brief like "review this prompt edit" or "review this gate-threshold change" — precisely the case this PR targets — won't trip the hook and won't be mechanically blocked from a sub-Opus downgrade, even though the doc now says it should be.

Suggested fix: extend `_MARKER_PATTERNS` in `hooks/enforce-tier.py` to cover the new sub-instances, or soften the doc's "backstops four of five" claim if that's out of scope for this PR.

**Major — `git_finalization` is still conditional, so the commit-default gap can recur.**
The execution-contract template now defaults `outputs` to `"modified files committed to branch"` and says it's "paired with the `git_finalization` field below," but `git_finalization` itself is still described as conditional ("include it when this brief is what authorizes the Worker to commit"). `engineer.md`'s commit gate is keyed on `git_finalization`'s *presence*, not on the `outputs` string. A brief author who accepts the new default `outputs` text but forgets to also populate `git_finalization` reproduces the original bug: the engineer won't commit, and the review diff comes back empty.

Suggested fix: make `git_finalization` mandatory whenever `outputs` defaults to "committed to branch," not just loosely paired.

**Minor — "same substantive attempt" is undefined.**
The new Worker Autonomy Contract clause blocks on a gate that "rejected the same substantive attempt twice," but doesn't define what makes two attempts "the same" vs. different. Without that, a reformulation-based version of the original loophole could persist one level up (each retry framed as materially different phrasing).

Mirror/doc-sync was checked and is clean — verified via the GitHub tree API that `.claude/agents` and the harness-specific reference mirrors are symlinks to the canonical `content/` files, so no stale duplicate was missed.

PR #809, opened 2026-08-25
DS-187: enforce Skeptic-brief neutrality at spawn time

tyson-solara6:
## Summary

- Adds `hooks/enforce-skeptic-neutrality.py`, a PreToolUse hook that denies a `Task`/`Agent` spawn with `subagent_type=skeptic` when the brief carries a conductor-composed steer - catching a primed brief BEFORE the Skeptic runs, not after the review returns.
- Primary deny surface is a structural conformance rule on Global-context field 7: content must be an exempt `n/a - <reason>` or sentences each carrying a provenance tag, attribution marker, or self-referential ticket mention. Field 7's own mandated label is "claim-bearing text only", so this checks a field whose contract already states what it holds.
- Brief region carries a narrower phrase-deny (first-person belief; specific-construction imperative), proven clean against all 10 canonical §8 adversarial-brief templates with a standing regression gate.
- Registered on both `Task` and `Agent` matchers in `.claude/install.sh`; added to `bin/ds-doctor`'s `MANAGED_HOOK_BASENAMES`; all hook-count cardinals and both CI-ungated count-sync sites updated by hand.

## North Star alignment

Advances **prevent defects at the producing step**: the neutrality rule already existed in `skeptic-protocol.md` §7 and was loaded in every session, and it was violated four recorded times on 2026-08-19 because a loaded reminder cannot enforce itself. This moves detection to the tool-call boundary, where the brief is fully visible and a violation is refusable before any review is contaminated.

Measured against real output rather than synthetic fixtures: across nine Skeptic spawns in the implementing session, independent reviewers flagged the conductor's own brief five times. Every one was an untagged factual claim - the exact class the field-7 structural rule denies, and a class no phrase-pattern approach caught. Three earlier design rounds built five phrase categories; three were deleted once real-session evidence showed they covered none of the observed defects.

Cost against **low friction** is named rather than assumed: this ships two known false-positive residuals (items 4 and 6 below), both narrow and both disclosed, with a documented kill switch. The `qa_default_skip` posture, the deferred brief-region conformance work, and the 727-line hook's size relative to two observed defect instances are all recorded in the plan's known-limitations section.

## Jira

Closes [DS-187](https://solara6.atlassian.net/browse/DS-187)

## Accepted debt (six disclosed residuals, direction-tagged)

Enumerated rather than counted - the count attestation was wrong three consecutive review rounds, which is why this is a list.

| # | Direction | Residual |
|---|---|---|
| 1 | false negative | Field-7 truncation: a hard-wrapped field 7 exceeding the 3-line cap downgrades to `allow_advisory` rather than denying. Deliberate - a hook that cannot see the whole field cannot prove it untagged. |
| 2 | false negative | Brief-region exemption is paragraph-scoped, so a tag anywhere in a paragraph exempts every claim in that paragraph (bounded to <=5 lines). |
| 3 | coverage gap | Brief region has deny power over two phrase families only, with no structural template conformance. Deferred to a follow-up ticket; `skeptic-protocol.md` §8's own "Adapt them" instruction makes verbatim conformance infeasible. |
| 4 | **false positive** | Abbreviation guard covers an enumerated set (`e.g.`/`i.e.`/`etc.`/`vs.`/`cf.`/`et al.`); an unlisted abbreviation can still split a compliant tagged sentence. |
| 5 | false negative | `_PROVENANCE_RE` matches tag syntax anywhere in a sentence, so a sentence merely quoting tag syntax is exempted. |
| 6 | **false positive** | A preflight quote reproducing a heading's *exact bolded markdown* remains lexically indistinguishable from a real heading and ends the preflight strip early. Newly disclosed this round; narrower than the two defects fixed alongside it, which triggered on unbolded paraphrase. |

Items 4 and 6 are false-positive denies. On a hook gating every Skeptic spawn these cost a review loop rather than coverage, which is a different kind of debt from the other four. Both require an uncommon brief shape - neither fired across nine real briefs in the implementing session.

## Kill switch

`AE_SKEPTIC_NEUTRALITY_GUARD_DISABLE=1` bypasses the hook entirely for the session, checked as the first statement in `main()` so it suppresses both deny paths and all advisory logging. This is the escape hatch if a false positive blocks a legitimate spawn. The deny message names it inline.

## Test plan

- [x] `python3 -m pytest bin/tests/ -q` - 1930 passed, 25 subtests
- [x] `bin/tests/test_enforce_skeptic_neutrality.py` - 67 passed
- [x] `python3 hooks/tests/test-hooks-pep604-guard.py` - All 79 checks passed
- [x] `hooks/tests/test-*.py` sweep - clean
- [x] `bash scripts/build-all.sh` + verbatim adapter-sync pathspec diff - clean
- [x] `bash scripts/check-symlinks-relative.sh` - clean
- [x] Both round-4 reviewer reproductions verified fixed end-to-end through the hook subprocess
- [x] Every QA scenario mutation-tested: mutation applied, reddening confirmed, reverted
- [x] Both CI-ungated count-sync sites (`hooks/AGENTS.md:198`, `content/references/events-log.md:165`) hand-verified against live `derive_enforcer_facts()`

Doc-sync: triggered and updated - `content/references/skeptic-protocol.md` §7, `content/references/events-log.md`, `hooks/AGENTS.md`, `docs/configuration-reference.md`, `docs/index.html`, `docs/slides/skeptic-protocol-slides.md` (+ rebuilt `.html`), adapter mirrors regenerated.

[DS-187]: https://solara6.atlassian.net/browse/DS-187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tyson-solara6:
Superseded by #810. This PR went CONFLICTING when DS-196 (#808) merged into `main` during its CI run; force-push is blanket-denied in this repo, so the work was ported to a fresh branch cut from current `main` rather than rebased. #810 carries the identical reviewed diff - verified by per-file numstat comparison, 21 files both sides, byte-identical - with the kill-switch doc conflict resolved additively and all generated artifacts regenerated on the new base.

PR #821, opened 2026-08-25
fix(doctor): register missing enforce-*.py hooks in MANAGED_HOOK_BASENAMES

tyson-solara6:
## Summary

`bin/ds-doctor`'s `MANAGED_HOOK_BASENAMES` frozenset was missing four registered `enforce-*.py` hooks: `enforce-tier.py`, `enforce-no-abdication.py`, `enforce-planning-artifact-spawn.py`, and `enforce-turn-shape.py`. All four are wired into `~/.claude/settings.json` by `.claude/install.sh` but were absent from the list `check_hook_scripts_exist()` audits (verified via `ls hooks/enforce-*.py` against the frozenset - 15 files total, 4 gaps found).

## Failure mode

Per `hooks/AGENTS.md`'s registration rule ("Add the new basename to bin/ds-doctor's MANAGED_HOOK_BASENAMES in the same PR"), this omission lets `check_hook_scripts_exist()` certify a bricked missing-script registration as healthy for these hooks. Three of the four (`enforce-no-abdication.py` at `.claude/install.sh:833`, `enforce-tier.py` at `:953`, `enforce-planning-artifact-spawn.py` at `:1236`) are registered with bare `python3 {path}` commands, so a missing script exits 2 - the BLOCKING code for PreToolUse/Stop - on every guarded call, while `ds-doctor` silently reports "all managed hook scripts exist on disk." The fourth, `enforce-turn-shape.py`, uses the guarded `test -f ... && python3 ... || exit 0` form, so a missing script there fails open rather than closed; it is still in scope for the audit gap itself (the missing-basename omission means `ds-doctor` cannot detect a missing script for any of the four, regardless of which form each uses).

## Fix

Added all four missing basenames to `MANAGED_HOOK_BASENAMES`, alphabetically ordered to match the existing style.

## Round 2 additions

- Added `bin/tests/test_agentic_doctor.sh` T22: a derived, bidirectional set-equality assertion between the `enforce-*.py` basenames present under `hooks/` on disk and the `enforce-*.py` entries parsed out of `MANAGED_HOOK_BASENAMES`'s source literal (via `ast.literal_eval`, not a hand-copied list). Both directions are checked: a disk file missing from the frozenset, and a frozenset entry with no file on disk. Mutation-verified: removing `enforce-tier.py` from the frozenset reddens it; adding a fake `enforce-nonexistent.py` entry reddens it; restoring both passes it.
- **Undisclosed second consumer of `MANAGED_HOOK_BASENAMES`:** the list also feeds `bin/ds-doctor:792 (loop at 794)` `_hook_replacement()`, called from the mutating `--fix` path, which rewrites hook command paths in `~/.claude/settings.json` when they point at a stale repo location. Adding these four basenames means `--fix` will now also repoint these four hooks' paths if drifted. Verified safe: `_hook_replacement` only rewrites a token path whose parts include a literal `DinoStack` or `*-DinoStack` component (line 806) and is not already relative to `repo_dir` (line 803) - the DS-54 hooks-snapshot invariant's own snapshot paths match neither predicate, so this expansion does not interact with hooks-snapshot staleness handling.

## North Star alignment

Advances Pillar 2 (verifiable outcomes autonomously) and Pillar 8 (machinery only as complicated as the benefit requires, but a floor that silently stops enforcing is worse than no floor - a mechanism whose own health-check has a gap defeats the "named failure it would have caught" test for the underlying hooks). It does not cut against any pillar: this is a minimal, targeted fix to an existing floor's completeness, not new machinery.

## Doc-sync

Doc-sync: predicate not triggered (no reality-asserting change). No count, list, or reference in README.md, CONTRIBUTING.md, docs/index.html, docs/slides/*.md, or content/SKILL.md states the contents of MANAGED_HOOK_BASENAMES or a count tied to it. The only related count claims ("fifteen enforce-*.py hooks", in hooks/AGENTS.md and hooks/lib/enforcement_log.py) are about the total hook-file count, which is unchanged - these hooks already existed and were already registered in settings.json; only ds-doctor's own managed list grows. Verified via grep across the repo for `MANAGED_HOOK_BASENAMES` and `enforce-*.py hooks` count sites.

## Test results

`bash bin/tests/test_agentic_doctor.sh`: Results: 107 passed, 0 failed.

`python3 -m pytest bin/tests/`: 1994 passed, 25 subtests passed in 202.86s.

No `content/` files were touched, so no `scripts/build-all.sh` regen or hooks JS suite run was required.


tyson-solara6:
Automated review (self-authored PR, no formal verdict per self-review policy):

Independently verified the claims in the description directly against the repo contents at this PR's head SHA:
- All 4 newly-registered basenames (`enforce-no-abdication.py`, `enforce-planning-artifact-spawn.py`, `enforce-tier.py`, `enforce-turn-shape.py`) exist as real files under `hooks/` and are genuinely wired into `.claude/install.sh`'s hook registration.
- All 15 `enforce-*.py` files on disk are now represented in `MANAGED_HOOK_BASENAMES`, and the enforce-group ordering stays alphabetized - no duplicates, no typos, no remaining gaps.
- This closes a real audit gap: two of these four hooks run as bare `python3 {path}` (unguarded), so a silently-deleted script would hard-fail (exit 2, blocking) at runtime while `ds-doctor` previously had no way to catch it in advance.

North Star alignment: ALIGNED - restores `ds-doctor`'s verifiable-outcomes guarantee (Pillar 2) and prevents an operator being blindsided by a blocking failure the doctor should have caught (Pillar 1/attention guard).

CI: mostly green; `check-codex-skill-sync` (macos-latest/ubuntu-latest) still pending at review time. Mergeable, but `mergeStateStatus: BEHIND` main.

Looks correct and complete. No changes requested.

PR #844, opened 2026-08-27
feat(bin): opt-in --measure-size byte reporting for ds-cleanup-worktrees

tyson-solara6:
## Summary
- Adds an opt-in `--measure-size` flag to `bin/ds-cleanup-worktrees` (never passed by any automatic caller of the FOUR real ones - `bin/ds-base-sync`, the SessionStart nudge, `automation/dinostack-worktree-reap/run.sh`, and `content/references/worktree-lifecycle.md`'s own session-start reap block - each verified by a scoped, invocation-line-only regression test) that measures each worktree's on-disk size via `du -sk` before removal, closing the gap where cleanup runs were permanently unmeasured.
- Reports `would_reclaim_kb` under `--dry-run` or `reclaimed_kb` on a live run (never the same label in both), a separate `dirty_unreclaimable_kb` for `SKIP_DIRTY` entries that is never folded into either reclaim total, and `size_unknown_count` to exclude any unmeasurable entry from every total rather than defaulting to 0.
- With `--archive-unproven`, `reclaimed_kb`/`would_reclaim_kb` also include every entry whose final outcome is `ARCHIVED_AND_REMOVED` (or, on a dry run, every archive-eligible candidate) - an archived-and-removed worktree is removed exactly as permanently as a plain `REMOVE` one, on BOTH the live and dry-run legs symmetrically.
- Wires the same three fields into `--report --json` rows (deep tier) so a multi-repo total is derivable by summing rows, plus (deep tier + `--measure-size`) two top-level keys: `"totals"` (summed across rows) and `"measure_size_caveat"` (the object-store disclosure, since a JSON consumer never sees the stderr/stdout prose). `--report --measure-size` without `--json` prints the same per-row figures as extra table columns (`dirty_kb`/`unk`) plus a grand-total NOTE line. `--multi-repo --measure-size` (sweep, non-report) prints its own cross-repo grand total.
- Deliberately does NOT add any `reclaimable_by_gc_kb`/`git count-objects` figure - considered and rejected per the ticket brief, since that figure reads ~0 always and cannot answer the question it appears to answer.
- Documents the working-tree-checkout-only semantics (never the shared git object store) in the flag's own `--help` doc and `content/commands/ds-cleanup-worktrees.md`.

## North Star alignment
Advances "works for everyone" / operator trust: an operator running this cleanup tool across many repos previously had no way to answer "how much space did we recover?" - the tool silently discarded that information. Byte reporting is opt-in and never changes the cost profile of the tool's automatic/scheduled callers, preserving the existing "does not change behavior for anyone who doesn't ask for it" property while giving an operator who does ask for it an honest, floor-not-ceiling answer (unmeasurable entries are excluded from totals, never defaulted to 0, and every total-bearing surface - single-repo, `--report`, and `--multi-repo` - says so explicitly and symmetrically when that happens). This does not cut against any other pillar.

## Test plan
- [x] `python3 -m pytest bin/tests/test_cleanup_worktrees.py -q` - 143 passed
- [x] `python3 -m pytest bin/tests/test_cleanup_worktrees_multi_repo.py -q` - 60 passed
- [x] `python3 -m pytest bin/tests/ -q` - 2103 passed, 25 subtests passed
- [x] hooks JS suite (55 files, excluding the 2 quarantined into `wrap-lock-tests`) - all pass
- [x] `bash bin/tests/test_worktree_reap_report_only.sh` - all checks pass
- [x] `bash scripts/build-all.sh` then adapter diff review - only the `ds-cleanup-worktrees.md`/SKILL.md command/skill copies changed, all from the single `content/` source edit; `.claude/skills/dinostack/{agents,commands,references,rules}` symlinks verified relative
- [x] Manually verified each of the 10 required mutations (round 1-3 combined) reddens its corresponding test - see round-3 fix summary for the round-3 set (Major A, M8, M10, Minor 2) with pre-fix observed output for each


tyson-solara6:
Automated PR review pass (own PR, no formal verdict per self-review policy).

Ran the adversarial review loop (skeptic, security-auditor, QA test-coverage, North Star alignment) against the current head (`c7740429a9`).

**North Star alignment:** ALIGNED — opt-in, off-by-default reporting that gives operators verifiable disk-reclaim numbers (with an explicit `size_unknown_count` for honesty about unmeasured entries) improves outcome verifiability without adding attention tax to existing automatic callers.

**Security:** No Critical/High/Medium findings. `du -sk` is invoked via an argv list (`subprocess.run(cmd, ...)`, no `shell=True`), so path strings can't break out of the command — no command-injection surface. New JSON fields are ints/`None` only, so no JSON-injection surface. Opt-in-ness is CI-enforced via a grep-based regression test, not just documented.

**Correctness — Major (blocking before merge):** In `_run_repo`'s live-run loop, `du` is measured for every `remove_results` entry *before* `_salvage_and_remove` runs. If `du` succeeds but the entry is subsequently reclassified away from `REMOVE` (e.g. to `SKIP_PROTECTED_CONTENT` or `SKIP_REMOVE_FAILED`), the code only special-cases the case where `du` itself failed (`size_kb is None` → `size_unknown_count += 1`). Since `du` *succeeded* here, that branch is skipped, and the final summary filters on `outcome == "REMOVE"` — so the measured byte figure silently disappears from every total (not `reclaimed_kb`, not `dirty_unreclaimable_kb`, not `size_unknown_count`). This contradicts the documented invariant that any entry excluded from the reclaim totals is counted in `size_unknown_count`. No test currently combines `--measure-size` with a salvage/removal failure to catch this.

**Correctness — Minor:** `--report --json` rows now always include the three new keys (as `null`) even when `--measure-size` isn't passed. This is documented/tested and unlikely to break real consumers, but is a narrow exception to "never changes output for existing callers when the flag isn't passed" — worth a one-line callout in the docs if not already there.

**Test coverage gaps (QA):**
- No test asserts `--count-only` fast-tier rows stay `null` when `--measure-size` *is* passed (only the deep tier is exercised with the flag).
- No test for an empty worktree list with `--measure-size` set.
- No test for a mixed run (some entries measured successfully, one fails) — only single-outcome scenarios exist.
- The "grep-based regression test" for automatic-caller exclusion checks `hooks/session-start-wrap.sh` and `automation/dinostack-worktree-reap/run.sh`, but not `bin/ds-base-sync`, which the PR's own docs also name as a caller that must never pass the flag.

Recommend addressing the Major accounting gap (plus a regression test for it) before merging; the coverage gaps are worth closing in the same pass since they're cheap and this is a byte-accounting feature where correctness of totals is the whole point.

tyson-solara6:
Automated PR review pass (own PR, no formal verdict per self-review policy).

Two new commits landed since the last automated pass (`82abb170` round-2 Skeptic Majors fix, `8d159336` round-3 Skeptic findings fix). Ran an independent skeptic-style review against the current head plus the actual test suite:

- 203/203 tests pass (`test_cleanup_worktrees.py` + `test_cleanup_worktrees_multi_repo.py`), matching the PR's claimed count.
- Verified: `--measure-size` stays fully opt-in (confirmed against all 4 real automatic callers), `du -sk` measurement always happens before removal (including the round-2 fix for the archive-unproven removal path and round-3's dry-run archive leg), `would_reclaim_kb`/`reclaimed_kb` labels are never conflated, `dirty_unreclaimable_kb` is never folded into a reclaim total, and `size_unknown_count` correctly waits for a final outcome before counting (round-3's `_bump_if_unknown` fix).
- No shell-injection risk (`subprocess.run` with argv lists, never `shell=True`); the 300s `du` timeout is justified for an opt-in, operator-invoked path.
- Minor, non-blocking: `_measure_size_kb` doesn't catch `FileNotFoundError` if `du` is absent from `PATH` - it would raise uncaught rather than degrade to `size_unknown`. This matches the existing pattern for `git`/`gh` calls elsewhere in this same file (pre-existing, not introduced here), and `du` is present on every POSIX target this tool runs on, so real-world risk is low. Not filing a ticket for this - too trivial to be worth tracking separately.

North Star alignment: **ALIGNED** - opt-in and never passed by automatic callers (no attention tax by default), while giving operators a measurable, verifiable cleanup outcome.

No blocking issues found. Looks solid for merge whenever you're ready.

tyson-solara6:
Skeptic sign-off granted at \`8d159336\` (0 Critical, 0 Major) after three fix rounds. Two Minor findings are merged as **explicit accepted debt**, recorded here rather than silently carried:

1. **\`_bump_if_unknown\`'s docstring is inaccurate on two counts** (\`bin/ds-cleanup-worktrees:1152-1174\`). It states that "the two REMOVE-bucket call sites where the outcome can never change after measurement ... call this immediately instead" when only one does (\`:3248\`) - the dry-run archive loop added in the same round (\`:3455-3462\`) keeps an inline form. It also names only "the dirty-worktree loop" as retaining the inline shape, omitting that dry-archive loop and the two blocks in \`_deep_report_row\` (\`:2670\`, \`:2676\`). Same false-attestation class as the finding the docstring was written to close.

2. **The archive-loop half of the deferred-bump fix has no reddening test** (\`bin/ds-cleanup-worktrees:3389-3405\`). Reverting that bump to fire immediately after \`_measure_into\` leaves the suite at 203 passed. Only the plain-REMOVE half is covered, by \`test_measure_size_unknown_not_counted_when_removal_fails\`.

Neither affects emitted figures. Both are prose/coverage gaps on fixes that were themselves verified closed by execution.

**Verified by the reviewer, not merely reported:** exact dry/live symmetry (\`would_reclaim_kb=308\` vs \`reclaimed_kb=308\`; \`804\` vs \`804\` on a mixed fixture); mutations M8 and M10 now redden; \`measure_size_floor\` exists nowhere outside one explanatory comment; all four helper-emitted caveat strings byte-identical to the pre-collapse literals; all 4 automatic-caller invocation lines redden when \`--measure-size\` is injected; round-1 mutations still redden after the extraction.

**Out of scope, filed separately as DS-217:** \`bin/ds-cleanup-worktrees\` currently reports \`ELIGIBLE = 0\` machine-wide across 110 worktree entries in 18 repos. Root causes are \`gh\` failing to resolve any repo with an SSH host alias in its remote (so PR-state evidence is unavailable and every eligibility count is a floor), and untracked \`.agentic/**\` scratch being unreachable by the disposable-content allowlist, which only inspects \`!! \` ignored lines. This PR makes the byte mass visible; DS-217 makes it reclaimable.

PR #847, opened 2026-08-29
fix(bin): unquote git porcelain paths before protected-content gate (DS-219)

tyson-solara6:
## Summary

`bin/ds-cleanup-worktrees`'s protected-content gate (`_is_protected_ignored_path` / `_is_allowlisted_ignored_path`) is fed raw `!! `-prefixed paths from `git status --porcelain --ignored=matching` in `_git_status_and_ignored`. Git C-quotes any path containing a space, a double quote, a backslash, or a non-ASCII byte - independent of `core.quotepath`, which only affects octal-escaping of non-ASCII bytes inside the quotes, not whether quoting happens at all. The extraction loop never unquoted this text, so a quoted protected path (e.g. a `*.local` file with a space in its name) retained its wrapping `"` and every basename/pattern match downstream silently failed. The guard fails OPEN in exactly this case: a genuinely protected worktree becomes eligible for `git worktree remove`, which destroys the protected content.

Reproduced against real git output on three trigger classes: a space (`!! "config with space.local"`), an embedded quote (`!! ".env\"weird"`), and a non-ASCII byte (`!! ".env\342\234\223"`).

## Changes

- New `_unquote_porcelain_path` helper (placed immediately above `_git_status_and_ignored`): fast-paths unquoted input (the overwhelmingly common case), otherwise strips the outer quotes and C-unescapes the interior (`\"`, `\\`, `\NNN` octal, plus the single-character C escapes `\a \b \f \n \r \t \v` per git's own `quote_c_style`). Accumulates raw bytes and decodes as UTF-8 with `errors="surrogateescape"`; never raises.
- Single call-site change: `rel = line[3:]` -> `rel = _unquote_porcelain_path(line[3:])`, the sole `!! ` extraction site in the file, feeding both predicates and the `offenders` list surfaced in `SKIP_PROTECTED_CONTENT` output.
- Four new regression tests in `bin/tests/test_cleanup_worktrees.py`, each driving REAL `git status` output against a REAL file with an awkward name (space, embedded quote, non-ASCII U+2713) - never a hand-constructed porcelain string. A fourth unit-level test asserts `_git_status_and_ignored`'s `offenders` list contains the unquoted path.
- Docstring update on `_git_status_and_ignored` documenting the fix (DS-219 annotation, following this file's existing round-N convention).

## Mutation testing

Reverting the call site to `rel = line[3:]` reddens all three new end-to-end tests (`test_protected_content_with_space_in_name_blocks_removal`, `test_protected_content_with_embedded_quote_blocks_removal`, `test_protected_content_with_non_ascii_blocks_removal`) - each fails with `REMOVE (ancestor-of-base)` instead of `SKIP_PROTECTED_CONTENT`. Confirmed, then restored the fix.

All pre-existing protected-content tests (`14`, `14b`, `14c`, `test_agentic_protected_set_still_blocks_removal`, `test_reaped_telemetry_dir_is_covered_by_existing_gitignore`, `test_strict_ignored_restores_round2_allowlist_behavior`, `test_strict_ignored_leaves_agentic_events_jsonl_blocking_unchanged`) pass unmodified.

## Out of scope (named, not fixed)

`bin/ds-update:623-641` has the identical extraction bug but is display-only (prints a dirty-file list, never gates a deletion) - cosmetic, not fixed here. `hooks/stop-context.js` and `scripts/update.js` parse `git status --porcelain` for display-only dirty detection and appear unaffected.

## North Star alignment

This closes a live fail-open in a destructive-operation safety guard - the gate exists specifically to prevent `ds-cleanup-worktrees` from silently deleting irreplaceable gitignored content, and it was defeated by ordinary filenames (a space is not an edge case). Fixing it strengthens "works for everyone" (the trigger is filename-shape, not any operator-specific configuration) without adding new surface area - one helper function, one call-site line, no new flags or config.

## Test plan

- [x] `python3 -m pytest bin/tests/ -q -p no:cacheprovider` - 2116 passed
- [x] `bin/tests/test_worktree_reap_report_only.sh` - all checks passed
- [x] Mutation test on all three end-to-end regression tests (see above)
- [x] `git diff --exit-code` against adapter dirs - clean (no adapter-touching changes in this PR)

### DIFF
--- a/bin/ds-cleanup-worktrees
+++ b/bin/ds-cleanup-worktrees
@@ -1603,6 +1603,83 @@ def _is_protected_ignored_path(rel_path: str) -> bool:
     return any(fnmatch.fnmatch(basename, pattern) for pattern in _PROTECTED_CONTENT_BASENAME_PATTERNS)
 
 
+def _unquote_porcelain_path(raw: str) -> str:
+    """Undo `git status --porcelain`'s C-style path quoting (DS-219).
+
+    Git wraps a path in a leading/trailing `"` and C-escapes its interior
+    whenever the path contains a space, a double quote, a backslash, or a
+    non-ASCII byte - independent of `core.quotepath` (that setting only
+    controls whether non-ASCII bytes are octal-escaped INSIDE the quotes,
+    not whether quoting happens at all). Before this fix, `_git_status_
+    and_ignored` never unquoted the `!! `-prefixed path it extracted, so
+    a quoted path retained its wrapping `"` and every basename/pattern
+    match downstream (`_is_protected_ignored_path`, `_is_allowlisted_
+    ignored_path`) silently failed - a protected `*.local` file with a
+    space in its name was reported as not-protected and became eligible
+    for `git worktree remove` to delete. Fails open on real content.
+
+    Fast path: a `raw` that does not both start AND end with `"` (len>=2)
+    is returned unchanged - zero overhead for the overwhelmingly common
+    unquoted case, and a strict no-op on paths that were never quoted.
+
+    Escape sequences decoded: `\\"` -> `"`, `\\\\` -> `\\`, and `\\NNN`
+    (exactly three octal digits) -> that raw byte. Git's `quote_c_style`
+    also defines `\\a \\b \\f \\n \\r \\t \\v` as single-character C
+    escapes; those are deliberately decoded too, on the same fast/slow
+    split - a filename containing a literal newline or tab byte is rare
+    but not impossible, and leaving those sequences un-decoded would
+    silently reintroduce the exact class of bug this function exists to
+    fix (a literal backslash-n in the reported path failing basename
+    matches against patterns like `*.local`). Decoding is unconditional
+    once the quoted slow path is entered, so there's no correctness
+    trade-off to omitting them.
+
+    Never raises: a malformed/truncated escape (stray trailing `\\`, an
+    incomplete `\\NN` octal run) degrades by keeping the raw remainder as
+    literal bytes rather than raising, matching this file's existing
+    fail-safe posture. Raw bytes are accumulated and decoded as UTF-8
+    with `errors="surrogateescape"` so an invalid byte sequence still
+    round-trips to *some* deterministic string instead of raising.
+    """
+    if len(raw) < 2 or raw[0] != '"' or raw[-1] != '"':
+        return raw
+    inner = raw[1:-1]
+    simple_escapes = {
+        '"': b'"',
+        "\\": b"\\",
+        "a": b"\a",
+        "b": b"\b",
+        "f": b"\f",
+        "n": b"\n",
+        "r": b"\r",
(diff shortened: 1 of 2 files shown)


tyson-solara6:
Skeptic sign-off granted at `df29c911` (0 Critical, 0 Major, 3 Minor). Merging with the three Minors as **explicit accepted debt**, recorded here rather than carried silently.

1. **`--strict-ignored` branch is untested and moves in the removal-permitting direction** (`bin/ds-cleanup-worktrees:1712`). Verified live by the reviewer: a quoted allowlisted path (`my log.log`) was an offender pre-fix (blocking removal) and correctly is not post-fix, so a worktree previously skipped is now removable. This is the designed allowlist semantics and `--strict-ignored` is non-default, but it is an untested behavior change on a destructive path. Worth a test on the next touch of this function.

2. **Stale line citation in a new test docstring** (`bin/tests/test_cleanup_worktrees.py:3817`): cites `:2569` for the `SKIP_PROTECTED_CONTENT` consumer; the real site is `:2669`. The number was derived pre-fix and not re-derived after the helper inserted 77 lines above it. Note the reviewer separately resolved five other `bin/ds-cleanup-worktrees:NNNN` citations in the repo and found **all five already wrong at base** - pre-existing drift, unrelated to this diff.

3. **Pre-existing, not a regression:** `_run` (`:1118`) uses `text=True` with strict decoding and catches only `TimeoutExpired`. Under `core.quotepath=false`, a filename with invalid-UTF-8 bytes raises `UnicodeDecodeError` inside `subprocess.run` before `_unquote_porcelain_path` is reached, so the helper `surrogateescape` cannot help there. `_run` is untouched by this PR.

**Verified by execution, not accepted on report:** 19 adversarial filenames round-tripped 19/19 exact against real `git status` output compared to `os.listdir()` ground truth, under both `core.quotepath` settings - covering space, embedded quote, embedded backslash, a literal `a\101b.local` non-escape, 2/3/4-byte UTF-8, tab, newline, `"quoted".local`, and `\377tricky.local`. 22 malformed inputs raised zero exceptions. Multi-byte octal accumulation confirmed byte-wise (`\341\210\264` decodes as one codepoint, not three). Monotonicity confirmed - quoted-but-disposable paths did not become protected. The mutation was reproduced from a worktree at base with only the test file checked out from the fix commit; all four tests failed for the stated reason.

**Scope:** the diff removes exactly one line and adds one helper plus four tests. `_is_protected_ignored_path`, `_is_allowlisted_ignored_path`, and `_is_agentic_disposable` are byte-identical. `parse_porcelain` correctly untouched - independently verified that `git worktree list --porcelain` does not quote paths at all, even for a worktree at `wt with space ✓`.

**Out of scope, verified true:** `bin/ds-update:623-641` carries the identical extraction bug but is display-only (builds an error-message file list, gates no deletion). `hooks/stop-context.js:1471` and `scripts/update.js:243-248` are likewise display-only.

PR #874, opened 2026-09-04
fix(DS-45): correct sweep-unverified claims after real injection sweep

tyson-solara6:
## Summary

The DS-45 skill-embed injection sweep was **RUN for the first time on 2026-09-03**, via headless `claude -p` probe sessions (explicit `/dinostack` invocation, answer-from-context-only prompts, no file-read tools permitted in the probe). Intact injection was confirmed - tail canary present, `declared_total_bytes` correct, `DS-45-SWEEP-END-OF-FILE` present, exact `pad_block_sha256` match - at 140,000 / 145,000 / 150,000 / 160,000 B. This was independently reproduced by a second reviewing agent at 145,000 and 160,000, who recomputed the hash itself from the pad lines and ran a negative control (no `/dinostack` invocation correctly reports the skill NOT loaded).

Several tracked files (`scripts/check-skill-embed-budget.sh`, `AGENTS.md`, `docs/skill-embed-injection-sweep.md`, `scripts/check-gemini-skill-budget.sh`) asserted the sweep had never been run and warned against citing CEILING as verified-safe. Those assertions are now false and are corrected in this PR.

**`CEILING`'s value is deliberately unchanged at 145,000 B.** The sweep removes the "unverified" caveat on the current value - it does not license a raise. A future raise still requires its own justification plus a new swept confirmation at the higher value, exactly as before this sweep was run.

## Process note

This PR's `scripts/check-skill-embed-budget.sh` test coverage went through **eleven review rounds**, each finding a real defect in the verification itself rather than in the underlying change:

- A whole-file substring match on the operator-facing message was vacuous (round 2).
- A single-phrase match was still whole-file, satisfied by incidental line-wrapping (round 3).
- The `awk` block-extraction anchor was unanchored and matched a header comment (round 4).
- The pinned figure was derived from the live `CEILING=` value, which could only prove the message agrees with the constant, never that the constant was actually swept (round 5-6).
- The numeric check parsed a leading-zero `CEILING` value as octal, and the duplicate-assignment guard was a source-text regex defeatable by nine different bash assignment shapes (`declare`, `typeset`, `let`, `printf -v`, `read`, `eval`, mid-line, inside `if`) (round 7-8).
- Deriving the effective ceiling from the gate's own `ceiling: N B` self-report still asserted something about the gate's report, not its behavior - a report decoupled from the actual comparison would still pass (round 9).
- The final, shipped form is **behavioral**: it builds a real 160,001 B fixture, runs the actual unmodified gate against it, and asserts rejection with the specific "ABOVE CEILING" reason - conjoining exit status with output content closed the last vacuous-pass class (round 10-11).

The shipped test is net-smaller than round 10's intermediate form, but the review cost across eleven rounds was disproportionate to what is, at its core, a documentation correction plus one regression test. That tension is named explicitly below under Pillar 8.

## North Star alignment

- **Pillar 2 (verifiable outcomes)**: this PR replaces an asserted-but-unverified claim ("CEILING is a safety boundary" with no swept confirmation) with a measured one (a real, independently-reproduced sweep result), and ships a genuinely behavioral regression test rather than a documentation-only fix.
- **Pillar 7 (prevent defects at the producing step)**: the shipped test now exercises the real gate against a real fixture, so a future CEILING raise that exceeds the swept bound fails a mechanical check before it ships, rather than only being catchable by a human rereading prose.
- **What cuts against it - Pillar 8 (machinery only as complicated as the benefit requires)**: eleven review rounds on a single test assertion is a real cost against this pillar. Each round found a genuine, reproducible defect (not bikeshedding), so the rounds were not wasted individually, but the cumulative process cost is disproportionate to the size of the underlying change (a doc correction plus one regression test). This is named here rather than hidden; the fallback the reviewers offered throughout - deleting the `CEILING_SCRIPT` half entirely and shipping only the `AGENTS.md` pin - remained available at every round and was not needed only because each round's fix held under the next round's adversarial pass.

## Doc-sync

Doc-sync: clause 5 (user-facing behavior/documented convention) triggered -> updated `scripts/check-skill-embed-budget.sh`, `AGENTS.md`, `docs/skill-embed-injection-sweep.md`, `scripts/check-gemini-skill-budget.sh`, `scripts/skill-embed-sweep-harness.sh`, and `bin/tests/test_skill_embed_sweep_harness.sh`: replaced every "sweep has not been run / do not cite as verified-safe" assertion with the 2026-09-03 measured result, and added a genuinely behavioral regression test for the CEILING/swept-bound invariant.

## Test plan

- [x] `bash scripts/check-skill-embed-budget.sh` - pass
- [x] `bash bin/tests/test_skill_embed_sweep_harness.sh` - 45 passed, 0 failed
- [x] `bash bin/tests/test_check_skill_embed_budget.sh` - 29 passed, 0 failed
- [x] `bash bin/tests/test_check_gemini_skill_budget.sh` - 30 passed, 0 failed
- [x] Full `bin/tests/` pytest suite (cleared `__pycache__`, `-p no:cacheprovider`) - 2196 passed, 126 subtests passed
- [x] hooks JS suite via the quarantine-aware set - all pass (`wrap-lock-tests` 33 passed separately)
- [x] Verified `.claude/skills/dinostack/SKILL.md` is 138,320 B, zero `DS-45-SWEEP` canaries, and clean `git status` throughout
- [ ] CI green on this PR


tyson-solara6:
Automated review pass (own-PR, no formal verdict per policy): North Star alignment check passed (advances verifiability). Flagging that the `python-bin-tests` CI check is currently failing on this PR - worth a look before merge: https://github.com/Space-Dinosaurs/DinoStack/actions/runs/33827261909/job/100882533301

tyson-solara6:
**Automated review (informational only — self-authored PR, no formal verdict per review policy)**

Reviewed the sweep-figure corrections across all 5 touched files plus the new behavioral test scenario. No blocking issues found:

- All five files consistently cite the 2026-09-03 sweep result (140k/145k/150k/160k B intact; CEILING unchanged at 145,000 B). No stale "unverified" language remains in any touched hunk.
- The new scenario in `test_skill_embed_sweep_harness.sh` genuinely exercises live CEILING-rejection behavior (builds a real 160,001 B fixture, runs the real unmodified `check-skill-embed-budget.sh` as a subprocess, asserts both non-zero exit AND the "ABOVE CEILING" string) rather than pattern-matching source text.
- The "do not raise CEILING on the basis of this sweep" guardrail is repeated near-verbatim in 4 places, so there's low risk of this PR being misread as license to raise the limit.

Two things worth confirming before merge, both apparently unrelated to this PR's content:
- `python-bin-tests` CI is currently failing on `test_tuid_index_concurrent_writes_not_lost` (lost 10/40 concurrent writer entries) in `bin/tests/test_enforce_skeptic_round_cap.py` — not touched by this diff, looks like a pre-existing flake, but worth rerunning/confirming rather than assuming it's safe to ignore.
- Branch is `BEHIND` base (no conflicts) — worth a rebase for a clean CI run.

tyson-solara6:
Automated review pass (own-PR, no formal verdict per self-review restriction).

CI flag: `python-bin-tests` is currently failing on this PR, but the failure is in `bin/tests/test_enforce_skeptic_round_cap.py::test_tuid_index_concurrent_writes_not_lost` (lost 10/40 concurrent writer entries under the `_tuid_index_lock()` flock) - a file this PR does not touch. Per the PR's own diff (`AGENTS.md`, `docs/skill-embed-injection-sweep.md`, `scripts/check-gemini-skill-budget.sh`, `scripts/check-skill-embed-budget.sh`, `scripts/skill-embed-sweep-harness.sh`, `bin/tests/test_skill_embed_sweep_harness.sh`), this looks like a pre-existing/flaky concurrency issue in the tuid-index locking mechanism, not something introduced here - worth a rebase + re-run to confirm it's flaky rather than a real regression before merging.

Note: I would normally file this as an out-of-scope Jira follow-up ticket, but the mcp-atlassian Jira tools were not available in this environment - flagging here instead so it isn't lost.

North Star alignment: ALIGNED (product-discovery check) - replaces an asserted-but-unverified safety claim with a measured, independently-reproduced one and ships a genuinely behavioral regression test.

PR #880, opened 2026-09-07
docs(comparison): add Ponytail to the public comparison page

tyson-solara6:
## What was added

Ponytail (github.com/DietrichGebert/ponytail) is added to the public comparison page `docs/agentic-engineering-comparison.html`:

- A **Ponytail column in all three comparison tables** - 6 + 7 + 7 = 20 rows filled, no blanks.
- An **at-a-glance card** in the alternatives grid.
- A **per-alternative summary block**.
- A **Sources entry**.
- A CSS **`min-width` bump from 2840px to 3040px** to accommodate the widened tables.

One file changed. Header counts and per-row cell counts were verified equal for all three tables (16 columns each), the file parses as HTML, and it contains zero em dash characters.

## Why

This mirrors the Ponytail entry already written into the local research-workspace master comparison. The public page lives in this repo while the master comparison lives in the research workspace, so it silently drifts unless every new alternative is synced across. This PR closes that gap for Ponytail.

## Honest framing

Ponytail is a **narrow-scope, single-axis ruleset plugin** - it governs code volume only, where DinoStack governs correctness, concurrency, and process. DinoStack wins nearly every capability row: independent adversarial review (`/ponytail-review` is the implementer grading its own diff), per-change risk classification, worktree isolation, lint/typecheck/test and QA gates, ticket and PR workflow, and planning artifacts.

**Two rows genuinely favour Ponytail, and the page says so plainly:**

1. **Distribution** - roughly 130k GitHub stars in under three months across ~20 harnesses, versus DinoStack's 11 adapters and no comparable public adoption.
2. **Effectiveness evidence** - Ponytail has a published, adversarially-revised, independently third-party-replicated benchmark (its own rebuilt four-arm run plus a JetBrains 80-task A/B replication). DinoStack has none.

The summary block does not soften either point.

## KNOWN FOLLOW-UP - pre-existing drift not closed here

The public page carried 13 alternatives before this PR and 14 after. The local master comparison carries roughly 18. At least **OpenSpec, Granular, Five Levels, and the Kun Chen workflow** exist locally but are missing from the public page.

That drift predates this change and **this PR deliberately does not close it** - adding four more alternatives would mean four more full columns across three tables and a much larger diff, and each needs its own verification pass. Tracking it here so the gap is recorded rather than forgotten.

## Not merging

DinoStack `main` requires human review; auto-merge is disabled. This ends at "PR open and green".

### DIFF
--- a/docs/agentic-engineering-comparison.html
+++ b/docs/agentic-engineering-comparison.html
@@ -233,7 +233,7 @@
     border-collapse: separate;
     border-spacing: 0;
     width: 100%;
-    min-width: 2840px; /* col1 (240px) + 13 data columns at 200px */
+    min-width: 3040px; /* col1 (240px) + 14 data columns at 200px */
     font-size: 14.5px;
   }
   caption { display: none; }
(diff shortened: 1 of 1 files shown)


tyson-solara6:
Superseded by #881 - the original commit was missing the DCO Signed-off-by trailer and force-push was unavailable, so the corrected commit was pushed on a fresh branch. Same diff, same PR body.

PR #882, opened 2026-09-07
feat(bin): ds-skill-load-rate - measure dinostack skill activation rate (DS-227)

tyson-solara6:
Adds `bin/ds-skill-load-rate`, a read-only diagnostic that measures how often the `dinostack` skill actually loads per session, and specifically how often it loads unprompted (model-initiated) versus because the operator typed `/dinostack`.

## Why

DS-143 removed the three `@`-import lines from `~/.claude/CLAUDE.md` when the skill symlink resolves, so the methodology body now loads only when `/dinostack` is invoked. `hooks/skill-auto-load-check.sh` is a nudge, not a mechanism - it prints a skill-load instruction and cannot force compliance. `docs/planning/ds-183-186-harness-trigger-load.md` replicates that pattern onto Codex, Gemini, Kimi and Copilot. No measurement of activation rate existed anywhere in the repo, so four more harnesses were queued to take the same bet on an assumption.

External evidence that this is a real risk rather than a hypothetical: independent JetBrains testing of a structurally identical hook-plus-skill plugin found the skill self-activated zero times in 10 unforced sessions.

## First measurement

Against `CLAUDE_CONFIG_DIR=/Users/tyson/.claude`, window `2026-08-04T14:12:42.965Z .. 2026-09-03T23:26:51.613Z`, 61 readable sessions:

- unprompted (model-initiated) load rate: **44.3% (27/61)**
- operator-typed `/dinostack`: 2/61
- no load at all: 32/61, of which 30 had the auto-load nudge fire with no load following

Reproduced independently by two reviewers at the same figure and window. This number is the reason the ticket exists; it is not an argument for a particular design change, and DS-227 explicitly excludes changing what is resident versus trigger-loaded.

## How it works

Reads top-level session transcripts (`<config_dir>/projects/<project-dir>/*.jsonl`) and distinguishes three record shapes:

- operator-typed: a `type:"user"` record containing `<command-name>/dinostack</command-name>`, no tool_use emitted
- model-initiated: a `type:"assistant"` tool_use `{"name":"Skill","input":{"skill":...}}`, matched on name plus `input.skill` and never on the `caller` field, since some backends omit it
- the nudge: a `type:"attachment"` record with `attachment.type:"hook_success"` and `hookName:"UserPromptSubmit"`, separable from an actual load

Counts both `dinostack` and its pre-rename name `agentic-engineering` (renamed 2026-08-09 in `1e777841`, which the measurement window straddles). Config dir resolves via `bin/_lib.py::resolve_claude_config_dir()`, never a hardcoded `~/.claude`. Reports `ABSENT` rather than `0` when there is no data, so a fabricated zero cannot be mistaken for a measurement.

Round 2 extracted four shared pieces (`MAX_TRANSCRIPT_BYTES`, `iter_jsonl_dict_records`, `transcript_is_oversize_or_unreadable`, `rate_or_none`) into `bin/_lib.py`; `bin/ds-learnings-retrieval-rate` now consumes them. Its local size check is deliberately retained - it returns a `(status, was_oversize)` tri-state the shared boolean cannot express.

## Review

Four engineer rounds, three full Skeptic reviews. Findings by round: 7 Major / 7 Minor, then 2 Major / 3 Minor, then 1 Major / 2 Minor. The most consequential was that the pre-rename skill name was excluded, which understated the headline by 5 points; it now has a regression test whose reddening mutation is documented.

`scripts/check-local.sh` did not complete locally before the last worker's turn ended and is NOT being reported as passed. CI runs the authoritative required set here. `test_grant_consumption_atomic_under_concurrency` in `bin/tests/test_enforce_ticket_batching.py` is a known pre-existing flake on `main` (failed 2 of 4 isolated runs at an earlier SHA on this branch) and is unrelated to this diff.

## North Star alignment

Advances **Pillar 2 (produce verifiable outcomes)**: it replaces an assumption about whether the methodology loads with a measured figure, and the tool's own docstring carries the retirement condition **Pillar 8** requires - this is a diagnostic that is deleted once the rate is measured and recorded, not standing telemetry.

Pillars that cut against, named rather than omitted: it is self-referential machinery that protects the methodology rather than shipping product behavior, which the "not self-defending machinery" non-goal warns about, and it is neutral-to-negative on **Pillar 1 (guard operator attention)** since it produces a number someone must read. Both are accepted on Pillar 8's own boundary, which prefers instrumentation over assumption when a mechanism's yield is unknown - and that is exactly why the retirement condition is mandatory here rather than optional.

**Pillar 4 (universality)**: resolves the config dir at runtime through the shared resolver and takes an explicit scope flag, so it behaves correctly for an operator with a different harness profile or a different project layout.

Refs DS-227.


tyson-solara6:
**Automated review (informational only - self-authored PR, no formal verdict per self-review policy since GitHub blocks self-review).**

Reviewed the full diff (1899 lines) with a fresh adversarial Skeptic pass plus a North Star alignment check.

**Skeptic findings: 0 Critical, 0 Major, 2 Minor.**
- Minor: `resolve_scope()` in `bin/ds-skill-load-rate` joins `--project-dir` directly onto `projects_root` with no `..`/traversal guard, unlike the properly-bounded `--repo-path`/`--all-projects` paths. Low real-world impact (local, operator-invoked diagnostic, no privilege boundary, only prints aggregated counts), but worth tightening for consistency with the tool's own "scoped, read-only" framing.
- Minor: could not independently verify whether `README.md`/`content/SKILL.md`/`CONTRIBUTING.md` need a bin-tool-count update, since no local checkout was available in the review environment. `bin/AGENTS.md` is correctly updated in the diff.

Everything else checked clean: record-shape classification (operator/model/nudge) is mutually exclusive with no double-count path, config-dir resolution is not hardcoded, the `bin/_lib.py` extraction doesn't regress `bin/ds-learnings-retrieval-rate`, and test coverage specifically targets rename-handling, ABSENT-vs-0, and oversize-transcript edge cases.

**North Star alignment: ALIGNED.** Produces a one-time verifiable measurement (Pillar 2) of whether skill self-activation actually works post-DS-143, and self-limits its own attention cost via a stated retirement condition rather than becoming standing telemetry (Pillar 1).

CI is fully green. `mergeStateStatus` is `BLOCKED` (expected for a self-authored PR pending a human review/approval, not a merge conflict).

No pre-existing out-of-scope issues were spotted, so no Jira follow-up tickets filed for this PR.

PR #895, opened 2026-09-13
fix(ds-team): deliver methodology to headless gemini/kimi workers, refuse rather than dispatch silently

tyson-solara6:
`bin/ds-team` dispatches headless Gemini and Kimi leaf workers with no methodology at all, silently. This delivers it as a file in the run directory, and refuses to dispatch rather than proceeding when it cannot.

## The defect

`bin/ds-team` built `augmented_brief = _LEAF_WORKER_CLAUSE + brief_text` identically for all nine harnesses. No harness received any methodology. For Gemini that matters because `activate_skill` is absent from the headless tool schema, and since DS-184 made the adapter trigger-loaded, `.gemini/GEMINI.md` is a pointer stub carrying no binding rules. So a dispatched worker got the brief and nothing else.

Confirmed by live runs against `@google/gemini-cli@0.56.0`, isolating one variable at a time:

```
gemini -p '<prompt>'                      -> "Tool 'activate_skill' is not declared in the available tool schema"
gemini --skip-trust -p '<prompt>'         -> activate_skill ABSENT
gemini --skip-trust --yolo -p '<prompt>'  -> activate_skill AVAILABLE
```

`--yolo` is what restores it, not workspace trust - and `--yolo` auto-approves `run_shell_command`, `write_file` and `replace`, which is not a posture for a dispatched worker. So skill activation is not a viable delivery route.

Kimi has the same gap, and `content/references/cross-harness-teams.md:142` claimed methodology was "loaded via inline skill content in the brief". A repo-wide grep found no code implementing that. The claim was aspirational; it is now corrected.

## What changes

At dispatch time, for gemini and kimi only, the harness's own full-corpus skill artifact is copied into the run directory and the brief carries a ~250 byte pointer to it. Gemini takes `SKILL.full.md` rather than `SKILL.md`, because the latter is the minimal corpus whose two "Deferred at this corpus" blocks point at a sibling file a throwaway workdir does not contain. Kimi's `SKILL.md` is already full corpus.

`_resolve_skill_artifact` returns one of four states. `ok` requires a real read-probe, not `is_file()` - a mode-000 file returns `is_file() == True`. `unreadable` does not fall through to a global candidate, so a broken project-local artifact is reported rather than masked. `empty` catches the zero-byte case, which matters because both build scripts write via a `{ ... } > "$dst"` redirect that truncates at open, so an aborted build leaves exactly that. `absent` is the remaining case.

Refusal happens before `run_dir.mkdir`, and a copy failure removes the run directory, so no refusal path leaves state behind. `AGENTIC_TEAM_ALLOW_NO_METHODOLOGY=1` is the deliberate-consent escape hatch, mirroring the existing `AGENTIC_TEAM_ALLOW_COPILOT`.

The seven other harnesses are untouched: `methodology_note` stays empty and the assembled brief is byte-identical to before.

## Why a file rather than inlining

The first design inlined the body into the brief. That cannot run on Linux: `MAX_ARG_STRLEN` caps a single argv element at 131,072 B, and the two artifacts are 148,550 B and 156,948 B - over the cap before the brief is added. `bin/ds-team` passes the brief as one argv element, so on Linux it would have raised an uncaught `OSError` after `run_dir.mkdir`, leaving an orphaned run directory. macOS has no per-arg cap, so local testing masked it, and all three `bin-tests.yml` jobs run `ubuntu-latest`. No test would have caught it either, because the scenarios used a small fixture.

The file route removes the constraint entirely and costs ~250 bytes of prompt instead of ~150 KB. Stdin was evaluated and rejected: only `codex exec -` has a confirmed prompt-from-stdin route.

## Verification

Sixteen tests, including the four states, the override, the seven-harness regression, a copy-failure injection, and a real-artifact scenario using the actual 148,550 B and 156,948 B files while asserting the added brief overhead stays under 1,024 B - the bound that would have caught the argv defect.

Each of the defects found during review has a mutation proving its test reddens: removing the read-probe, computing the pointer but never assigning it, dropping the `rmtree`, and removing the empty check each redden one specific test.

`pytest bin/tests/` 2,383 passed with 126 subtests. All seven budget gates OK. `build-all.sh` clean over the full 17-entry adapter-sync pathspec. `scripts/check-local.sh` was not run - it exceeds the tool's time cap and cannot complete in one call; CI is the authoritative gate here.

## Known limitations, stated rather than hidden

Nothing forces the worker to open the file. The pointer is imperative and the leaf-worker clause precedes it, but compliance is the same reliability class every skill-trigger mechanism depends on. Confirming it needs a live dispatch, deferred until Gemini quota resets.

A partial-but-nonempty truncation is not caught. The zero-byte case is the documented hazard of the `> "$dst"` pattern; a size floor was considered and rejected because any fixed threshold would either miss a large partial truncation or false-positive when an artifact legitimately shrinks.

The delivered file contains conductor and git-workflow content written for a different role than the one receiving it. The pointer says so explicitly and the leaf clause precedes it, but this does not rewrite the content.

`.gemini/skills/dinostack/` has no `references/` siblings, so `SKILL.full.md`'s internal pointers resolve only when the workdir happens to be a DinoStack checkout.

pi and omp ship stub skill files and were not audited. Both doc surfaces say so.

## North Star alignment

Advances **Pillar 2**: a worker silently operating without the protocol produces output nobody can trust was risk-classified or reviewed, and the failure was invisible in both directions. Advances **Pillar 4**: the cross-harness team feature worked as documented on neither gemini nor kimi. Advances **Pillar 7**: the artifact check moves to dispatch time rather than being discovered later by a reviewer noticing a worker ignored the methodology.

Cuts against **Pillar 8**: this is a new per-harness table and a new refusal path. Its ship test - the failure it catches is live-confirmed silent methodology loss; it retires if a future gemini-cli or kimi-cli gains a headless skill-trigger route this repo has live-verified end to end rather than merely documented.

A behavior change worth naming: a dispatch from a workdir with no project-local adapter and no global install now hard-fails where it previously succeeded silently. That is the intended change, and the override exists for callers who want the old behavior deliberately.

Refs DS-229.


tyson-solara6:
## Automated review (self-authored PR - no formal verdict per self-review policy)

Ran adversarial (Skeptic) review against the diff plus a North Star alignment check.

**North Star alignment: ALIGNED.** Closes a silent-failure gap where headless gemini/kimi workers produced unverifiable output with no signal to the operator - directly advances Pillar 2 (verifiable outcomes) by converting an invisible methodology-skip into a hard refusal. The small added refusal path is justified friction (surfaces a genuine decision point rather than adding routine babysitting), so it doesn't meaningfully regress the attention test.

**Skeptic findings: 0 Critical, 1 Major, 2 Minor.**

**Major - stale doc copy left out of sync.** `.codex/references/cross-harness-teams.md` still asserts the now-false claim "methodology loaded via inline skill content in the brief" for kimi, and its gemini row has no mention of the new file-delivery/refusal mechanism at all. The other 6 copies of this same harness table (`.copilot/`, `.cursor/`, `.gemini/`, `.hermes/SKILL.md`, `content/references/`, `docs/cross-harness-teams.md`) were updated identically but this one wasn't. Recommend syncing it before merge so a codex-worker reading this specific adapter copy doesn't get stale info about gemini/kimi methodology delivery.

**Minor - `_resolve_skill_artifact` directory-level OSError edge case.** The per-base loop does `except OSError: continue` around `candidate.is_file()`. If `is_file()` itself raises (e.g. an unsearchable intermediate directory), that base is silently treated as "not present" rather than surfacing as `"unreadable"` - inconsistent with the file-level unreadable handling a few lines below, and not covered by any test in the diff. Low-likelihood edge case, not blocking.

**Minor - test count in the PR description doesn't match the diff.** The description claims "16 new tests"; the diff adds 9 (`grep -c '^+def test_' ` on the diff). Several loop over `("gemini", "kimi")` internally rather than using `pytest.mark.parametrize`, which masks whether the second harness's assertions would also pass if the first fails. That said, the specific mutation-tested defects named in the PR body (dropping the read-probe, dropping `rmtree` on copy failure) are genuinely exercised by dedicated tests - the coverage claim holds up even though the "16" count doesn't.

Everything else the brief probed for (4-state artifact resolution logic, refusal-path ordering/cleanup, byte-identical briefs for the other 7 harnesses, the `AGENTIC_TEAM_ALLOW_NO_METHODOLOGY` escape hatch scoping) checked out clean.

PR #911, opened 2026-09-21
Delete three measured-false harness-behavior claims about worktree locks and auto-cleanup

tyson-solara6:
Two claims about Claude Code's isolation-worktree behavior were falsified by measurement. Removing
them took 4 rounds and 24 sites; rounds 2, 3 and 4 each followed a Skeptic withholding sign-off, and
rounds 2 and 3 each *created* a defect while fixing one. Round 4 stopped patching named sites and
restated the region once from measured facts.

## Measured basis

Re-measured on this checkout at round 4, before any prose was written:

- **24** directories under `.claude/worktrees/`, **22** registered entries, **exactly 1 locked**.
- The oldest registered isolation worktree was created **2026-08-10** and is still present at
  **42 days**.
- Earlier: a probe agent sampled its own worktree's lock 4x over ~88 s, all identical, naming the
  *parent session's* pid with a start timestamp ~70 min older than the subagent;
  `agent-a5934e3b920a98bbb` stayed locked ~21 h after its agent finished and its PR merged.

Behavior read off `bin/tests/worktree_model.py::disposition_for`: `locked -> SKIP_LOCKED`,
`dirty -> SKIP_DIRTY`, `clean + unlocked + merge-proven -> ELIGIBLE`.

## What the region now says, and why each sentence is true

| Statement | Why it is true |
|---|---|
| The harness locks each isolation worktree **at creation** | Measured; recorded in `AGENTS.md`; already cited at `bin/ds-cleanup-worktrees:770` |
| **While that lock is held**, git refuses non-force removal and branch deletion from a concurrent session | git behavior; the basis of `SKIP_LOCKED` |
| The lock is **routinely absent**, and its absence does not prove a worktree is abandoned | Measured 1 locked of 22; canonical caveat lives in `bin/ds-cleanup-worktrees`, now **pointed to** from both sites |
| A **dirty** worktree is never auto-removed and persists until explicit removal | `disposition_for -> SKIP_DIRTY` |
| A **clean, unlocked, merge-proven** isolation worktree **is** reaped | `disposition_for -> ELIGIBLE` |
| **Nothing outside this methodology** reclaims a stale isolation worktree | The 42-day-old registered worktree above |

## Round 4 - the three Majors

**Major 1 - round 3 traded an implicature for a falsehood.** "Isolation worktrees persist until the
conductor explicitly removes them" is false, and section 11's own paragraphs 8 and 12 plus §Standing
authorizations all describe the reap that contradicts it. The `with changes` qualifier round 3
deleted was independently true (`SKIP_DIRTY`), not merely a presupposition carrier.

Fixed **by subtraction, not by restoring the false contrast**: paragraph 8 already owns *both*
removal triggers, so paragraph 14's trailing cleanup sentence was redundant as well as wrong and is
deleted outright. Paragraph 14 now describes the lock and nothing else. The deck bullet, which has
no such neighbouring paragraph, states both halves truly instead.

**Major 2 - R2-Major-2 was not closed.** The round-3 insert bounded only the consequent; the
paragraph still opened with an unqualified standing property, and neither site carried the caveat or
a pointer to it. Both sites now bound the opening *and* carry the pointer to the canonical caveat.
The Pillar-8 rationale for not restating that caveat in full is only valid while a pointer exists -
so the pointer is itself pinned.

**Major 3 - the harness-sweep claim, unreachable by any lock-pattern grep.** The deck asserted
"Claude Code's own 30-day orphan sweep" with a cadence and reasoned from it. Falsified by the
42-day-old worktree. **The contradiction sweep then found a third copy nobody had named** -
`content/references/worktree-lifecycle.md:980`, the source the deck callout was derived from. Both
now state that the per-session prune is the only sweep, not a backstop to one.

## Round 4 - Minors

- `bin/ds-cleanup-worktrees` held **two contradictory copies** of the justification round 3 corrected
  elsewhere in the same file. Both now give the sound reason: no directory left to protect, and the
  lock is cross-session protection this tool must never override.
- §Standing authorizations read *narrower* than measured reality. It now states that the lock it
  cites is routinely absent, so the residual is the common case - a risk disclosure grounding an
  automatic destructive reap should not read as narrower than it is.

## Regression pins - corrected rationale

**The previous round's stated reason for adding no test was wrong on the facts.** It claimed the only
available shape was a negative pin that goes permanently green; in fact
`bin/tests/test_worktree_lifecycle_spec.sh` already used positive, fail-loud `grep -qF` prose pins,
which is the shape `AGENTS.md` prescribes. That file now carries `check_lock_claim_pins()` with
**8 positive pins**, each naming the mutation that reddens it:

| Pin | Reddening mutation |
|---|---|
| `locks each isolation worktree at creation` (§11) | delete `at creation` -> restores the unscoped standing-property claim |
| `each isolation worktree at creation` (conventions) | same |
| `Do not read the lock as standing protection` (§11) | delete the caveat -> lock reads as universal protection |
| `Do not plan around it as a standing guarantee` (conventions) | same |
| `(Locked handling) is canonical` (§11) | drop the pointer -> Pillar-8 rationale becomes invalid |
| `Locked handling)` (conventions) | same |
| `nothing outside this methodology reclaims a stale isolation worktree` | restore the 30-day-orphan-sweep sentence |
| ``is never auto-removed (`SKIP_DIRTY`)`` (deck) | re-de-qualify to the false "all worktrees persist" claim |

**Mutation-tested: 8/8 redden, 0 decorative**, with the tree restored and the suite green afterward
(`baseline exit=0` -> 8 mutations each `exit=1` -> `post-restore exit=0`).

## Slide overflow - fixed, not baselined

The corrected bullet pushed slide 5 to **798px** against a 720px boundary - a *new* overflow, which
`AGENTS.md` says to fix rather than baseline. Trimmed to **0 new overflows** by shortening two
bullets, merging a redundant code comment, and reducing that slide's scoped code-block size; no
content dropped. The baseline stays at **68** entries (round 3 had ratcheted it 69 -> 68 by deleting
a duplicate callout).

## Gates

Run against the committed tree, no concurrent repo activity:

- `bash scripts/check-local.sh` - adapter drift (verbatim `adapter-sync.yml` pathspec),
  symlink-relativity, methodology drift, all 73 `bin-sh-tests`, 56 `hooks-js-tests`,
  `hooks-sh`/`hooks-python`/`wrap-lock`, and all 7 budget gates: **pass**.
- `pytest bin/tests/`: **2392 passed, 130 subtests, exit 0** standalone.
- All seven `scripts/check-*budget*.sh`: **exit 0**. **Kimi is the figure to watch** - this round's
  added caveat and pointer took it to 240 B headroom and a wording trim recovered it to **309 B**.
  It passes, but it is the tightest of the seven and the next prose addition to this region should
  check it first.
- `bash scripts/build-all.sh` (11 adapters) + verbatim-pathspec `git diff --exit-code`: clean.
  `.codex/skill-compatibility.yml` regenerated twice (a tracked-token line changed, shifting
  `line_fingerprint`) via temp-file + `mv`, never a truncating redirect; the drift was exactly the
  one new `bin/ds-cleanup-worktrees` occurrence this PR's pointer adds.
- `check-methodology-drift.sh --regenerate` then re-verified: OK, 12 files.
- `build-slides.sh` with regenerated `.html` committed; `check-slide-overflow.sh` 0 new / 0 stale.
- Codex, in a detached scratch worktree: `test_codex_skills.py` **127 tests OK**;
  `.codex/build.sh` + `git diff --exit-code` clean.
- Em-dash check scoped to the staged diff; symlinks verified unabsolutized before each stage.

**One flaky gate, characterised not worked around.**
`test_enforce_ticket_batching.py::test_grant_consumption_atomic_under_concurrency` failed one
full-suite run in rounds 2 and 4. A naive sequential control (25/25 pass on pristine `origin/main`
vs 4/25 fail on the branch) looked like a regression; that was a system-load artifact. Interleaving
both so they saw identical load gave **branch 5/40, base 4/40** - indistinguishable - and this PR
touches no importable Python. Standalone re-runs pass 2392/2392. Sibling of the known
`test_enforce_skeptic_round_cap.py` concurrency flake.

Separately, `test_kimi_install_symlink.sh` fails while this change is staged-but-uncommitted: its
`_reset_repo_symlinks` uses `git checkout --` (restores from the *index*) then asserts `git status
--porcelain` is empty. It passes 3/3 on the committed tree. Pre-existing property of that test.

## Deliberately left, checked and reported

- The `/ds-wrap` lock passages (`wrap-ticket.md`, `ds-implement-ticket.md` Phase 11b,
  `wrap-context-format.md`) - a different lock, byte-pinned by `test-phase11b-release-scoping.js`.
- The deck's "once the lock releases" - temporally agnostic, matching "once the lock is genuinely
  released" that the methodology keeps elsewhere.

## What stays

The `ds-cleanup-worktrees` age floor is untouched and still fires. The §Guardrail's payload - never
unlock or double-force-remove a worktree whose directory still exists - survives byte-for-byte.
**No enforcement floor is removed.**

## North Star alignment

**Pillar 2 (verifiable outcomes)** - primary. These sentences asserted harness behavior nobody had
measured; one justified a safety gate, one a named residual risk, one a slide's whole argument. The
region is now stated from a census and from `disposition_for`, and pinned so it cannot silently
revert.

**Pillar 8 (machinery only as complicated as the benefit requires)** - advanced. Binding prose at one
canonical site: 24 sites carrying 2 claims is exactly the duplication that let this survive an
earlier falsification (`KNW-20260823-001`). Round 4 deletes a redundant sentence rather than
correcting it, and replaces a restated caveat with a pointer. The 8 pins are the one addition, and
each carries a named reddening mutation per Pillar 8's own "name the failure it would have caught".

**Pillar 5 (guard agent context)** - net deletion from the always-loaded body across the PR.

**Cutting against, named honestly:**

- **The Pillar-5/6/7/8 boundary** - an enforcement floor is never removed to satisfy a pillar. This
  PR edits the stated *justification* for the age floor, the §Guardrail, and a named residual, and
  it *adds* 8 pins rather than removing machinery. No gate's code, default, or behavior changes.
  Round 4's §Standing authorizations edit *widens* a disclosed risk to match measurement rather than
  narrowing it.
- **Pillar 3 (low friction)** - the region no longer tells a reader when the lock ends, only that it
  is set at creation and routinely absent later. Accepted deliberately: inventing a replacement
  account of harness behavior is the exact defect being removed, and 4 rounds of evidence say a
  confident-sounding replacement is how this recurs.
- **Pillar 1 / Pillar 6 (attention, wall-clock)** - 4 review rounds on a prose correction is a poor
  ratio. The convergent move was subtraction over a bounded region plus mechanical pins, not a
  fifth patch; that is the reusable lesson, recorded here rather than left implicit.

## Docs currency

- `docs/index.html` - checked, not in scope: carries no copy of either claim.
- All 18 `docs/slides/*.md` - hand-grepped each round (`check-slides-sync` cannot see a claim stale
  in both `.md` and `.html`). Copies deleted/corrected in rounds 1, 3, and 4; `.html` regenerated.
- `README.md`, `CONTRIBUTING.md`, `content/SKILL.md` - no copy, no stale count or reference.
- Generated adapter surfaces - regenerated via `build-all.sh`, never hand-edited.


tyson-solara6:
Superseded by #912.

Closing per the stop rule recorded in `.agentic/skeptic-round-pr911-delete-false-lock-claims-4bcc7033b6.json` before round 4's review: ship if findings are confined to the descriptive layer, close and rescope if the review again finds a Major created by the round's own fix. Both round-4 Majors were created by round 4's own fix, making it 4 of 4 rounds introducing or leaving a defect.

The discriminating pattern across four rounds: every round that authored new prose introduced a defect, while every deletion held. #912 keeps this PR's deletions and drops every newly-authored assertion - the branch-deletion scoping, the `nothing outside this methodology reclaims a stale isolation worktree` universal negative, the `routinely absent` generalization, and the `at creation, and while that lock is held` insertion.

Two findings from this PR's review arc that #912 acts on directly. Round 4's reviewer measured on git 2.55.0 that `git branch -D` is refused identically whether a worktree is locked or unlocked - the refusal comes from the branch being checked out, not from the lock - so #912 corrects three sites that attributed branch-deletion refusal to the lock, rather than scoping that attribution. And this PR's pointer pins omitted the target path and stayed green when the path was deleted, while the caveat they pointed at was pinned by nothing at all; #912's pins require both.

Work preserved on branch `fix/delete-false-worktree-lock-claims` at `2edb35dc`.

